PR #1, opened 2025-10-19
Optimize Properties and apply clippy suggestions

unschlagbar:
What should i say??
just followed some clippy suggestions and made the return type of on properties trait fn from Vec<String> to Box<[&str]>,
that means no allocation without a reason + Box<[]> saves a usize for capacity.
All list that are not mean to grow should be this type!"


unschlagbar:
ops, im gonna fix that tomorrow

unschlagbar:
Done

4lve:
Thanks, first PR!

PR #5, opened 2025-10-25
packet write perf improvements

unschlagbar:
removed bytes crate (bytes type has a good amount of overhead).
made a new type: `FrontVec` to leave a few bytes in the front reserved for additional bytes like the packet len


4lve:
LGTM! Thanks

PR #10, opened 2025-11-02
Implement registry cache

Joniii11:
This PR cleans up how registry data and tags are sent during the server’s configuration phase. The main change is adding a new RegistryCache that builds and stores all the registry and tag packets ahead of time, making things faster (i hope) and a lot cleaner.

**What’s new:**
* RegistryCache: New struct that precomputes all registry and tag data so we don’t rebuild it every time a client connects.
* Server: Now has a shared registry_cache (using Arc) that’s set up when the server starts.
* Config refactor: The config code now just grabs data from the cache instead of generating it inline.
* Trait tweaks: Added support for using `Arc<T>` for packet trait to allow it to work with the RegistryCache (maybe give better solutions, this was what i came up with)
* Cleanup: Removed old registry/tag setup logic and unused imports.

Also i used a macro to not need the repetetive code to build the registries

### DIFF
--- a/steel-protocol/src/packet_traits.rs
+++ b/steel-protocol/src/packet_traits.rs
@@ -238,21 +238,3 @@ impl EncodedPacket {
 pub trait CBoundPacket: PacketWrite + WriteTo {
     fn get_id(&self, protocol: ConnectionProtocol) -> Option<i32>;
 }
-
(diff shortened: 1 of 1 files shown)


4lve:
LGTM thanks

PR #12, opened 2025-11-04
Change static item to have only one lock instead of a lock for every item

unschlagbar:
Looks cleaner and works better.
This change make the release build 68_000 bytes smaller!


4lve:
LGTM thanks

PR #15, opened 2025-11-04
Implement BitSet codec needed for Chunk Data impl

teknostom:
This PR implements the [BitSet](https://minecraft.wiki/w/Java_Edition_protocol/Packets#Type:BitSet) codec needed for the light masks in the chunk data packet.


unschlagbar:
why lenght instead of len. and why handle a usize to i32 convertion. it should just panic to indicate a bug, not worth handling it this way because of the code look and error string storage

unschlagbar:
not a big deal but you should import result from io: `Result<Self, std::io::Error>` -> `Result<Self>`

PR #18, opened 2025-11-06
chunk restructure

unschlagbar:
For today something to start with


steel-world/src/chunk/level_chunk.rs
```diff
@@ -0,0 +1,47 @@
+use steel_utils::BlockStateId;
+
+use crate::chunk::{paletted_container::BlockPalette, section::ChunkSection};
+
+#[derive(Debug)]
+pub struct LevelChunk {
+    pub sections: Box<[ChunkSection]>,
```
4lve: Should be an interiour SteelRwLock<Box<[ChunkSection]>>

PR #22, opened 2025-11-10
micro optimise xoroshiro

coco875:
This optimisation have been made after a long talk and multiple try. (There is a lot of cpu wizardry here can't explain some of them)


coco875:
arm haven't really gain performance without changing the meaning of the code

coco875:
the last time that I test I have around +30% on x64

4lve:
Alr thx

PR #28, opened 2025-11-20
Added Nylium Adapter for a GUI binary

Guibi1:
This PR adds the `steel-nylium` crate, which adds a simple GUI to the server.
To test, simply run `cargo run -p steel-nylium`.

For the actual GUI features, see https://github.com/Guibi1/Nylium.

Whats currently missing for the PR to be ready:
- [x] Console
- [ ] Log files
- [ ] Commands
- [ ] Player list
- [ ] Settings

### DIFF
--- a/steel/src/main.rs
+++ b/steel/src/main.rs
@@ -47,4 +47,7 @@ async fn main() {
         .await; */
 
     steel.start().await;
+
+    // Keep the server running until cancellation
+    steel.cancel_token.cancelled().await;
 }


4lve:
Thanks for your initiative! It has been decided that the GUI crate should not be in the steel repo to seperate the server and the GUI.

I would happily link your repo in the discord though. You could also look into using the [steel repo as dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories) so you only need the GUI binary in that repo.

PR #31, opened 2025-11-29
made ticking faster ig

Joniii11:
i got times down to 3 to 3.5ms on average per tick. Using some shenanigans or whatever you wanna call it
take a look at `chunk_generation_task.rs` in the StaticCache2D struct and uhh some other stuff which i already shared


Joniii11:
yo wtf why is there so often a Merge branch dev into dev thas stupid

PR #32, opened 2025-12-05
Light Generation

teknostom:
Adds block states that will be needed later. For now, we need it for the opacity.

Adds sky light initialization to the generation steps. Sets block light to 0, as it does not need to be 15 for fullbright.

No propagation yet, but we also don't save the chunks, so no need to do propagation, becasue the client handles that fine on it's own. Once we store chunks, we can impl the light spreading.

### DIFF
--- a/steel-core/src/chunk/chunk_generator.rs
+++ b/steel-core/src/chunk/chunk_generator.rs
@@ -2,6 +2,27 @@
 
 use crate::chunk::chunk_access::ChunkAccess;
 use enum_dispatch::enum_dispatch;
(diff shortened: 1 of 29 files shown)


teknostom:
Reduced the added json from 150k to 64k

tn-lorenz:
Wouldn't it be possible to SIMD iff the light doesn't cross a chunk border? I'm asking because the offsets are vectors

4lve:
We talked in discord about not implementing light until after structures, and once that time comes opening a new PR will be easier. Thanks anyways!

PR #34, opened 2025-12-10
Secure and unsecure chat.

teknostom:
This now works with Signed and unsigned. It uses the same player chat package for all chats though, compared to vanilla which uses disguised chat, for simplicity.


teknostom:
Found some extra time, and got this fully working. Both signed and unsigned. At least on my machine :p

PR #37, opened 2025-12-17
Implement Player sync

teknostom:
This PR implements the neccesary logic for seeing what other players are doing including:

- Movement
- Rotation
- Arm swinging
- Pose shifting

### DIFF
--- a/steel-core/src/chunk/chunk_holder.rs
+++ b/steel-core/src/chunk/chunk_holder.rs
@@ -1,10 +1,12 @@
 //! `ChunkHolder` manages chunk state and asynchronous generation tasks.
 use arc_swap::{ArcSwapOption, Guard};
 use futures::Future;
+use rustc_hash::FxHashSet;
 use std::fmt::Debug;
-use std::sync::atomic::{AtomicU8, AtomicUsize, Ordering};
+use std::mem;
+use std::sync::atomic::{AtomicBool, AtomicU8, AtomicUsize, Ordering};
 use std::sync::{Arc, Weak};
-use steel_utils::{ChunkPos, locks::SyncMutex};
+use steel_utils::{BlockPos, ChunkPos, SectionPos, locks::SyncMutex};
 use tokio::select;
 use tokio::sync::{oneshot, watch};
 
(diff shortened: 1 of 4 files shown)


steel-core/src/player/mod.rs
```diff
@@ -421,12 +435,64 @@ impl Player {
...
+
+            self.world.players.iter_sync(|_, p| {
+                if p.gameprofile.id != self.gameprofile.id {
+                    p.connection.send_packet(move_packet.clone());
+                    if packet.has_rot {
+                        p.connection.send_packet(head_packet.clone());
+                    }
+                }
+                true
+            });
```
4lve: We should mark the player pos as dirty then next time we iter players (tick) we send all update packets. Also need to make sure the players we send to have the chunk in their chunk view

steel-core/src/world/world_entities.rs
```diff
@@ -10,11 +12,21 @@ impl World {
...
-            self_clone.chunk_map.remove_player(&player);
+
+            // Broadcast removal to all remaining players
+            let remove_entity = CRemoveEntities::single(entity_id);
+            let remove_info = CRemovePlayerInfo::single(uuid);
+            self.players.iter_sync(|_, p| {
+                p.connection.send_packet(remove_entity.clone());
+                p.connection.send_packet(remove_info.clone());
+                true
+            });
```
4lve: Same with this, we want to buffer them per tick

steel-core/src/world/world_entities.rs
```diff
@@ -52,18 +67,44 @@ impl World {
...
+                let (existing_yaw, existing_pitch) = existing_player.rotation.load();
+                player.connection.send_packet(CAddEntity::player(
+                    existing_player.entity_id,
+                    existing_player.gameprofile.id,
+                    existing_pos.x,
+                    existing_pos.y,
+                    existing_pos.z,
+                    existing_yaw,
+                    existing_pitch,
+                ));
```
4lve: Same buffering + distance check here

4lve (commented):
Looks good but could use some buffering for certain packets until next tick and we need a distance check. Otherwise looks good

PR #38, opened 2025-12-28
WIP Plugin support

coco875:
have generate a basic implementation of plugin with abi_stable

things who are needed to discuss
* [ ] api with ferrumc
* [ ] split in a separte repo (currently steel-plugin-api depend on nothing so we can make as repo where plugin a link against it and would work with both steel and ferrumc without changing nothing)
* [ ] maybe start with registries api to get id ?


4lve:
Thanks for experimenting a bit with plugins. When the time comes we will most likely use stabby

PR #40, opened 2026-01-09
add Debug mode generator

coco875:


4lve:
Thanks but I don't think this has a usecase for now

PR #43, opened 2026-01-11
Add player pose sync

teknostom:
This pr adds pose sync for players, and also adds some infra for entities.


teknostom:
I implemented the EntityBase, which is shared between the Player and entities; it will be handy when we implement entities. The code looks big, but the code part is pretty small:
[tom@archie SteelMC]$ git diff HEAD~2 --shortstat -- "*.rs"
 14 files changed, 881 insertions(+), 206 deletions(-)

PR #47, opened 2026-01-16
add note block

coco875:


4lve:
Could you also add the sound when clicked?

4lve:
Not currently. You could look into how it's done in vanilla or I can when I have time

coco875:
hmm look like to attach a sound to the world with a position https://mcsrc.dev/#1/1.21.11_unobfuscated/net/minecraft/world/level/Level#L389 that are after all send to player https://mcsrc.dev/#1/1.21.11_unobfuscated/net/minecraft/server/level/ServerLevel#L1034

so maybe you want to do it with optimsation ?

PR #50, opened 2026-01-17
Use text components crate + Tellraw

DarkMrMelther:
## Changes
Changed the text components to work with the text component crate.
Added entity, player, and text component arguments for commands.
Implemented the tellraw command
Added insertion, click event, and hover event to player chat messages to match vanilla.

## Known bugs
There is a bug that shows as incorrect the AskServer arguments, I don't know why since auto completion it's working fine

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2570,8 +2570,7 @@ dependencies = [
 [[package]]
 name = "text_components"
 version = "0.1.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


4lve:
Thanks for the contribution. Though given how critical these text components are, I’d prefer not to pull them from an external dependency. Could we maybe do a local copy into the codebase instead?

4lve:
I did add some hashing and nbt parsing into text components, could you look into that and see if you could replicate it. We also have an org now so we can look into moving the text-components repo and adding you as maintainer

PR #55, opened 2026-01-19
Adding docs

JunkyDeveloper:
I think the discord channel is still good but have also a proper documentation which can be displayed on the homepage later or for people to try out steel without the joining of our discord


JunkyDeveloper:
Didn't know that exists :D

PR #59, opened 2026-01-22
redo improve ci

coco875:
same as https://github.com/Steel-Foundation/SteelMC/pull/41 (redo it because of the difference of change)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,14 +13,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: nightly
-          components: clippy, rustfmt
-          override: true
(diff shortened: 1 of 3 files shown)


coco875:
(I mean when I was trying to upstream master a bunch of merge conflict appears for no apparent reason)

4lve:
Does github have nightly installed by default or do you need to specify it?

coco875:
because of rust-toolchain.toml, it use nightly and it download it automatically so no need to install or specify it

coco875:
https://rust-lang.github.io/rustup/overrides.html
> The toolchain is chosen in the order listed above, using the first one that is specified. There is one exception though: directory overrides and the rust-toolchain.toml file are also preferred by their proximity to the current directory. That is, these two override methods are discovered by walking up the directory tree toward the filesystem root, and a rust-toolchain.toml file that is closer to the current directory will be preferred over a directory override that is further away.

PR #63, opened 2026-01-24
Full noise gen (DO NOT MERGE!!!)

teknostom:
This is a full noise gen example impelementation, so that people can check it out, and try optimizing or finding better ways to do it. Will split this up later into smaller PRs once i feel done with it.

### DIFF
--- a/steel-core/src/server/mod.rs
+++ b/steel-core/src/server/mod.rs
@@ -33,7 +33,7 @@ use crate::command::CommandDispatcher;
 use crate::config::{ConfigWorldType, STEEL_CONFIG};
 use crate::player::Player;
 use crate::server::registry_cache::RegistryCache;
-use crate::world::{World, WorldType, WorldTickTimings};
+use crate::world::{World, WorldTickTimings, WorldType};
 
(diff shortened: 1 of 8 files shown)


4lve:
I do believe all files in here are based on vanilla and not pumpkin. We found out that pumpkin chunk generation mismatched with vanilla

Snowiiii:
> Ah okay, thanks for the update. Even then Pumpkin is already mentioned in the projects readme file which is enough for the MIT license Pumpkin has.

Actually Pumpkin like you already saw on our discord we discussed changing the license (i assume that's why you also recently changed) so Pumpkin is now GPL

teknostom:
> > Ah okay, thanks for the update. Even then Pumpkin is already mentioned in the projects readme file which is enough for the MIT license Pumpkin has.
> 
> 
> 
> Actually Pumpkin like you already saw on our discord we discussed changing the license (i assume that's why you also recently changed) so Pumpkin is now GPL 

Accually, the version i based my code on has MIT license in the repo, don't think a LICENSE can apply backwards.

PR #68, opened 2026-01-27
Custom logger with autocompletion

DarkMrMelther:
Full logging system with input, configurable, with history, autocompletion of commands, selection and more...
All integrated as a Trace Layer. 😎


DarkMrMelther:
Conflicts solved, but I want to clean it a bit before beign reviewed, I will write you on Discord when it's finished!

PR #69, opened 2026-01-29
Add fire and stuff

teknostom:


4lve:
We finally have some real foundation around scheduled ticks and we have an unimplemented fire block, feel free to open a new PR using these new systems. Thanks!

PR #85, opened 2026-02-06
slabs and stairs

Joniii11:
adds slabs and stairs


4lve:
Thanks, feel free to reopen a PR using the new block system

PR #91, opened 2026-02-07
Entity Events Packet + Fast Gamemode Change

DarkMrMelther:
Added the Entity Event Packet, Change Game Mode packet, and set the player op level to 4 when adding player.

### DIFF
--- a/steel-core/src/server/mod.rs
+++ b/steel-core/src/server/mod.rs
@@ -194,7 +194,7 @@ impl Server {
                 dimension: dimension_key,
                 seed: hashed_seed,
                 game_type: player.game_mode.load(),
-                previous_game_type: None,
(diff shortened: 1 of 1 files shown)


steel-protocol/src/packets/game/c_entity_event.rs
```diff
@@ -0,0 +1,145 @@
+use steel_macros::{ClientPacket, WriteTo};
+use steel_registry::packets::play::C_ENTITY_EVENT;
+
+/// Status type for the entity event packet.
+#[derive(WriteTo, Clone, Copy, Debug, PartialEq, Eq)]
+#[write(as = VarInt)]
+pub enum EntityStatus {
```
4lve: Can we extract this from the extractor instead?

PR #92, opened 2026-02-12
Flint cicd

JunkyDeveloper:
my cicd version


bensaru (commented):
this will work? seems like there is a bug in at 32 line.

JunkyDeveloper:
This should be now a working ci with flint-steel. It will fail then a test fails which was run successful in the past.
New or already failing tests, will have no influence. So I think at the end that's a good solution. for also the block implementations. so you can see if something breaks a block impl in a PR

PR #107, opened 2026-03-04
Add POI system

Purdze:
## Summary
- Add POI type registry with O(1) block state lookup, tag support, and build script codegen from extracted data
- Add per-section and world-level POI storage with spatial queries (square, circle, nearest, sorted-by-distance)
- Integrate POI scanning into chunk loading and block state changes

## Test plan
- [x] Unit tests for pack/unpack, add/remove, and occupation filtering
- [ ] Verify POI population on chunk load with a real world
- [ ] Test nether portal lookup via `get_nearest`

### DIFF
--- a/steel-core/src/poi/poi_storage.rs
+++ b/steel-core/src/poi/poi_storage.rs
@@ -155,6 +155,7 @@ impl PointOfInterestStorage {
     }
 
     /// Reserves a ticket at the given position. Returns `true` if successful.
+    #[must_use]
     pub fn reserve_ticket(&mut self, pos: BlockPos) -> bool {
         let (chunk_pos, section_y, packed) = resolve_pos(pos);
         let Some(set) = self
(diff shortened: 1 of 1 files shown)


steel-core/src/poi/poi_set.rs
```diff
@@ -0,0 +1,197 @@
...
+use super::poi_instance::PointOfInterest;
+use super::poi_storage::OccupationStatus;
+
+/// Stores all POIs within a single chunk section (16x16x16).
+///
+/// Uses a packed local position `(y << 8 | z << 4 | x)` as the key,
+/// with a secondary index by POI type for type-filtered queries.
+pub struct PointOfInterestSet {
+    pois_by_pos: FxHashMap<u16, PointOfInterest>,
+    pois_by_type: FxHashMap<usize, FxHashSet<u16>>,
```
4lve: Maybe use the PoiRef as the key here as well

PR #116, opened 2026-03-10
Crops and Bonemeal

CuzImClicks:
azalea, bamboo, bamboo_sapling, beetroots, flowers, seagrass, tall_seagrass, tall_grass, double_plant_block, nether sprouts, nether warts and torchflower

### ToDo

- [x] crops dont drop an item when the block below them is broken
- [x] two high crops only drop an item when the bottom block is broken
- [x] two high the placing of top half of a two high block needs review (vanilla uses setPlacedBy, this uses on_place)
- [x] bonemealing a two high block should call `fn pop_resource(self: &Arc<Self>, ...)` but i only have a &World
- [x] pitcher crops and other vegetation blocks are missing (might leave them for somebody else)

~~still have to do more bonemeal implementations and adapt cacti and sugarcane to the new system.~~

~~also the current bonemeal system needs to be reworked to support spreading and being called from the bonemeal item instead of use_item_on being implemented for every single bonemealable block~~


4lve:
Thanks, this uses the old block registration system though. Feel free to reopen if you rework or port it to the new system

steel-core/build/classes.json
```diff
@@ -2400,11 +2402,15 @@
     },
     {
       "name": "carrots",
-      "class": "CarrotBlock"
+      "class": "CarrotBlock",
+      "max_age": 7,
+      "clone_item_stack": "carrot"
```
4lve: What purpose do these serve? I think it might be better to just hardcode them like vanilla

CuzImClicks: Reduces code duplication since they all do precisely the same

4lve: For sure, will take a look

PR #108, opened 2026-03-04
Hunger

Joniii11:
i am hungry now

PR #121, opened 2026-03-15
Add clientbound packets: ChangeDifficulty, SetDefaultSpawnPosition, SetExperience

OmarAfet:
- **Add CChangeDifficulty clientbound packet**
Informs the client of the current difficulty level and lock state.
- **Add CSetDefaultSpawnPosition clientbound packet**
Sets the player's compass target and default respawn position with dimension, coordinates, and angles.
- **Add CSetExperience clientbound packet**
Sends the player's experience bar progress, level, and total points.


Joniii11:
difficulty packet is implemented here #108

PR #132, opened 2026-03-16
Software Heritage Archive

JunkyDeveloper:
I already made a PR: #48 
This will now be a draft until I think world gen is finished then I will open it. So steel is saved forever!

A GitHub Action that saves the GitHub repository it is being run on to the [Software Heritage Archive](https://www.softwareheritage.org/).

This is something like the wayback machine but for software code, so even if github will go down or shutdown or paywall this software will not be lost!


tn-lorenz (changes requested):
Grammatikalisch ist das idiomatischer :D

PR #133, opened 2026-03-17
Cakes and Candle Cakes

CuzImClicks:
depends on #108 and adds a new Extractor for <Candle to CandleCake>


JunkyDeveloper:
@CuzImClicks is this still open and ready to be merged, so I can do a review in the next days if you want

PR #138, opened 2026-03-18
better gitignore

chocodev11:


JunkyDeveloper (changes requested):
added some things which can be deleted because flint changed

PR #139, opened 2026-03-19
More vectors

suprohub:
This pr tries use vectors more.
Instead of using _x _y _z better use vector.


suprohub:
Maybe i continue later

JunkyDeveloper:
@suprohub any updates here?
Don't want to make pressure :D

suprohub:
Maybe continue on thisa week or seconda week

suprohub:
too outdated

PR #154, opened 2026-04-01
Fix update-minecraft-src.sh script for MacOS

Morb0:
## Summary
- Fix sed -i compatibility in update-minecraft-src.sh for macOS
  - macOS sed requires a backup extension argument, unlike GNU sed
  - Use sed -i.bak which works on both macOS and Linux


4lve:
I'm not sure if this version pinning is required anymore with the latest GitCraft version. Could you try removing it completely?

PR #160, opened 2026-04-02
Campfires and Smelting Recipes

CuzImClicks:
Depends on #108


CuzImClicks:
i dont want to maintain this anymore. recipes need full rework and then campfire is trivial

PR #168, opened 2026-04-21
Fully macro registries

JunkyDeveloper:
remove 30 lines in 30 files.

Claude generated a macro, which I tried to write by hand, which does what I wanted to achieve.
So the only problem what i see, we have at least 4 macros of depth. So macro in macro in .... .
I check the macros and they look in general good and ok and make sense in general.

Only my concern is the compile time, so a person with more knowledge should check the macro and see how to improve it!

This PR also holds the code of #167

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1763,6 +1763,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "paste"
+version = "1.0.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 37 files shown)


4lve:
I'm not sure if this is the right approach or not since it abstracts a lot of the registries away. Need to consider pros and cons

4lve:
Thanks for the PR but I think this abstracts too much as the current way is already pretty abstracted. Thanks anyways!

PR #172, opened 2026-05-01
ban/shadowban/whitelist for IPs

JunkyDeveloper:
Added the commands:
- /whitelist-ip list
- /shadowban-ip add/remove/list
- /ban-ip <ip|player> <reason|expire> <reason>
- /pardon-ip <ip>
- /kick <reason>

Thanks @ht06 for this work and help here!

it saves in ip-bans.toml.
More information and the documentation about everything can be found in this [PR](https://github.com/Steel-Foundation/SteelDocs/pull/18).
Also a converter from minecraft format to steel format is added.
If steel doesn't find at boot the ip-bans.toml it looks if banned-ip.json can be found and load the data from that.
This can be also triggert via CLI with: `./steel convert banned-ip` which loads the data and saves it to disk. The server will not be started


hthug06:
And also the /kick command

JunkyDeveloper:
ahh yes forgot that, good catch!

steel-core/src/command/arguments/duration.rs
```diff
@@ -0,0 +1,89 @@
...
+            ArgumentType::String {
+                behavior: ArgumentStringTypeBehavior::SingleWord,
+            },
+            Some(SuggestionType::AskServer),
+        )
+    }
+
+    /// ONLY FOR THE CONSOLE\
+    /// (If you want to also suggest to the client,
+    /// put the `SuggestionType` to `AskServer`)
```
coco875: I don't think that this comment are necessary

steel-core/src/command/commands/ban_ip.rs
```diff
@@ -0,0 +1,260 @@
...
+                         ctx: &mut CommandContext| {
+                            ban_ip(ctx, targets, Some(reason), Some(duration))
+                        },
+                    )),
+            )
+            // IP + reason
+            .then(argument("reason", StringArgument::greedy()).executes(
+                |(((), ip), reason): (((), Option<IpAddr>), String), ctx: &mut CommandContext| {
+                    ban_ip(ctx, ip, Some(reason), None)
+                },
```
coco875, line 69: I feel like there is a duplication of code here, there is no way to clean it up ?

steel-core/src/command/commands/ban_ip.rs
```diff
@@ -0,0 +1,260 @@
...
+    // Check if this is a player, or the console / command block or something else
+    let final_sender = sender.get_player().map_or_else(
+        || "Server".to_string(),
+        |sender| sender.gameprofile.name.clone(),
+    );
+    // The reason is a string because a TextComponentArgument can't work with the duration (it consume it)
+    // So we need to check if the reason is empty or not and then have the real reason
+    // Then, try to get it from the snbt format. If this is compatible, use it, else use TextComponent::plain
+    let real_reason = reason.unwrap_or("Banned by an operator.".to_string());
+    let text_component_reason = match TextComponent::from_snbt(real_reason.as_str()) {
```
coco875, line 130: if you expect a textCompenent use TextComponent in command ? or there is more ?

steel-core/src/command/commands/kick.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    let real_reason = reason.unwrap_or(MULTIPLAYER_DISCONNECT_KICKED.msg().component());
+
+    for player in players {
+        // First, disconnect the player
+        player
+            .connection
+            .disconnect_with_reason(real_reason.clone());
+
+        // Then inform the sender of the command
```
coco875: a bit too much comment int the function to my taste, don't know if it's ok for other

steel-core/src/command/commands/steel.rs
coco875, line 1: why there is this change ?

coco875: (I mean it's format and change how a comment are so not necessary for the pr)

steel-core/src/network/ip_access_policy.rs
```diff
@@ -0,0 +1,521 @@
...
+    pub expires: Option<DateTime<Utc>>,
+    /// Operator-provided reason shown to the banned client.
+    #[serde(with = "reason_json_string_format")]
+    pub reason: TextComponent,
+}
+
+/// On-disk shape of `config/ip-bans.toml`. Holds both the metadata-rich
+/// ban list and the bare blacklist. In-memory the blacklist is a
+/// `FxHashSet`; TOML has no native set type so it round-trips through
+/// `Vec<IpAddr>`.
```
coco875: can you add a ref to the class who are the real IpBans struct ?

steel-core/src/network/ip_access_policy.rs
```diff
@@ -0,0 +1,521 @@
...
+            expires: v.expires,
+            reason: v
+                .reason
+                .filter(|s| !s.to_string().is_empty())
+                .unwrap_or_else(|| TextComponent::plain(DEFAULT_BAN_REASON)),
+        }
+    }
+}
+
+// TODO: DELETE ME
```
coco875: ?

steel-core/src/network/ip_access_policy.rs
```diff
@@ -0,0 +1,521 @@
...
+    Ok(Some(
+        vanilla_entries.into_iter().map(BannedIP::from).collect(),
+    ))
+}
+
+/// Path to vanilla's banned-ips file (`config/banned-ips.json`).
+pub const VANILLA_BANNED_IPS_PATH: &str = MINECRAFT_BANNED_IP_PATH;
+
+/// Path to Steel's bans file (`config/ip-bans.toml`).
+pub const STEEL_IP_BANS_PATH: &str = BANS_PATH;
```
coco875: what is the point having two constante who as the same value ?

steel-core/src/network/ip_access_policy.rs
JunkyDeveloper: Thanks good catch XD

JunkyDeveloper: fixed in f971d71a

steel-core/src/network/ip_access_policy.rs
JunkyDeveloper: good catch, fixed it here: 4858cf0a

steel-core/src/command/mod.rs
```diff
@@ -46,6 +50,7 @@ impl CommandDispatcher {
         dispatcher.register(commands::time::command_handler());
         dispatcher.register(commands::tp::command_handler());
         dispatcher.register(commands::weather::command_handler());
+        dispatcher.register(commands::whitelist_ip::command_handler());
         dispatcher.register(commands::difficulty::command_handler());
         dispatcher.register(commands::steel::command_handler());
         dispatcher.register(commands::xp::command_handler());
```
coco875, line 59: a comment that I forget, it look like to be sorted A-Z but you put shadowban too early

JunkyDeveloper: correct order: c179ea50 16638da2

steel-core/src/command/commands/steel.rs
JunkyDeveloper: only format but as long as cargo fmt don't complain don't see a problem

steel-core/src/command/arguments/duration.rs
JunkyDeveloper: As a person who don't really work with commands, it's a good hint to understand that argument better. For me this comments has a purpose

steel-core/src/command/commands/ban_ip.rs
JunkyDeveloper: What I understand it, it's already as undup as possible, as you can see ban_ip is the function the rest is to make all optional possible, which we provide

steel-core/src/command/commands/ban_ip.rs
JunkyDeveloper: As I remember, we had a personal discussion how to solve that. Because TextComponent needs to be at the end of each command and we needed to do this "workaround" to make it working.
Maybe that is a bug in TextComponentArgument so @DarkMrMelther can give some insides. If no one understands why, this is the best version

steel-core/src/command/commands/kick.rs
JunkyDeveloper: yes agree with: a791b334

steel-core/src/network/ip_access_policy.rs
JunkyDeveloper: I think we clarified that on discord. IpBans struct is the real struct which is used to organize the banned ips

steel-core/src/command/commands/ban_ip.rs
DarkMrMelther: This is my fault, from_snbt lacks of enough control, I need to rewrite it to show where an error ocurred, and return the unused text at the end

JunkyDeveloper: ok, I would still like to get it merged and solve that issue after my exams :)

steel-core/src/command/arguments/duration.rs
coco875: hmm I think this kind of command a better on the trait then here

steel-core/src/network/ip_access_policy.rs
```diff
@@ -0,0 +1,507 @@
...
+const DATETIME_FORMAT: &str = "%Y-%m-%d %H:%M:%S %z";
+const FOREVER: &str = "forever";
+/// Substituted as the ban reason when a vanilla `banned-ips.json` entry has
+/// no reason or a `null` / empty one. Steel always stores a reason; this
+/// string is the user-visible message a vanilla-imported ban falls back to.
+const DEFAULT_BAN_REASON: &str = "Your IP was banned";
+
+/// Path to Steel's bans file (`config/ip-bans.toml`).
+pub const STEEL_IP_BANS_PATH: &str = "config/banned-ips.toml";
+const MINECRAFT_BANNED_IP_PATH: &str = "config/banned-ips.json";
```
coco875, line 34: don't know if 4lve prefer no file and if migration it should have, it will be with a tool and not steel but not sure, personally I am fine with that

steel-core/src/network/ip_access_policy.rs
coco875: yes I suggest that in comment to make it more clear.

steel-core/src/network/ip_access_policy.rs
```diff
@@ -0,0 +1,507 @@
...
+}
+
+/// Internal state bundle held behind [`IpAccessPolicy`]'s `SyncRwLock`.
+struct IpAccessPolicyState {
+    banned_ips_config_all: Vec<BannedIP>,
+    banned_ips: FxHashSet<IpAddr>,
+    shadowbanned_ips: FxHashSet<IpAddr>,
+    white_list_ips: FxHashSet<IpAddr>,
+    // less cpu cycles then calling vec.len() every time in loop
+    has_whitelist: bool,
```
coco875: no sure about this "optimisation" because it's only on whitelist. I have made some basic search and found https://stackoverflow.com/questions/72087040/computational-complexity-of-hashsetlen-in-rust and to sum up len just ruturn a number so very low the overhead. So it would remove all assignation to state.has_whitelist and read of it, that I think is simpler.

coco875 (commented):
just small comment

steel-core/src/network/ip_access_policy.rs
JunkyDeveloper: removed

steel-core/src/network/ip_access_policy.rs
JunkyDeveloper: I will let this comment open for him

steel-core/src/command/arguments/duration.rs
JunkyDeveloper: fixed? 03ae5058

steel-core/src/network/ip_access_policy.rs
JunkyDeveloper: found out, there is a command names cli params like ./steel convert banned-ip which converts the minecraft to steel format XD thanks for codex to point that out :D

coco875 (approved):
LGTM

4lve:
Thanks for the PR. I like the idea with the steel CLI. But after reviewing it I think the PR is trying to solve too many things at once and has some foundation issues we should not merge as-is

I’m going to close this PR for now instead of asking for a large rewrite. The work is still useful as a reference!

JunkyDeveloper:
discussion what is missing: https://discordapp.com/channels/1428487339759370322/1428487341093421139/1519021214050422906

PR #173, opened 2026-05-02
Stub Implementation of Game Events

CuzImClicks:
this is a stub implementation of game events that can be used by everyone while the game event listener registry is being built. in axe.rs you can see how they would be used


4lve:
Thanks, I went ahead and finished the implementation side so it wouldn't be stuck as a stub implementation forever

PR #175, opened 2026-05-06
Recipe Property Sets (again)

CuzImClicks:
this time you can add new items to the sets before the freeze happens


steel-registry/build_assets/recipe_property_sets.json
coco875, line 1: it's from the extractor ? else I think it's now datapack driven no ?

CuzImClicks: Im reworking recipe property sets again anyways

CuzImClicks:
recipe property sets will be actually derived instead of extracted

PR #181, opened 2026-05-19
remove unwanted "literal_suffixes"

unschlagbar:
This might be one of the most pointless changes ever proposed. A random signal in my brain told me to make the code a tiny amount cleaner by wasting an hour removing _u64, _i64, _u16, _u8, and _f64 suffixes nobody asked for. The good side: i started to bring the steel code structure back into my head. I was careful to not remove a suffix that would change the type


steel-core/src/behavior/blocks/vegetation/cactus_block.rs
```diff
@@ -135,7 +135,7 @@ impl BlockBehavior for CactusBlock {
         }
 
         // Count cactus blocks below
-        let mut height = 1u32;
+        let mut height = 1;
```
4lve, line 138: All cases like this should still contain the type information. Do `let mut height: u32 = 1`

unschlagbar:
i cant resolve conflicts, i hate this fu**ing git system

PR #184, opened 2026-05-20
remove cpu=native flag to let steel run on everything

coco875:


4lve:
Thanks, this would remove it from local dev builds as well though. I made a commit that would only target the github action.

coco875:
why not revome from local build because it modify benchmark number and make problem if we want to do cross compilation ?

PR #187, opened 2026-05-21
refactor math related function into his own crate

coco875:


coco875:
conflict have been solve

PR #188, opened 2026-05-23
Dockerfile

JunkyDeveloper:
if anyone is interested
@tangjuyo please check or if you say it isn't I will close, for me that works

Added also a github workflow, which builds every commit on master.
It is currently deactivated but everything is prepared to push to github registry.
every commit on master will be pushed with the nightly tag and fixed version for each release.
so latest, is not used in this configuration, because docker uses by default latest and minecraft server don't want to update all the time (or as I do at my work with other software) because they need to port extensions, etc to the newest version first and test it.

For people how experiment is the nightly tag, means they get always the newest commit from master.


JunkyDeveloper:
Will close and merge with pelican

PR #191, opened 2026-05-28
feat: modern knockback (pvp)

tn-lorenz:
This pr aims to implement entity knockback in accordance to modern combat mechanics. Custom and classic combat behaviour will follow in a later pr.

# Todo
- [ ] broadcast packet
- [ ] probably some other stuff, which I already forgot :)


steel-core/src/entity/mod.rs
```diff
@@ -546,3 +546,77 @@ pub trait LivingEntity: Entity {
         }
     }
 }
+
+/// A trait containing combat-related functions, based on the `LivingEntity` trait.
+pub trait Attackable: LivingEntity {
```
unschlagbar, line 552: do this `pub trait Attackable: LivingEntity + Sized`

steel-core/src/entity/mod.rs
```diff
@@ -546,3 +546,77 @@ pub trait LivingEntity: Entity {
...
+        let new_vel = DVec3::new(old_vel.x * 0.6, old_vel.y * 1.0, old_vel.z * 0.6);
+
+        self.set_velocity(new_vel);
+    }
+
+    /// Blocks an attack
+    fn block_using_item(&self, attacker: &dyn Attackable)
+    where
+        Self: Sized,
+    {
```
unschlagbar, line 611: So you can remove this ugly `where`

tn-lorenz:
quick question: is `&Self` better than checking for `&dyn Attackable`?

tn-lorenz:
@unschlagbar 

Okay nvm I need to revert changes, I just remembered why I didn't add `+ sized` to the trait impl, obviously we want arbitrary pairs of entities to be able to attack each other, so I want to use `&dyn Attackable`, but making the trait `+ sized` renders it not `dyn` compatible anymore, the previous version is better

unschlagbar:
If it doesn't work otherwise its fine with where. do it how it works and if you done, i can check if i can make it better.

4lve:
Thanks, attacking/pvp will be coming as a part of the entity refactor though. After that has been merged you could still look into doing the 1.8 pvp/profiles we dicussed in discord.

PR #193, opened 2026-05-28
fix @DarkMrMelther bucket bug

JunkyDeveloper:
bucket behaviour more like vanilla


DarkMrMelther:
Reviewed and tested, now it works like vanilla thx! :100:

4lve:
Thanks, but as we discussed in discord this is more of a temporary workaround, and I wouldn't like to have those in steel.

PR #194, opened 2026-05-29
Feature/fall damage

Joniii11:
Adds vanilla style fall damage handling for players. Including fall damage attributes, gamerule checks, water fall handling stuff, game events for landing on stuff like farmland, and block hooks :O

Also adds behavior for slime, bed, honey, farmland, turtle eggs, owder snow.
Also add world rng (but only implement for fall damage related stuff)


4lve:
Thanks for your PR! However I implemented fall damage in #200 as a part of refactoring all of entity physics. I'm still thankful for your PR though! I talked to you on discord and you said you might of have implemented some blocks I missed, if that is the case feel free to open a new PR.

PR #195, opened 2026-05-31
Feature/falling blocks

tangjuyo:


steel-core/src/behavior/blocks/container/anvil_block.rs
```diff
@@ -0,0 +1,132 @@
...
+    #[must_use]
+    pub const fn new(block: BlockRef) -> Self {
+        Self { block }
+    }
+
+    /// Returns the next damage level block state: ANVIL → CHIPPED → DAMAGED → None (destroyed).
+    #[must_use]
+    pub fn damage(state: BlockStateId) -> Option<BlockStateId> {
+        let block = state.get_block();
+        let facing = state.try_get_value(&BlockStateProperties::HORIZONTAL_FACING);
```
JunkyDeveloper, line 43: Please add a pub const for Horizontal_facing like this:
```
const FACING: EnumProperty<Direction> = BlockStateProperties::HORIZONTAL_FACING;
```
To be more consistent over all blocks, so easier to maintain

steel-core/src/behavior/blocks/container/anvil_block.rs
```diff
@@ -0,0 +1,132 @@
...
+        let with_facing = |new_state: BlockStateId| match facing {
+            Some(f) => new_state.set_value(&BlockStateProperties::HORIZONTAL_FACING, f),
```
JunkyDeveloper, line 45: Use const also here

steel-core/src/behavior/blocks/container/anvil_block.rs
```diff
@@ -0,0 +1,132 @@
...
+        } else {
+            None
+        }
+    }
+}
+
+impl BlockBehavior for AnvilBlock {
+    fn get_state_for_placement(&self, context: &BlockPlaceContext<'_>) -> Option<BlockStateId> {
+        Some(self.block.default_state().set_value(
+            &BlockStateProperties::HORIZONTAL_FACING,
```
JunkyDeveloper, line 61: Use const also here

steel-core/src/behavior/block.rs
```diff
@@ -498,6 +499,74 @@ pub trait BlockBehavior: Send + Sync {
...
+            direct_entity_id: Some(direct_entity_id),
+            causing_entity_id: None,
+            source_position: None,
+        }
+    }
+
+    /// Returns true if this block is a concrete powder variant.
+    fn is_concrete_powder(&self) -> bool {
+        false
+    }
```
JunkyDeveloper, line 546: Only for a discussion, why is this needed like this?
Does no tag exists, which you can check? Because otherwise, we can start doing it for every Block 😅

JunkyDeveloper:
Does damage should work?

https://github.com/user-attachments/assets/11755714-49b0-4b07-8d91-aa7ad83b2deb

JunkyDeveloper (changes requested):

4lve:
Thanks for the PR! Closing this for now though because it introduces a large entity/block-behavior foundation while leaving several vanilla-critical pieces incomplete.

I checked this against local vanilla FallingBlockEntity/FallingBlock/ConcretePowderBlock/AnvilBlock for the target version. The branch leaves unresolved or TODO behavior for portals, applyEffectsFromBlocks, DirectionalPlaceContext replacement checks, block entity NBT restoration, living-entity filtering for fall damage, and proper BlockState codec persistence. It also currently has merge conflicts in core entity/world/behavior files.

PR #197, opened 2026-06-01
refactor around noise and micro optimise

coco875:
don't have done number measurement yet


coco875:
early test show a regression of 10%... will done some profiling

steel-worldgen/src/noise/improved_noise.rs
```diff
@@ -173,29 +145,29 @@ impl ImprovedNoise {
         // Get permutation indices for the 8 corners
         let x0 = self.p(x);
         let x1 = self.p(x + 1);
-        let xy00 = self.p(x0 as i32 + y);
-        let xy01 = self.p(x0 as i32 + y + 1);
-        let xy10 = self.p(x1 as i32 + y);
-        let xy11 = self.p(x1 as i32 + y + 1);
+        let xy = [
```
coco875: ... not the right order

steel-worldgen/src/noise/improved_noise.rs
```diff
@@ -173,29 +145,29 @@ impl ImprovedNoise {
...
 
         // Apply smoothstep interpolation
         let x_alpha = smoothstep(xr);
         let y_alpha = smoothstep(yr_original);
         let z_alpha = smoothstep(zr);
 
-        lerp3(
-            x_alpha, y_alpha, z_alpha, d000, d100, d010, d110, d001, d101, d011, d111,
-        )
+        lerp3_simd(x_alpha, y_alpha, z_alpha, d0, d1)
```
coco875: From what I test lerp3_simd are slower then just lerp3

steel-worldgen/src/noise/improved_noise.rs
```diff
@@ -335,103 +309,82 @@ impl ImprovedNoise {
...
         // Gradient dot products at each corner
-        let d000 = grad_dot(h000, xr, yr, zr);
-        let d100 = grad_dot(h100, xr - 1.0, yr, zr);
-        let d010 = grad_dot(h010, xr, yr - 1.0, zr);
-        let d110 = grad_dot(h110, xr - 1.0, yr - 1.0, zr);
-        let d001 = grad_dot(h001, xr, yr, zr - 1.0);
-        let d101 = grad_dot(h101, xr - 1.0, yr, zr - 1.0);
-        let d011 = grad_dot(h011, xr, yr - 1.0, zr - 1.0);
-        let d111 = grad_dot(h111, xr - 1.0, yr - 1.0, zr - 1.0);
+        let d0 = grad_dot_4x(h0, x4, y4, z4);
```
coco875: Also not the correct order

steel-worldgen/src/noise/improved_noise.rs
```diff
@@ -335,103 +309,82 @@ impl ImprovedNoise {
...
+        let bx = bx1 - bx0;
+
+        let cx1 = d1;
+        let cx0 = d0;
+        let cx = cx1 - cx0;
+
+        let x00 = DVec3::new(bx[0b00], ax[0b00], cx[0b00]);
+        let x10 = DVec3::new(bx[0b10], ax[0b10], cx[0b10]);
+        let x01 = DVec3::new(bx[0b01], ax[0b01], cx[0b01]);
+        let x11 = DVec3::new(bx[0b11], ax[0b11], cx[0b11]);
```
coco875: Not sure about that

steel-worldgen/src/noise/improved_noise.rs
coco875, line 257: https://godbolt.org/z/fYr6GThab less code in asm

steel-math/src/noise_math/grad_dot.rs
```diff
@@ -1,29 +1,55 @@
...
-#[expect(clippy::inline_always, reason = "hot-path noise primitive")]
-#[inline(always)]
+/// Gather gradient components for 8 hashes into separate x/y/z SIMD vectors of size 8,
+/// then compute the dot product with the given position vectors.
+#[inline]
 #[must_use]
-pub fn grad_dot(hash: usize, x: f64, y: f64, z: f64) -> f64 {
-    let g = &GRADIENT[hash & 15];
-    g[0] * x + g[1] * y + g[2] * z
+pub fn grad_dot_8x(hashes: [usize; 8], x: f64x8, y: f64x8, z: f64x8) -> f64x8 {
```
coco875: Not use in the final code so can remove it

coco875:
I can't confirm performance because of noise around number but should be slightly better and have mostly rework code around for less cast (i32 to f64) and make some generic function to allow generic of simd

coco875:
(a number that appears frequently in benchmark are around 2% faster but not sure)

coco875:
I try pregen 130 chunk and I get 981.7 with this branch instead of 894.7 with master

JunkyDeveloper:
Why did you renamed floor to fast_floor?

coco875:
normally I also update the description, but because it's a function from https://cgvr.cs.uni-bremen.de/teaching/cg_literatur/simplexnoise.pdf who are call fastfloor 
image

steel-math/src/noise_math/fast_floor.rs
```diff
@@ -0,0 +1,79 @@
...
+pub fn fast_floor(v: f64) -> i32 {
+    let i = v as i32;
+    if v < f64::from(i) { i - 1 } else { i }
+}
+
+/// SIMD implementation of `fast_floor`.
+#[expect(clippy::inline_always, reason = "hot-path noise primitive")]
+#[inline(always)]
+#[must_use]
+pub fn fast_floor_simd<F, I, const N: usize>(v: Simd<F, N>) -> Simd<I, N>
```
JunkyDeveloper, line 30: Looks like this is never used or at least on my phone can't find this function in this PR

steel-math/src/noise_math/fast_floor.rs
```diff
@@ -0,0 +1,79 @@
...
+    let i = v.cast::<I>();
+    let b = v.simd_lt(i.cast::<F>());
+    b.select(i - Simd::splat(1).cast(), i)
+}
+
+/// A 3D vector implementation of `fast_floor` using `DVec3` and `IVec3`.
+#[expect(clippy::inline_always, reason = "hot-path noise primitive")]
+#[inline(always)]
+#[must_use]
+pub fn fast_floor_3x(v: DVec3) -> IVec3 {
```
JunkyDeveloper: Same for here unused complex code I would personally put it in here

JunkyDeveloper (changes requested):
Longer review I can do on the Weekend

steel-math/src/noise_math/fast_floor.rs
coco875: it's used in improved noise, noise_simd

steel-math/src/noise_math/fast_floor.rs
coco875: yes this one are not used, hmm I will remove it same for other 3x unused

4lve:
Thanks for the PR but in my testing it doesn't show any speedup
image

coco875:
with native or not ? and there is other thing not related to optimisation in the work case I can revert only optimisation but rename, remove of code duplication, can benefit

4lve:
> with native or not ? and there is other thing not related to optimisation in the work case I can revert only optimisation but rename, remove of code duplication, can benefit

The testing was with native. You can reopen the PR and rebrand it more as a cleanup rather than a speed improvment

coco875:
it already don't have the name speed improvement ? 
image

coco875:
the two optimisation are just use u8 for index and avoid cast in generated file

coco875:
(will try revert transpile.rs)

4lve:
Ah okay. I think the build fails on run-native as well but might be due to merge

coco875:
I can't find why it's slower, I have try to revert transpiler but don't look like to change.

PR #198, opened 2026-06-02
workaround for broken sounds on buttons

stuxt-og:
For some reason, when the buttons turning off, they play:
wooden_button(oak, birch, etc) -> wooden_pressure_plate
warped_button -> wooden_pressure_plate
crimson_button -> wooden_pressure_plate
stone_button plays block.stone.break
blackstone_button plays block.stone.break
warped_button & crimson_button -> nether_wood_button.click_on (previously it would randomly play one of the pressure plate sounds too)
tried to find cause in server & extractor code but nothing

vid with this bug: https://drive.google.com/file/d/1oJjsXu_7zi2_HkAXOVm2Iovlrr_4GzCc/view?usp=sharing

this commit has workaround only for wooden_button(oak, birch, etc), couldn't find sounds for others


JunkyDeveloper:
Is resolved

PR #199, opened 2026-06-04
remove validation in creation

JunkyDeveloper:
it did the validation twice, so removed in the creation methods.
Tested with broken configs:

```
[domains.empty]
default = false
storage.type = "steel:ram"

[[domains.empty.worlds]]
name = "empty"
default = true
generator = "steel:empty"
```

```
[domains.flat]
default_gamemode = "creative"
storage.type = "steel:ram"

[[domains.flat.worlds]]
name = "overworld"
generator = "minecraft:flat"
default = true

[[domains.flat.worlds]]
name = "the_nether"
generator = "minecraft:flat"
config.dimension_type = "minecraft:the_nether"
config.layers = []
```

In both cases the validation of config, is still active and gives feedback of these errors.
While config resolving the config parameter already get's validated here:
```
fn resolve_world_config(
    domain_name: &str,
    world: &WorldEntryConfig,
    domain_defaults: DomainWorldDefaults,
    domain_storage: &StorageSelection,
    generator_registry: &WorldGeneratorRegistry,
    storage_registry: &WorldStorageRegistry,
) -> Result<ResolvedWorldConfig, String> {
    validate_world_name(&world.name, domain_name)?;
    let world_key = Identifier::new(domain_name.to_owned(), world.name.clone());

    let generator_config = world
        .config
        .clone()
        .unwrap_or_else(|| toml::Value::Table(Map::new()));
    generator_registry.validate_config(&world.generator, &generator_config)?;
```
https://github.com/Steel-Foundation/SteelMC/blob/f58bb6ad2e53ef685544a413ebfc4b1886a823ad/steel-core/src/config.rs#L448


steel-core/src/worldgen/registry.rs
```diff
@@ -237,8 +237,8 @@ fn parse_flat_config(config: &toml::Value) -> Result<FlatGeneratorConfig, String
         .map_err(|e| format!("invalid minecraft:flat config: {e}"))
 }
 
-fn create_overworld(config: &toml::Value, seed: i64) -> Result<GeneratorOutput, String> {
-    validate_empty_config(config)?;
+#[expect(clippy::unnecessary_wraps, reason = "because of flat generator")]
```
coco875: personally I would like to have the clippy comment and the result in output by just make a lamda function like `|_, seed| Ok(create_overworld(seed))` I have give you the full patch on discord

PR #201, opened 2026-06-07
Feat docker

### LINKED ISSUE #119: Dockerfile
What about docker?

JunkyDeveloper:
Folder docker:
- has a docker compose for the a normal boot if needed locally
- Dockerfile alpine based for small image size, the image has the size of 38mb

added .dockerignore, to let the image as small as possible

Also a workflow was added which builds docker container for amd64 and arm64.
the newest commit on master will have the tag `nightly`
releases will have the tag from github

removed eggs, and started process at pelican

Closes #119


docker/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM rustlang/rust:nightly-alpine3.23 AS builder
+LABEL authors="junkydeveloper"
+
+WORKDIR /steel
+
+RUN apk add --no-cache lld
+
+ENV RUSTFLAGS="-Zthreads=0 -Clink-arg=-fuse-ld=lld"
```
coco875: no longer need it with latest commit

JunkyDeveloper: removed

pelican/egg-steel-mc.json
coco875, line 1: we need to have this file ?

JunkyDeveloper: removed

coco875 (approved):
look fine to me

JunkyDeveloper:
need to merge master + add stdin for the egg and add both correct eggs (need to change the egg install script)

PR #204, opened 2026-06-09
Pin TextComponents revision to avoid breaking Steel on lib update

DarkMrMelther:
Auto explanatory mini PR


Cargo.toml
```diff
@@ -114,11 +114,15 @@ md5 = "0.8.0"
...
 hex = "0.4.3"
 rsa-der = "0.3.0"
-text_components = { git = "https://github.com/Steel-Foundation/TextComponents.git", features = [
-    "custom",
-    "nbt",
-    "serde",
-] }
+text_components = { 
+    git = "https://github.com/Steel-Foundation/TextComponents.git", 
+    rev = "e058b936e66b496385ba162b2e20344d14572a6a",
```
JunkyDeveloper, line 119: Why not using git tags? for me that makes a bit more sense, to see what changed, etc and what that jump was

JunkyDeveloper (changes requested):
Small question

Cargo.toml
DarkMrMelther: 4lve wanted the git directly due to the library being in early development, but tags automatically uploads the update to crates.io, so essentially would be the same use the crates.io version and the tags, that's why per revision it's preferred rn

JunkyDeveloper: ok sounds good

JunkyDeveloper (approved):

JunkyDeveloper:
Got now a LGTM state

PR #205, opened 2026-06-11
26.2

JunkyDeveloper:
WIP but sulfur caves are already working


Joniii11:
update: 100% vanilla worldgen already

package-content/config.schema.json
```diff
@@ -76,14 +76,14 @@
           "description": "Vanilla chat spam threshold window in seconds; values <= 0 disable throttling",
           "minimum": -2147483648,
           "maximum": 2147483647,
-          "default": -1
+          "default": 10
```
JunkyDeveloper, line 79: @4lve why did you changed from -1 to 10?
the idea was to disable it by default, so to have an opt-tin, if you really want to do that (I don't like that feature)

package-content/config.schema.json
```diff
@@ -76,14 +76,14 @@
...
         },
         "command_spam_threshold_seconds": {
           "type": "integer",
           "description": "Vanilla command spam threshold window in seconds; values <= 0 disable throttling",
           "minimum": -2147483648,
           "maximum": 2147483647,
-          "default": -1
+          "default": 10
```
JunkyDeveloper, line 86: same here

package-content/config.schema.json
4lve: That's the vanilla default

JunkyDeveloper: yeah, but do you really like to have that enabled

PR #206, opened 2026-06-11
no buitin_datapack in the repo, fetch it instead

coco875:
instead of having the minecraft datapack in the repo that can be considered like mojang properties, fetch it from their api. It only download it if it wasn't here. If the wrong version are present it redownload it. It don't download at each build


steel-core/src/behavior/blocks/building/slab_block.rs
```diff
@@ -149,10 +149,7 @@ impl BlockBehavior for WeatheringCopperSlabBlock {
 
 #[cfg(test)]
 mod tests {
-    use steel_registry::{
-        blocks::block_state_ext::BlockStateExt as _, test_support::init_test_registry,
-        vanilla_blocks,
-    };
+    use steel_registry::{test_support::init_test_registry, vanilla_blocks};
```
JunkyDeveloper: Can you extract that please to another PR, so this can be merged asap and I can merge that PR also in my PRs so not much merge conflicts should result

steel-core/src/behavior/blocks/building/stair_block.rs
```diff
@@ -194,7 +194,6 @@ impl BlockBehavior for WeatheringCopperStairBlock {
 
 #[cfg(test)]
 mod tests {
-    use steel_registry::blocks::block_state_ext::BlockStateExt as _;
```
JunkyDeveloper: same here

JunkyDeveloper (changes requested):
please extra PR

steel-core/src/behavior/blocks/building/slab_block.rs
coco875: done

steel-core/src/behavior/blocks/building/stair_block.rs
coco875: done

JunkyDeveloper (changes requested):
To much files to mark the correct line in density_function.rs you change the path.
so for the steel-registry you create a symlink and in worldgen you do a relative path.

maybe do at both places the same thing, so do symlink or the relative path. I think the relative path is the best, because you can remove the symlink creation which is OS specific.

coco875:
will change path, like you mention symlink are os depend and I know some filesystem (like exFAT) don't support symlink at all.

JunkyDeveloper (approved):
LGTM all other code, and I think that will help to reduce repo size in the future. Maybe later even rewrite the histroy to remove all of that stuff

PR #208, opened 2026-06-12
fix lint

coco875:
I guess because of an update of clippy or rust nightly now there is a error on pr, so this pr fix it


coco875:
an example of pr ci who fail because of it https://github.com/Steel-Foundation/SteelMC/actions/runs/27222773322/job/80381719972?pr=204

PR #209, opened 2026-06-13
Delete damage_types_tags.rs

coco875:


JunkyDeveloper (changes requested):
Delete that file but rename the damage_type_tags to damage_types_tags, so it has the same name schema of damage_types.

Or what do you think?

coco875:
it's your pr to port to 26.1 that add it https://diffshub.com/Steel-Foundation/SteelMC/pull/96 but don't reference so I guess an artifact of the port ?

JunkyDeveloper:
Always the intern ....
Yeah I think its good to be deleted :D

JunkyDeveloper (approved):
Interns fault

PR #210, opened 2026-06-14
AABB refactor & More vectors

suprohub:
AABB fully refactored for having abstract fields. Some (i32, i32, i32) and [i32; 3] replaced to IVec3


steel-core/src/entity/movement_sync.rs
```diff
@@ -221,8 +221,8 @@ impl EntityMovementSyncPackets {
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct EntityPositionSyncSnapshot {
     entity_id: i32,
-    position: DVec3,
-    velocity: DVec3,
+    pos: DVec3,
+    vel: DVec3,
```
JunkyDeveloper: Only want to highlight, would like use the ful name and only vel or pos at variables.
What was the result of that discussion?

coco875: we have more or less conclude to keep full name in structure, and in worst case, don't touch more then necessary so I agree with junky here

JunkyDeveloper: So request a change?

coco875: yes to revert back to orignal

JunkyDeveloper: revert back please see above thread

JunkyDeveloper (changes requested):
At least here

steel-core/src/entity/movement_sync.rs
suprohub: agree

suprohub: ok

PR #212, opened 2026-06-14
Adds sponge

Joniii11:
As you can see the sponge is sponging :D

and in the nether it dries up and becomes a normal sponge when you place it as a wet sponge


steel-core/src/behavior/blocks/building/wet_sponge_block.rs
```diff
@@ -0,0 +1,60 @@
...
+use steel_macros::block_behavior;
+use steel_protocol::packets::game::SoundSource;
+use steel_registry::blocks::BlockRef;
+use steel_registry::{sound_events, vanilla_blocks};
+use steel_utils::{BlockPos, BlockStateId, types::UpdateFlags};
+
+use crate::behavior::{BlockBehavior, BlockPlaceContext};
+use crate::world::World;
+
+const PARTICLES_WATER_EVAPORATING: i32 = 2009;
```
JunkyDeveloper: Do we not have a registry, or something there all particales are saved, because that are pain in the ass to find things like this

JunkyDeveloper (changes requested):

steel-core/src/behavior/blocks/building/wet_sponge_block.rs
Joniii11: oh yeah lol

coco875 (approved):
look good to me, there is just the CI of test to re-run because of some weird "error sending request for url"

PR #213, opened 2026-06-14
Trapdoors & Budding Amethyst

DarkMrMelther:
Implemented all types of trapdoors:
- TrapDoorBlock
- WeatheringCopperTrapDoorBlock

Also implemented BuddingAmethyst & AmethystCluster

Tested with 2000 tick rate, both weathering and amethyst grow works fine


steel-core/src/behavior/blocks/vegetation/carpet_block.rs
```diff
@@ -1,15 +1,15 @@
 use steel_macros::block_behavior;
```
JunkyDeveloper, line 1: Why have the carpet in vegetation and not in decoration blocks, which I think is a better place

coco875: maybe it include also moss carpet

steel-core/src/behavior/blocks/building/trapdoor_block.rs
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    fn is_randomly_ticking(&self, _state: BlockStateId) -> bool {
+        self.weathering.is_randomly_ticking()
+    }
+
+    fn random_tick(&self, state: BlockStateId, world: &Arc<World>, pos: BlockPos) {
+        self.weathering.change_over_time(state, world, pos);
+    }
+    // TODO: Implement WeatheringCopper::get_age
```
JunkyDeveloper: I thought weathering copper is fully done?

steel-core/src/behavior/blocks/vegetation/carpet_block.rs
JunkyDeveloper: ohhh yeah that can make sense.
But if I search for carpet I would never search in vegetation. but moss carpet makes sense

JunkyDeveloper:
And fix lint :D

steel-core/src/behavior/blocks/vegetation/carpet_block.rs
DarkMrMelther: I can confirm is because of moss carpets, also I only finished the implementation

steel-core/src/behavior/blocks/building/trapdoor_block.rs
DarkMrMelther: Maybe it's a block specific function then, but sounds weird to me

JunkyDeveloper: As I remember you get the age or more the weather state in the weather block

DarkMrMelther: I've made the `weathering` value public, it should be the same as implementing get_age

4lve:
Thanks for the PR! I also noted that you incorrectly used clicked_pos instead of relative_pos.
This was an easy mistake to make in steel since in vanilla
BlockPlaceContext.getClickedPos() overrides UseOnContext.getClickedPos() making it so it uses relative_pos under the hood.

Therefore I went ahead and made a small refactor renaming stuff and clearing them up for the future

PR #215, opened 2026-06-17
JJThunder Datapack

JunkyDeveloper:
This is only a draft PR of the PoC for JJThunder to the Max and it will be always a draft or closed for native datapack support

Version is 26.1 because sulfur caves are not added in 26.2 and JJthunder needs to update his datapack on his own.
Maybe some changes can be used for the master, to make datapack support better (ignore the json)

If you want to try out JJThunder datapack, clone my branch and do:

1. Compile steel
2. it will download in steel-utils in build_assets/builtin_datapacks the minecraft dataset with the name minecraft
3. please delete the folder
4. download this [minecraft_my_version.zip](https://github.com/user-attachments/files/29025772/minecraft_my_version.zip) zip which has the minecraft folder in it and put it at the same position, as the previous minecraft folder.
5. compile steel in release mode
6. start steel and enjoy


JunkyDeveloper:
I will close this PR in favor of cocos work which is better implemented

PR #216, opened 2026-06-17
Fixes and improvements to the CLI

DarkMrMelther:
Added sending handlers through New Space character and Ctrl+T to allow other systems like Pelican use it nicely. 
Reworked how the user input is handled to make it appear in one line only, this solves many bugs related to the input going out of screen. 
Made the command auto completion adapt to the terminal height. 
Added file logging with file rotation
Added more options to the log config:
- To change the log directory path (Now defaults to `./logs`)
- To change the logging level 
- To enabling/disabling the file logging
- To change the amount of time of log rotation
- To change the maximum amount of input history items

Changed the panic logic to work with the logger

I've tested it myself, but more testing is appreciated


JunkyDeveloper:
Melther and I worked on the Ctrl+t together, so i tested until fast push and it works via cmd and egg console. So that is a huge improvement

But didn't made a code review, can do that later today (because I really want to have it merged)

PR #221, opened 2026-06-20
Implement rooted dirt and simple vegetation block behaviors

Joniii11:
title


JunkyDeveloper:
Please add in the PR description, which blocks you did and which functionality is still missing from each block :D

PR #222, opened 2026-06-21
More config

coco875:


JunkyDeveloper:
Please also link here the PR for Docs where you add the new config and explain which mod is for that necessary!

coco875:
I add config in config.toml, and for the client I use https://modrinth.com/mod/c2me-fabric to allow render distance above 32, but other mod like bobby and any other

JunkyDeveloper (changes requested):
Now I request that Docs PR ;)
because why should I limit the thread pool? and which number is recommended for which system like a 2 core VPS
Otherwise the PR lgtm

coco875:
you should be more happy now https://github.com/Steel-Foundation/SteelDocs/pull/30 x)

4lve:
Thanks for the PR. I modified the wording of the allow_extended_view_distance a bit and I also made it so you can configure the amount of threads per thread pool

coco875:
that also mean that https://github.com/Steel-Foundation/SteelDocs/pull/30 need to be update

PR #223, opened 2026-06-21
chore: make the script automatically choose the second lastest ver to fetch

chocodev11:


script/update-minecraft-src.sh
coco875, line 1: hmm instead of latest version it would be possible to fetch the version that steel target ?

coco875: (so here 26.2 but if somone want to backport or that steel don't update fast enough don't have the incorrect version)

JunkyDeveloper: @coco875 @4lve said he want's to have at least a diff of the current version and 1 version before, so we need always 2 versions.

chocodev11: actually it is 1 version behind the lastest release. using 26.1 for now.

JunkyDeveloper:
before approval, I it needs to be tested, I don't have time for that, sorry

GreenedDev:
it works for me

4lve:
Thanks for the PR!

I see trap was removed which was quite useful cause it still removed the folder when the script crashed mid run or got cancelled.

I also see you removed `--refresh-min-version` which is required so the final minecraft-src contains the correct code. I learned that the hard way when a movement validation check missed an if check and I had to decompile the actual server jar to check the difference.

I'm also not sure about the reaching out to piston-meta to fetch the current version as it seems like a bit of over-engineering but I haven't fully decided on that one yet. Maybe an arg that overrides the minimum version would be better?

Also note that you should try to prevent using branch names like master, dev. That makes it so I can't commit into the PR, so for that reason I'm closing the PR so you can open a new one

PR #224, opened 2026-06-21
remove UseOnContext and UseItemContext

coco875:


Joniii11:
but why remove that context, i kinda like it more with context

coco875:
because it's just a wrapper of var at the end (it can be tuple that nothing will change) and compare to block behavior we don't have that kind of thing. (and personally I prever without because shorter line and more direct access to kind of native type)

Joniii11:
hm, ok
LGTM

Joniii11 (approved):

CuzImClicks:
im actually against this for several reasons.
when you add an argument you now have to change 30 files because of the function signatures. 

i also dont like adding and removing the underscores all the time. 

additionally function signatures take way more lines when everything is spelled out.

```rust
fn use_on(&self, context: &mut UseOnContext) -> InteractionResult {}

// ---

fn use_on(
    &self,
    _player: &Player,
    _hand: InteractionHand,
    hit_result: BlockHitResult,
    world: &Arc<World>,
    inv: &mut InventoryAccess,
) -> InteractionResult {}
```

in my opinion the second one just looks cluttered and less readable

4lve:
Yea, as @CuzImClicks states that were some of the reasons the context was chosen to begin with. Will close this but feel free to discuss in the discord

coco875:
would be good to have an issue to track that ? (matter mostly if we want api uniform)

PR #225, opened 2026-06-22
Entity refactor part 2

4lve:
This refactor makes entity registrations more ergonomic like blocks and items.
It also changes the ownership of entities from being chunk-owned to now being owned by a seperate manager.
It also adds entity AI foundation with the pig entity having all its goals implemented


JunkyDeveloper:
Please fix lint

PR #231, opened 2026-06-25
fix rsa public key encoding

Purdze:
server sent the rsa modulus as a negative integer, breaking client encryption. use the proper spki encoder.


JunkyDeveloper (approved):
Trust you and looks good and you proofed it works

Joniii11 (approved):
lgtm

PR #232, opened 2026-06-25
Feat sleeping, bed with respawn point, anchor thingy with priming

Joniii11:
yeah,

I hope its not bad if i have 223 commits lol cause i was based on entity refactor branch


steel-core/src/behavior/blocks/building/bed_block.rs
```diff
@@ -2,29 +2,43 @@ use std::sync::Arc;
...
+/// TODO: Mirror vanilla `BedBlock.kickVillagerOutOfBed` once villager sleeping
+/// entities exist.
 #[block_behavior]
 pub struct BedBlock {
     block: BlockRef,
 }
 
 impl BedBlock {
+    const HEAD_PLACE_FLAGS: UpdateFlags = UpdateFlags::UPDATE_ALL;
+    const UPDATE_NEIGHBOR_SHAPES_LIMIT: i32 = 512;
```
coco875: we don't have something global for this ?

steel-core/src/behavior/blocks/building/bed_block.rs
```diff
@@ -49,11 +63,204 @@ impl BedBlock {
...
+                Self::find_standup_position_at_offset(world, entity, block_pos, &offsets, true)
+            {
+                return Some(safe_pos);
+            }
+
+            Self::find_standup_position_at_offset(world, entity, block_pos, &offsets, false)
+        }
+    }
+
+    fn find_bunk_bed_standup_position(
```
coco875, line 132: the function can't be simplified as just for loop with [-1, 1] ?

steel-core/src/behavior/blocks/decoration/banner_block.rs
coco875, line 1: why did you implement banner ?

steel-core/src/behavior/blocks/portal/respawn_anchor_block.rs
```diff
@@ -1,37 +1,187 @@
...
+            pos.south().east(),
+        ]
+    }
+
+    fn has_charge(state: BlockStateId) -> bool {
+        state.get_value(&BlockStateProperties::RESPAWN_ANCHOR_CHARGES) > 0
+    }
+
+    fn can_be_charged(state: BlockStateId) -> bool {
+        state.get_value(&BlockStateProperties::RESPAWN_ANCHOR_CHARGES) < 4
```
coco875: replace 4 with a const ? (ex: MAX_RESPAWN_ANCHOR_CHARGES)

steel-core/src/entity/dismount_helper.rs
coco875, line 1: your pr add more then bed and anchor ?

steel-core/src/entity/living_entity.rs
```diff
@@ -2586,21 +2586,114 @@ pub trait LivingEntity: Entity {
...
+
+        self.set_pose(EntityPose::Standing);
+        self.clear_sleeping_pos();
+    }
+
     /// Stops the entity from sleeping.
     fn stop_sleeping(&self) {
-        self.clear_sleeping_pos();
+        self.default_stop_sleeping();
     }
```
coco875, line 2783: why not just rename default_stop_sleeping to stop_sleeping ? (so only have one function do stop sleeping)

steel-core/src/player/lifecycle/respawn.rs
```diff
@@ -74,6 +172,54 @@ impl ServerJob for PlayerRespawnJob {
 
         loop {
             match &mut self.phase {
```
coco875, line 190: a technical question does respawn are done on an another thread or the main thread ? (because by memory join before can freeze tick until all chunk are load, so does the same can happen here ?)

steel-core/src/player/lifecycle/respawn.rs
```diff
@@ -207,37 +450,53 @@ impl Player {
...
         }
 
+        let keep_inventory =
+            target_world.get_game_rule(&KEEP_INVENTORY) || self.game_mode() == GameType::Spectator;
+        if keep_inventory {
+            self.inventory
+                .lock()
+                .refresh_all_player_equipment_attribute_modifiers();
+        } else {
+            self.inventory.lock().clear_content();
```
coco875, line 506: there is also the all the item who are drop no ?

coco875 (changes requested):
a first review, I didn't finish to read all will continue tomorrow

steel-core/src/behavior/blocks/decoration/banner_block.rs
Joniii11: for the find nearest respawn point

steel-core/src/entity/dismount_helper.rs
Joniii11: sleeping is also there dismount helper exists for the leaving of the bed

steel-core/src/behavior/blocks/decoration/banner_block.rs
coco875: I still don't get the link outside of can_spawn function

.gitignore
```diff
@@ -21,6 +21,9 @@
 CLAUDE.md
 .env
 /minecraft-src
+/minecraft-src-26.2
+/minecraft-src-26.3-snapshot-1
+/minecraft-src-26.3-snapshot-2
```
coco875: ?

steel-utils/src/direction.rs
```diff
@@ -324,6 +317,18 @@ impl Direction {
         }
     }
 
+    /// Returns whether this direction is facing the given yaw angle
+    #[must_use]
+    pub fn is_facing_angle(self, yaw: f32) -> bool {
```
coco875: maybe is_facing_angle_yaw ? because there is also pitch for up and down

steel-utils/src/direction.rs
```diff
@@ -106,19 +106,12 @@ impl Direction {
...
     pub fn from_yaw(yaw: f32) -> Direction {
-        let adjusted = yaw.rem_euclid(360.0);
-        match adjusted {
-            y if !(45.0..315.0).contains(&y) => Direction::South,
-            y if y < 135.0 => Direction::West,
-            y if y < 225.0 => Direction::North,
+        match ((f64::from(yaw) / 90.0 + 0.5).floor() as i32) & 3 {
+            0 => Direction::South,
+            1 => Direction::West,
+            2 => Direction::North,
```
coco875, line 114: I think that what you have done are less efficient because float division and float_floor that try to be IE754 compliant

steel-core/src/behavior/blocks/building/bed_block.rs
Joniii11: good catch

steel-core/src/behavior/blocks/building/bed_block.rs
Joniii11: good catch x2

steel-core/src/behavior/blocks/decoration/banner_block.rs
Joniii11: its for the spawn search, so like every other block would deny you to spawn in this block. but the banner block overwrites it:
```
    fn is_possible_to_respawn_in_this(&self, _state: BlockStateId) -> bool {
        true
    }
```

steel-core/src/behavior/blocks/portal/respawn_anchor_block.rs
Joniii11: agreed did this, thanks :D

steel-core/src/player/lifecycle/respawn.rs
Joniii11: Inventory drops already happen in Player.die() method through the take_death_drops() before respawning

steel-core/src/entity/living_entity.rs
Joniii11: stop_sleeping is already the overiddable of livingentity trait method. The Player is already overriding it. because of rust has no java style super.stopSleeping (this is what vanilla does) i decided to do it like this. else it would just call itself recursivly

steel-core/src/player/lifecycle/respawn.rs
Joniii11: this is from the ai cause i was to dumb to explain: Respawn state transitions run on the main server thread, but they do not synchronously wait for chunks. `PlayerRespawnJob::poll` checks the chunk request and returns `Pending` while chunks are loading, so processing continues on a later tick. The shared-spawn search also has a per-tick candidate budget. Therefore a respawn should not freeze the tick waiting for all required chunks, unlike the old blocking join path.

coco875 (commented):
mostly question

.gitignore
Joniii11: oh yeah that is because i needed to add the different minecraft sources to compare between them. ima remove that

steel-utils/src/direction.rs
Joniii11: i mean this is how vanilla named it and the docs are sufficient imo

steel-utils/src/direction.rs
Joniii11: I mean but this is how exactly how vanilla does it (also rem_euclid also uses those float division)

steel-utils/src/direction.rs
coco875: personally I prefer if it's even more transparent so only the name of the function are enough for example

steel-utils/src/direction.rs
coco875: I think both are equivelent but https://godbolt.org/z/q68qqGETq it's heavier to do the div then floor then only en euclid then if (that get remove with compiler optimisation)

coco875 (dismissed):

JunkyDeveloper (dismissed):

coco875 (approved):

JunkyDeveloper (approved):

PR #233, opened 2026-06-26
Implement ender pearls

Purdze:
Ender pearls with flight, collision, owner teleport, chunk loading and player bound persistence.


steel-core/src/behavior/items/ender_pearl.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct EnderPearlItem;
+
+impl ItemBehavior for EnderPearlItem {
+    fn use_item(&self, context: &mut UseItemContext) -> InteractionResult {
+        // TODO: enforce the 1-second throw cooldown once an item cooldown system exists.
+        let player = context.player;
+        let world = context.world;
+
+        // Vanilla throw pitch: 0.4 / (random * 0.4 + 0.8).
+        let pitch = 0.4 / (rand::random::<f32>() * 0.4 + 0.8);
```
Joniii11, line 31: why not use world.random().lock()

Joniii11 (changes requested):
i only found this elkse it LGTM

steel-core/src/behavior/items/ender_pearl.rs
coco875: hmm in https://mcsrc.dev/1/26.2/net/minecraft/world/item/EnderpearlItem#L33 it use world random, so do you consider for perfect parity it's necessary or we can neglect that ? (purdze) I know that it's important for speedrun or very specific mechanics that use world random, or we can use any random ?

Joniii11: obviously i would use it lol and if we already have access to world why not use it?

coco875: it can be for performance because lock the world can have a cost

Purdze:
test

PR #237, opened 2026-06-28
multiple micro optimisation

coco875:
with my pc (i7 14700kf) with `PREGEN_SIZE=301 cargo run --release` (NOT NATIVE)
before: 1586.1, 1486.3, 1555.2

after: 1637.5, 1696.8, 1628.8


coco875:
* replace a floor with a fast_floor (so avoid the complex floor of rust and use the simple one as it should in noise)
* use optimisation of structure from https://modrinth.com/mod/structure-layout-optimizer
* have try random thing and found that create a function where it do permission are faster of 5 cycles even if the number of instruction are greater https://godbolt.org/z/4hbv7zPax

GreenedDev:
a whopping 8 cps increase for me.
380 -> 388

coco875:
(at first read I was thinking you mention an increase in time but no the invert x))

PR #241, opened 2026-07-04
fix: flying downward play step sound

chocodev11:


steel-core/src/player/mod.rs
```diff
@@ -1783,6 +1783,14 @@ impl Entity for Player {
         }
     }
 
+    fn movement_emission(&self) -> EntityMovementEmission {
+        if self.is_flying() || self.on_ground() && self.is_discrete() {
```
coco875, line 1787: personally I prefer add parenthesis when we have mulitple operation

coco875 (approved):
look ok, and I assume you have try to confirm that the bug have been fix

Joniii11 (approved):
i like it like that so LGTM

PR #242, opened 2026-07-05
chore: bump typos to 1.48

teknostom:
This makes local typos installations not fail on new typos in upstream.


coco875 (approved):

JunkyDeveloper (approved):
after hard and careful review, I like that change

PR #243, opened 2026-07-06
extend feature of datapack (without breaking parity)

coco875:
This pr add alias that datapack can (datapack can use minecraft:noise and noise for example), modify some type to better corespond to all types that field can accept (like for color using both string and number) and other adjustement.

Why doing this ?
While the vanilla datapack for now don't use all of this, it's possible value that the datapack can evolve to, so doing now avoid to doing later when they do the change. I also have another branch that this time allow to build steel with datapack (only json not mcfunction) https://github.com/coco875/SteelMC/pull/2.


coco875:
image
I have check and it match vanilla, will do a quick self review

JunkyDeveloper:

coco875:
I have update the branch, so will check file that I have modified and then I will make it non draft

steel-registry/build/loot_tables/functions.rs
```diff
@@ -190,31 +222,117 @@ pub(super) fn generate_function(function: &LootFunctionJson) -> TokenStream {
...
         "minecraft:exploration_map" => {
-            let destination = function
-                .destination
-                .as_deref()
-                .unwrap_or("minecraft:buried_treasure");
-            let destination = destination
-                .strip_prefix("minecraft:")
-                .unwrap_or(destination);
-
-            let decoration = function.decoration.as_deref().unwrap_or("minecraft:red_x");
```
coco875, line 206: left from old version

steel-registry/build/loot_tables/values.rs
coco875, line 1: I found that modification on this file are hacky, but it's in the continuity of what we already have, I think that a loot_table pr will fix that but would also require modify steel-core but don't want to do this work on this pr, so do I do something to cleaning up or it will be discarded anyway in a loot table pr so no need to make it clean ?

coco875: at then just revert most modification on loot table to let another pr implement loot table properly

coco875:
ok ready again

PR #245, opened 2026-07-06
fix: bucket duplicated in creative

chocodev11:


4lve:
Thanks for the PR. I added item utils so the fix becomes foundation instead of a localized fix

PR #246, opened 2026-07-06
Implement nether and end portals

4lve:
This makes nether and end portals fully functional. It also allows overriding the target world in worlds.toml, however using the end exit portal always takes you back to your spawnpoint like vanilla

### DIFF
--- a/steel-core/src/behavior/blocks/portal/end_portal_frame_block.rs
+++ b/steel-core/src/behavior/blocks/portal/end_portal_frame_block.rs
@@ -1,13 +1,16 @@
 //! End portal frame block implementation.
 
+use std::sync::Arc;
+
 use steel_macros::block_behavior;
 use steel_registry::blocks::BlockRef;
 use steel_registry::blocks::block_state_ext::BlockStateExt;
 use steel_registry::blocks::properties::BlockStateProperties;
-use steel_utils::BlockStateId;
+use steel_utils::{BlockPos, BlockStateId};
 
 use crate::behavior::block::BlockBehavior;
 use crate::behavior::context::BlockPlaceContext;
+use crate::world::World;
 
 /// Behavior for end portal frame blocks.
 #[block_behavior]
@@ -21,6 +24,10 @@ impl EndPortalFrameBlock {
     pub const fn new(block: BlockRef) -> Self {
         Self { block }
     }
+
+    const fn analog_output_signal(has_eye: bool) -> i32 {
+        if has_eye { 15 } else { 0 }
+    }
 }
 
 impl BlockBehavior for EndPortalFrameBlock {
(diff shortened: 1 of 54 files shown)


coco875:
maybe also add a pr to the doc ?

steel-core/src/behavior/blocks/portal/end_portal_frame_block.rs
```diff
@@ -30,4 +37,43 @@ impl BlockBehavior for EndPortalFrameBlock {
...
+    }
+
+    fn get_analog_output_signal(
+        &self,
+        state: BlockStateId,
+        _world: &Arc<World>,
+        _pos: BlockPos,
+    ) -> i32 {
+        Self::analog_output_signal(state.get_value(&BlockStateProperties::EYE))
+    }
```
coco875, line 52: it's good to add redstone related function without redstone ?

steel-core/src/chunk/chunk_map.rs
```diff
@@ -61,12 +62,17 @@ pub struct ChunkMapGameTickTimings {
...
     pub tickable_count: usize,
     /// Total number of loaded chunks.
     pub total_chunks: usize,
 }
 
+struct TickableChunk {
+    holder: Arc<ChunkHolder>,
+    simulation_level: ChunkTicketLevel,
+}
+
```
coco875, line 75: this can probably conflict with ender pearl pr ? So did you use the same system or make a new one ?

steel-core/src/entity/entities/raw.rs
```diff
@@ -82,6 +84,14 @@ impl Entity for RawEntity {
...
     }
 
+    fn ender_pearl_owner_uuid(&self) -> Option<Uuid> {
+        if self.entity_type != &vanilla_entities::ENDER_PEARL {
+            return None;
+        }
+
+        self.ender_pearl_owner_uuid_from_nbt()
+    }
+
```
coco875: why ? don't look like related to portal directly

steel-core/src/entity/mod.rs
coco875, line 1: this file have more then 4000 lines, would be probably good to split it in another pr

steel-core/src/player/chat_state.rs
```diff
@@ -336,7 +336,7 @@ impl Player {
 
         let update_packet =
             CPlayerInfoUpdate::update_chat_session(self.gameprofile.id, protocol_data);
-        self.get_world().broadcast_to_all(update_packet);
+        self.server().broadcast_to_online(update_packet);
     }
```
coco875, line 340: look a bit out of scope of this pr

steel-core/src/player/mod.rs
coco875, line 1: a bit out of topic but find a bit weird that player aren't classify as entity as mention in discord

steel-core/src/behavior/blocks/portal/end_portal_frame_block.rs
4lve: It's good to fully implement the block imo so once we implement redstone we don't need to do as much

steel-core/src/chunk/chunk_map.rs
4lve: Yea, I'm planning to merge this one before the ender pearl PR and then fix the ender pearl PR. Wanted to do it properly in here

steel-core/src/behavior/blocks/portal/end_portal_frame_block.rs
coco875: ok, it's more that it's hard to test without restone component to confirm like comparator, etc (especially on more complex block like chest, hopper, etc)

4lve: Yea, could make tests for it with flint once we are working on redstone

steel-core/src/entity/mod.rs
4lve: Yea, planning to split it in a future entity refactor

steel-core/src/player/chat_state.rs
4lve: It was required due to the weird state players are in when they are watching the end credits

steel-core/src/player/mod.rs
4lve: Will take another look at it

steel-core/src/player/chat_state.rs
coco875: yes but it's a rename from what I see

4lve: It makes it so all online players are also stored in the server instead of just on the world

steel-core/src/entity/entities/raw.rs
4lve: Will take another look, seems to be a fail safe casuse ender pearls aren't implemented in this PR

coco875: yeah but in the worst case there is no enderpearl entity that can be create

coco875: so the fail safe would be never trigger (only in ender pearl pr who are not merge)

4lve: This is now fixed, it's now a more generic projectile owner with no specific workarounds

steel-core/src/player/mod.rs
4lve: The primary reason is that the player needs some special treatment is that it's not recreated from nbt data when moving across dimensions like non-player entities. It also needs some to get packets and such

coco875: hmm ok, I still think that there is a benefit to make player more align to entity but it goes way beyond the topic of this pr, so I will try on my side

PR #247, opened 2026-07-09
Implement segmentable blocks (leaf litter, pink petals, wildflower)

colonthreeing:
Leaf litter and similar blocks are implemented by this pr, as well as the function can_be_replaced (from vanilla's canBeReplaced in blockbehaviour.java line 831) 

in specific:
- implemented placement rules (rotational placing, increasing stack count when clicked on or when placed in the same block)
  - behaviours are abstracted into a new module called segmentable_blocks matching vanilla.
- can_be_replaced function implemented on the block behaviour trait, still backs up to the config value but is replaceable.
- - needed for this pr specifically because all of these blocks override the behaviour.
- also makes them break when they are not supported and drop the correct material
- bonemealing works
  - bonemeal adds segments til it reaches the max, then creates an item


steel-core/src/behavior/block.rs
```diff
@@ -548,6 +549,25 @@ pub trait BlockBehavior: Send + Sync {
...
+    /// Returns whether this block can be replaced by the held item during placement.
+    ///
+    /// Vanilla parity: `BlockState.canBeReplaced(BlockPlaceContext)`.
+    ///
+    /// Default behavior mirrors prior Steel behavior and only checks the
+    /// block config's `replaceable` flag.
+    #[expect(
+        unused_variables,
+        reason = "default trait implementation ignores context-specific params"
+    )]
```
coco875: If you see above we don't put the clippy warning we just add _ in front of variabel

steel-core/src/behavior/blocks/vegetation/segmentable_block.rs
```diff
@@ -0,0 +1,140 @@
...
+use crate::behavior::{BlockBehavior, BlockPlaceContext, InteractionResult, InventoryAccess};
+use crate::world::{LevelReader, ScheduledTickAccess, World};
+use crate::player::Player;
+
+pub const MAX_SEGMENT_AMOUNT: u8 = 4;
+const FACING_PROPERTY: EnumProperty<Direction> = BlockStateProperties::HORIZONTAL_FACING;
+
+pub trait SegmentableBlock: BlockBehavior {
+    fn block_ref(&self) -> &BlockRef;
+    fn segment_property(&self) -> &IntProperty;
```
coco875: Not a fan of this especially since it's a const

coco875 (changes requested):

coco875 (approved):

PR #249, opened 2026-07-11
Brain system and villager

teknostom:
Add fully working brain system and implement villager.

Some villager stuff is blocked or i just coulndn't care less about them. Should all be in TODOs.


steel-core/src/world/mod.rs
```diff
@@ -4082,6 +3408,59 @@ impl World {
...
+            if let Some(player) = self.players.get_by_entity_id(entity_id) {
+                let player_pos = player.position();
+                let dx = player_pos.x - pos.x;
+                let dy = player_pos.y - pos.y;
+                let dz = player_pos.z - pos.z;
+                if dx * dx + dy * dy + dz * dz <= MAX_DISTANCE_SQ {
+                    player.connection.send_encoded(encoded.clone());
+                }
+            }
+        }
```
coco875: by memory particle can have more information then that and look like to only doing it for level, so maybe add a todo ?

steel-core/src/player/player_inventory.rs
```diff
@@ -954,7 +956,7 @@ impl Player {
     ///
     /// # Arguments
     /// * `provider` - The menu provider containing the title and factory
-    pub fn open_menu(&self, provider: &impl MenuProvider) {
+    pub fn open_menu(&self, provider: &impl MenuProvider) -> u8 {
```
coco875: wouldn't be better to return an enum instead of a u8 ?

steel-core/src/entity/ai/brain/behavior/set_entity_look_target.rs
```diff
@@ -0,0 +1,74 @@
...
+        Self {
+            state: BehaviorState::new(ENTRY_CONDITION),
+            can_look_at: Box::new(can_look_at),
+            max_dist_sqr: f64::from(max_dist) * f64::from(max_dist),
+        }
+    }
+
+    fn find_closest(&self, entities: &[SharedEntity], mob_position: DVec3) -> Option<SharedEntity> {
+        let mut best: Option<(&SharedEntity, f64)> = None;
+        for entity in entities {
```
coco875: hmm maybe a bit inefficient ?

coco875 (commented):
some comment

steel-core/src/entity/ai/brain/behavior/set_entity_look_target.rs
teknostom: Idk if there is any better way to do it.

steel-core/src/player/player_inventory.rs
teknostom: Idk, makes sense to me that the open menu returns the opened menu id, because you already have all the other stuff. But maybe a rework of the inv system would be to make a type of the container id, so that it's typed as so.

steel-core/src/world/mod.rs
teknostom: Done

steel-core/src/entity/ai/brain/behavior/set_entity_look_target.rs
coco875: hmm maybe search in colision entity if we didn't do a structure to see who are the closest already

teknostom: Done

steel-core/src/player/player_inventory.rs
teknostom: Done

steel-core/src/behavior/blocks/fluid/bubble_column_block.rs
```diff
@@ -245,7 +246,7 @@ mod tests {
         fn new() -> Self {
             Self {
                 base: EntityBase::new(
-                    1,
+                    EntityId::new(1),
```
CuzImClicks, line 249: next_entity_id

CuzImClicks (changes requested):

JunkyDeveloper:
Will close this PR, because of the new ones

PR #250, opened 2026-07-12
Refactor the command parser, executor and implement permissions. 

4lve:
More info can be found by reading https://github.com/Steel-Foundation/SteelDocs/pull/32


package-content/config.toml
```diff
@@ -15,7 +15,9 @@ simulation_distance = 10
 online_mode = true
 # Optional authentication endpoint for online mode. Omit to use Mojang's session server.
 # auth_server = "https://sessionserver.mojang.com/session/minecraft/hasJoined"
-# Whether to enable encryption for client-server communication
+# Optional name-to-profile lookup endpoint used by commands such as /op.
+# profile_server = "https://api.minecraftservices.com/minecraft/profile/lookup/name"
```
coco875, line 20: I don't see this new option in the pr to doc (so keep update doc about config.toml)

steel-core/src/command/brigadier/LICENSE-MIT
coco875, line 1: why ?

steel-core/src/command/builtins/clear.rs
coco875, line 1: why all of this file haven't been move to keep the git historic

steel-core/src/behavior/items/default.rs
```diff
@@ -29,7 +29,7 @@ impl ItemBehavior for DefaultItemBehavior {
         match result {
             EquipmentSwapResult::Success(overflow) => {
                 if !overflow.is_empty() {
-                    context.player.drop_item(overflow, false, false);
+                    let _ = context.player.drop_item(overflow, false, false);
```
coco875, line 32: I don't get why there is this change

steel-core/src/block_entity/entities/end_gateway.rs
coco875, line 1: I also don't get why you add test in block_entities don't seems related to command outside of nbt but nbt should already have test.

steel-core/src/chunk/chunk_map.rs
```diff
@@ -1087,6 +1087,7 @@ impl ChunkMap {
                             player.connection.send_encoded(encoded.clone());
                         }
                     }
+                    world.broadcast_block_entity_if_needed(block_pos);
```
coco875, line 1090: seems to be fix, but related to what command ? (outside of command I guess it's for redstone etc)

steel-core/src/chunk/level_chunk.rs
```diff
@@ -797,16 +797,16 @@ impl LevelChunk {
         let y = pos.0.y;
         let section_index = self.get_section_index(y);
 
-        // Bounds check - return air if out of range
+        // `LevelChunk` returns air outside its section array; `World` handles void air.
         if section_index >= self.sections.sections.len() {
-            return REGISTRY.blocks.get_base_state_id(&vanilla_blocks::VOID_AIR);
+            return REGISTRY.blocks.get_base_state_id(&vanilla_blocks::AIR);
```
coco875, line 802: it's a rename or not ?

steel-core/src/command/brigadier/builder.rs
```diff
@@ -0,0 +1,408 @@
...
+            executor: self.executor.as_ref().map(Arc::clone),
+            requirement: self.requirement.clone(),
+            execution_requirement: self.execution_requirement.clone(),
+            redirect: self.redirect.clone(),
+        }
+    }
+}
+
+/// Creates a literal using the standard synchronous Brigadier runtime.
+pub(crate) fn literal<S>(name: impl Into<Box<str>>) -> CommandNodeBuilder<S> {
```
coco875, line 86: why Box<str> and not just String ?

steel-core/src/command/brigadier/context.rs
```diff
@@ -0,0 +1,592 @@
...
+
+#[derive(Clone, Debug, PartialEq)]
+struct ParsedArgument<V> {
+    range: StringRange,
+    value: V,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+struct ParsedArguments<V> {
+    values: Vec<(Box<str>, ParsedArgument<V>)>,
```
coco875, line 18: why not an hashmap ? (and also use String instead of Box<str>

package-content/groups.toml
```diff
@@ -0,0 +1,31 @@
...
+# allow = [
+#     "minecraft.command.gamemode{domain=lobby}",
+#     "plugin.region.build{world=lobby:spawn,plugin:region=spawn}",
+# ]
+# Namespaced typed values consumed by Steel subsystems and plugins.
+# More-specific contexts win; equally specific entries use group priority.
+# metadata = [
+#     { key = "plugin:prefix", value = "[Builder]" },
+#     { key = "plugin:max_homes{domain=survival}", value = 5 },
+# ]
```
JunkyDeveloper: would it make not more sense to link to the docs instead of this?

Joniii11: i mean i think that this can stay there just maybe also add ontop of it the link to the docs.

steel-core/src/behavior/blocks/portal/fire.rs
```diff
@@ -118,8 +118,11 @@ impl FireBlock {
         effect_collector.run_after(
             InsideBlockEffectType::FireIgnite,
             Box::new(move |entity| {
-                if !entity.fire_immune() {
+                if !entity.fire_immune()
+                    && let Some(world) = entity.level()
```
JunkyDeveloper: at campfire and magma block you called it entity_world and I like consistent things :D
and if I read world I associate it directly with the block worlds (would be also easier to understand it quickly

steel-core/src/behavior/blocks/vegetation/cactus_block.rs
```diff
@@ -204,7 +204,11 @@ impl BlockBehavior for CactusBlock {
         _effect_collector: &mut InsideBlockEffectCollector,
         _is_precise: bool,
     ) {
+        let Some(world) = entity.level() else {
```
JunkyDeveloper: same here

steel-core/src/command/commands/steel.rs
JunkyDeveloper, line 1: How is the current command for that?

4lve: You can use /domain and for same domain world travel you can use /execute in

JunkyDeveloper: never really used execute, but didn't found the execute.rs file maybe I'm blind :D But I trust you that we have that command :)

steel-utils/src/nbt/snbt.rs
```diff
@@ -0,0 +1,974 @@
...
+use uuid::Uuid;
+
+use crate::{UuidExt, java};
+
+/// Error returned when parsing SNBT text.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SnbtError {
+    cursor: usize,
+    message: String,
+}
```
Laptop59, line 21: I think `SnbtError` should store some 'error type' enum or translation key instead of a `String` for the error `message`, as Minecraft displays SNBT errors via translatable text components.

Looks like suggestions are not implemented as well, and maybe for tab completion in the server terminal or console, it could be good to have suggestions for some parser errors.

(I'm not sure if the project is looking to have 100% command error parity)

steel-utils/src/nbt/snbt.rs
```diff
@@ -0,0 +1,974 @@
...
+                    .map(|value| NbtTag::Byte(i8::from(value)))
+                    .ok_or_else(|| Self::error_at(name_cursor, "bool expects a numeric tag"))
+            }
+            "uuid" => {
+                let value = self.parse_tag()?;
+                self.expect_char(')')?;
+                let NbtTag::String(uuid) = value else {
+                    return Err(Self::error_at(name_cursor, "uuid expects a string tag"));
+                };
+                let uuid = Uuid::parse_str(uuid.as_str().to_str().as_ref())
```
Laptop59: Something to be a bit careful about as Java's `UUID` class does not parse UUIDs the same way as the `uuid` crate in Cargo.

steel-core/src/command/builtins/perms/config.rs
```diff
@@ -0,0 +1,338 @@
...
+    Ok(true)
+}
+
+pub(super) fn delete_group(
+    config: &mut PermissionGroupsConfig,
+    group: &str,
+) -> Result<bool, PermissionGroupEditError> {
+    if group == OP_GROUP {
+        return Err(PermissionGroupEditError::Required(group.to_owned()));
+    }
```
coco875, line 70: why hard code it ? you can't just add a permission in op_group to force this behavior ?

steel-core/src/command/builtins/perms/config.rs
```diff
@@ -0,0 +1,338 @@
...
+    if let Some((child, _)) = config
+        .groups
+        .iter()
+        .find(|(_, child)| child.inherits.iter().any(|parent| parent == group))
+    {
+        return Err(PermissionGroupEditError::InheritedBy {
+            group: group.to_owned(),
+            child: child.clone(),
+        });
+    }
```
coco875, line 86: maybe allow a delete recursif, so to allow delete a whole branch easily (like you have a group for tree able to remove also rolve who plant tree, who cut tree, etc) instead of deleting each role manually

steel-core/src/command/builtins/perms/config.rs
```diff
@@ -0,0 +1,338 @@
...
+            group_config.allow.push(value);
+            Ok(allow_count != 1 || deny_count != 0)
+        }
+        Some(PermissionState::Deny) => {
+            group_config.deny.push(value);
+            Ok(deny_count != 1 || allow_count != 0)
+        }
+        None => Ok(allow_count != 0 || deny_count != 0),
+    }
+}
```
coco875, line 151: a more random question does we can emit error if there is permission that conflict (like have one make unable to place block and another who allow but only dirt so there is a contradiction)

steel-core/src/command/builtins/perms/config.rs
```diff
@@ -0,0 +1,338 @@
...
+                    "permission group '{group}' cannot inherit itself"
+                )
+            }
+        }
+    }
+}
+
+impl Error for PermissionGroupEditError {}
+
+pub(super) fn create_group(
```
coco875, line 51: why not making it part of an impl PermissionGroupEditError (I think the same for most function below)

steel-core/src/command/builtins/clear.rs
```diff
@@ -0,0 +1,222 @@
...
+fn command() -> CommandNodeBuilder<CommandSource, SteelCommandRuntime> {
+    literal("clear").executes(clear_self).then(
+        argument("targets", SteelArgumentType::players())
+            .executes(clear_targets)
+            .then(
+                argument("item", SteelArgumentType::item_predicate())
+                    .executes(clear_matching)
+                    .then(
+                        argument("maxCount", ArgumentType::integer(0, i32::MAX))
+                            .executes(clear_matching_with_limit),
```
coco875, line 32: can we limit the  number of tabulation for command ? because I found them a bit to big

steel-core/src/command/builtins/clear.rs
```diff
@@ -0,0 +1,222 @@
...
+                    ),
+            ),
+    )
+}
+
+fn clear_self(context: &SteelCommandContext<CommandSource>) -> Result<i32, CommandSyntaxError> {
+    let Some(player) = context.source().player() else {
+        return Err(CommandSyntaxError::dynamic(TextComponent::from(
+            &translations::PERMISSIONS_REQUIRES_PLAYER,
```
coco875, line 41: hmm a bit a weird name for the error, I think

steel-core/src/command/builtins/clear.rs
coco875, line 1: a random question but permission have no way to scope ?
so like for example allow a player to clear him self but not other, or allow an admin to clear user but not operator/owner, etc ?

steel-core/src/command/builtins/difficulty.rs
```diff
@@ -0,0 +1,126 @@
...
+        ));
+    }
+
+    for world in worlds {
+        let mut level_data = world.level_data.write();
+        level_data.data_mut().difficulty = difficulty;
+        let locked = level_data.data().difficulty_locked;
+        drop(level_data);
+        world.broadcast_to_all(CChangeDifficulty { difficulty, locked });
+    }
```
coco875, line 80: maybe wrap it in a function in world so to have world.get_difficulty and world.set_difficulty that do all the drop and broadcast, to simplify

steel-core/src/command/builtins/gamemode.rs
```diff
@@ -0,0 +1,356 @@
...
+use crate::player::Player;
+use crate::server::Server;
+use crate::world::World;
+
+const GAME_MODES: [GameType; 4] = [
+    GameType::Survival,
+    GameType::Creative,
+    GameType::Adventure,
+    GameType::Spectator,
+];
```
coco875, line 32: hmm was thinking of maybe also make GameType as their own set of permissions like adventure who are survival but you can't break block, etc. And also being able to extend it but maybe a bit outside of the pr

steel-core/src/command/builtins/gamerule.rs
```diff
@@ -0,0 +1,200 @@
...
+    registration::CommandRegistration,
+};
+
+pub(super) fn registration() -> CommandRegistration<CommandSource> {
+    CommandRegistration::new(Identifier::vanilla_static("gamerule"), |_| command())
+}
+
+fn command() -> CommandNodeBuilder<CommandSource, SteelCommandRuntime> {
+    let mut command = literal("gamerule");
+    for (_, rule) in REGISTRY.game_rules.iter() {
```
coco875, line 20: not the correct file but was thinking of putting some gamerule as permission like pvp (some other like entity cramming don't work) but the point are define gamerule as the default permission for player but a group can have another permission that take priority

steel-core/src/command/builtins/give.rs
coco875, line 1: similar to clear allow more finetune permission to allow for example to block the give of op block (to avoid have a troll for example that use command block to kill everyone at each tick)

steel-core/src/command/builtins/give.rs
```diff
@@ -0,0 +1,215 @@
...
+                item.make_fake_item();
+            }
+            play_pickup_sound(player);
+            player.broadcast_inventory_changes();
+        } else if let Some(item) = player.drop_item(stack, false, false) {
+            item.set_no_pickup_delay();
+            item.set_owner(Some(player.gameprofile.id));
+        }
+    }
+}
```
coco875, line 123: also think that it would be better to move in a method of player (to allow also be re-use)

steel-core/src/command/builtins/perms.rs
```diff
@@ -0,0 +1,1508 @@
...
+};
+use crate::permission::{
+    PermissionContext, PermissionEntry, PermissionExpr, PermissionKey, PermissionMetadataEntry,
+    PermissionMetadataExpression, PermissionMetadataValue, PermissionResolutionSource,
+    PermissionRuleExpression, PermissionState, PermissionSubjectState,
+};
+
+pub(super) const MANAGE_ALL_PERMISSION: &str = "steel.permission.manage.*";
+pub(super) const GROUP_ALL_PERMISSION: &str = "steel.permission.group.*";
+pub(super) const METADATA_PERMISSION: &str = "steel.permission.metadata";
```
coco875, line 28: it's used only for test ?

steel-utils/src/nbt/snbt.rs
4lve: We don't require exact parity on edge cases like these, like the rust f64 formatter differs a bit from java as well which is already documented. Good catch though will document it with a comment.

package-content/config.toml
```diff
@@ -1,4 +1,4 @@
-#:schema https://raw.githubusercontent.com/4lve/SteelMC/refs/heads/master/package-content/config.schema.json
+#:schema https://raw.githubusercontent.com/Steel-Foundation/SteelMC/refs/heads/master/package-content/config.schema.json
```
coco875, line 1: maybe also add a link to doc like groups.toml

coco875 (changes requested):
part 1

steel-core/src/command/builtins/perms.rs
```diff
@@ -0,0 +1,1508 @@
...
+}
+
+fn user_allow(
+    context: &SteelCommandContext<CommandSource>,
+) -> Result<PermsCommandSuspension, CommandSyntaxError> {
+    user_permission(context, Some(PermissionState::Allow))
+}
+
+fn user_deny(
+    context: &SteelCommandContext<CommandSource>,
```
coco875, line 338: why doing this split and regroup ? in the comment you split per operatation and then in function you regroup again with enum why not set the enum before to reduce the number of function or manually inline the function ?

steel-core/src/command/builtins/perms.rs
```diff
@@ -0,0 +1,1508 @@
...
+    Ok(PermsCommandSuspension {
+        source,
+        order,
+        broadcast_to_admins: order == CommandSuspensionOrder::Global,
+        receiver,
+        task: Some(task),
+    })
+}
+
+async fn run_operation(
```
coco875, line 720: here it's where it regroup again while most of the time operation var are static.

steel-core/src/command/builtins/perms.rs
```diff
@@ -0,0 +1,1508 @@
...
+    PermissionMetadataExpression, PermissionMetadataValue, PermissionResolutionSource,
+    PermissionRuleExpression, PermissionState, PermissionSubjectState,
+};
+
+pub(super) const MANAGE_ALL_PERMISSION: &str = "steel.permission.manage.*";
+pub(super) const GROUP_ALL_PERMISSION: &str = "steel.permission.group.*";
+pub(super) const METADATA_PERMISSION: &str = "steel.permission.metadata";
+
+pub(super) fn registration() -> CommandRegistration<CommandSource> {
+    CommandRegistration::new(Identifier::new_static("steel", "perms"), |_| command())
```
coco875: maybe silmilar to vanilla also have a Identifier::from_steel

steel-utils/src/nbt/snbt.rs
4lve: Thanks, fixed

steel-core/src/command/brigadier/LICENSE-MIT
4lve: Moved it into mod.rs

steel-core/src/command/builtins/clear.rs
4lve: Commands were rewritten from the ground up with the new executor so they weren't really moved. Not sure if you can move it afterwards though

steel-core/src/behavior/items/default.rs
4lve: Vanilla parity. It returns the item entity so you can create a fake item drop used in the /give command

steel-core/src/block_entity/entities/end_gateway.rs
4lve: This is for the /execute command which needs access to block entity data in some cases

steel-core/src/command/brigadier/LICENSE-MIT
coco875: hmm don't match with convertation that we have on discord
> Will prob remove it since the code is not even similar to their code anymore

so can you elaborate on why we have code under mit and not agpl like the rest of our code ?

steel-core/src/command/builtins/gamerule.rs
4lve: Could be done in some cases. I think something like minecraft.bypass.movement might be a nice addition

steel-core/src/behavior/items/default.rs
coco875: (personally I think move it in a pr to fix all vanilla parity signature would be better because don't look like to be required by permission, etc)

steel-core/src/command/builtins/give.rs
4lve: I like the idea but I think it might be to specialized to be in core. Like you can give yourself a command block just by being in creative mode

steel-core/src/command/builtins/perms.rs
4lve: Also used with builder.declare_permission to add it as an autocomplete suggestion

package-content/config.toml
4lve: Yea, good idea

package-content/config.toml
4lve: Will fix thanks

steel-core/src/command/builtins/gamerule.rs
coco875: I thin that:
keepInventory
allow_entering_nether_using_portals
ender_pearls_vanish_on_death
fall_damage
immediate_respawn
limited_crafting
player_movement_check
players_nether_portal_default_delay
pvp
recipesunlock

are gamerule that can be permissions (link: https://minecraft.wiki/w/Game_rule)
so like that we can controle different group of player (like if we want to manage different kind of player in one world)

steel-core/src/command/builtins/give.rs
4lve: Not sure about that. This is /give specific behavior which plays a sound and adds a fake item entity. If so it would be worth looking into it once we are working on the plugin api

steel-core/src/command/builtins/teleport.rs
coco875, line 1: hmm I think overhaul this whole file look more complicated then what it does but I don't really have suggestion

steel-core/src/command/builtins/tellraw.rs
```diff
@@ -0,0 +1,108 @@
...
+
+fn send_message(context: &SteelCommandContext<CommandSource>) -> Result<i32, CommandSyntaxError> {
+    let targets = context.players("targets")?;
+    let Some(message) = context.text_component("message") else {
+        return Err(CommandSyntaxError::dynamic(
+            "Parsed text component is missing from the command context",
+        ));
+    };
+    let result = i32::try_from(targets.len()).map_err(|_| {
+        CommandSyntaxError::dynamic("Target player count exceeds the command result range")
```
coco875, line 34: hmm the error message are a bit weird, I think that "The message whould be less then 2**31 (the real number of course)" would be better

steel-utils/src/nbt.rs
coco875, line 1: we already don't have something like that in TextComponent (or pr before related to it) ?

steel-core/src/entity/manager.rs
```diff
@@ -318,13 +318,45 @@ struct ManagerState {
     chunk_visibility: FxHashMap<ChunkPos, EntityVisibility>,
     live_by_id: FxHashMap<i32, EntityEntry>,
     live_by_uuid: FxHashMap<Uuid, i32>,
-    by_section: FxHashMap<SectionPos, FxHashSet<i32>>,
+    accessible_order: OrderedEntityIds,
+    by_section: BTreeMap<PackedSectionPos, OrderedEntityIds>,
```
coco875, line 322: just to report the conversation about that on discord here, it's to make order of selector match vanilla (like `@e`)

steel-core/src/entity/mob.rs
```diff
@@ -1323,7 +1323,7 @@ pub trait Mob: LivingEntity {
 
     /// Handles vanilla `Mob.doHurtTarget`.
     #[must_use]
-    fn do_hurt_target(&self, target: &SharedEntity) -> bool {
+    fn do_hurt_target(&self, world: &World, target: &SharedEntity) -> bool {
```
coco875, line 1326: I know it's to match vanilla signature but does it need to be in the same pr as command ? (does there is command or permission that need it ?)

steel-core/src/entity/mod.rs
coco875, line 1: by memory I also made a comment like that in another pr but this file are too big (to avoid forget it to refactor it later).

steel-core/src/command/execution/argument.rs
```diff
@@ -0,0 +1,1492 @@
...
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct TimeParser {
+    minimum: i32,
+}
+
+impl_downcast_type!(TimeParser, "steel:command/parser/time");
+
+impl SteelArgumentParser for TimeParser {
```
coco875, line 595: hmm now that I think of it why not just call it ArgumentParser ? (the steel Prefix give nothing)

coco875 (changes requested):
part 2

steel-core/src/chunk/level_chunk.rs
4lve: Our old behavior was wrong

steel-core/src/command/brigadier/builder.rs
4lve: Box doesn't need to store capacity

coco875: hmm forget that String as a Vec so also have a capacity so yeah also a bit less space on the heap

steel-core/src/command/brigadier/context.rs
4lve: Not enough arguments to justify a HashMap

steel-core/src/command/builtins/perms/config.rs
4lve: If you could delete the op group the /op command would stop working

steel-core/src/command/brigadier/context.rs
coco875: hmm we don't have a lib call SmallVec also kinda make for this kind of situation ?

4lve: I think the types are too complex to justify usage of smallvec here

steel-core/src/command/builtins/perms/config.rs
coco875: I know but we can't just add a permission like can.delete.group and you can delete group at the same level that you but only lower ? and also does it mean that the server will emit an error if you delete op in groups.toml ?

4lve: Yea, the op group is required

steel-core/src/command/builtins/perms/config.rs
4lve: We already have conflict resolution in docs, I think it's intended behavior

steel-core/src/command/builtins/perms/config.rs
coco875: and for the first question ? to avoid delete op group but also avoid that an admin to delete his group, etc.

steel-core/src/command/builtins/clear.rs
4lve: You can split them up like we do with /perms and /execute but sometimes like in this case the indentation makes it easier to follow

steel-core/src/command/builtins/clear.rs
4lve: Blame vanilla xD

steel-core/src/command/builtins/clear.rs
4lve: Yea, such a permission check could easily be added. Out of scope for this PR though

steel-core/src/command/builtins/gamemode.rs
4lve: Yea, cool idea but out of scope

steel-core/src/command/builtins/perms.rs
4lve: Added

steel-core/src/command/builtins/clear.rs
coco875: change an error message would be more qol then break vanilla x)

steel-core/src/command/builtins/teleport.rs
4lve: Yea, worth looking into in the future

steel-core/src/command/builtins/clear.rs
coco875: ok, if you are fine with that, personally perfer have just one line = one addition of argument instead of having a shift to the right

steel-core/src/command/builtins/tellraw.rs
4lve: Yea, this error would realistically never happen

steel-utils/src/nbt.rs
4lve: Don't think so

steel-core/src/command/builtins/tellraw.rs
coco875: with command block like we can copy paste big blob of text I think it can happen

steel-core/src/entity/mob.rs
4lve: `/execute in overworld` for example

steel-core/src/entity/mod.rs
4lve: Yea, def needs a refactor/split. Out of scope here though

steel-core/src/command/execution/argument.rs
4lve: Our Brigadier implementation is designed to be generic so the SteelArgumentParser is the concrete type

steel-core/src/command/builtins/tellraw.rs
4lve: You would need 2^32 logged in players so I don't think that is even remotley possible

coco875: hmm never say never x) so just in case it happen have a clear fallback

steel-core/src/command/builtins/clear.rs
4lve: Would make it hard to know if you are implementing
`/test 1`

A following argument `/test 1 2`
An alternative argument `/test 2`

steel-core/src/chunk/chunk_map.rs
4lve: This was introduced for /execute store result|success block. That command changes block-entity NBT without changing the block state

steel-core/src/command/builtins/clear.rs
coco875: then when there is an or have a tabulation to make clear but we have more often chain of argument then alternative to argument so for me it make sense

steel-core/src/command/builtins/perms/config.rs
4lve: inheritance is a graph not a tree so in this case it wouldn't make sense. A child can inherit multiple parents

steel-core/src/command/builtins/difficulty.rs
4lve: Thanks, fixed

steel-core/src/command/builtins/perms/config.rs
coco875: hmm was more thinking of a case when there is one parent or when it's use like a kind of subfolder for group

steel-core/src/command/builtins/perms/config.rs
4lve: That would expose /perms specific behavior on the whole struct

steel-core/src/command/builtins/perms.rs
4lve: Brigadier provides a borrowed context while profile lookup and persistence run in a spawned async task. It stores the parsed runtime arguments that must outlive the callback

steel-core/src/command/builtins/perms/config.rs
4lve: Could look into that in the future in that case

steel-core/src/command/brigadier/argument_tests.rs
coco875, line 1: hmm something that I didn't view when doing the review is why now you split implementation and test ?

like
steel-core/src/command/brigadier/argument.rs
steel-core/src/command/brigadier/argument_tests.rs

steel-core/src/command/brigadier/context_chain.rs
steel-core/src/command/brigadier/context_chain_tests.rs

4lve: I find it cleaner if tests grow too large, then you can have only logic in one file and only invariant tests in another

coco875: then there is other pr that need to be open for also large file with large test

PR #251, opened 2026-07-13
feat: implement cobwebs (web block)

SomeYellowGuy:
Implements `WebBlock`, the block responsible for the behavior of cobwebs.


coco875 (approved):
Look good (I didn't compare to vanilla personally to double check)

PR #252, opened 2026-07-13
feat: implement ladders

Laptop59:
This PR implements the behavior for ladder blocks. Any feedback is greatly appreciated!


coco875 (approved):
Look good (I didn't compare to vanilla personally to double check)

PR #253, opened 2026-07-13
Startup optimizations

DarkMrMelther:
- Parallelized stronghold ring positions
- Now stronghold rings are saved into [world_folder]/data/stronghold_rings.toml
- Parallelized tree_build for possible_biomes_refs

Before:
- Debug: ~7s
- Release: ~1s

After (First start):
- Debug: ~1.2s
- Release: ~0.16s

After (Hot start):
- Debug: ~0.12s
- Release: ~0.02s


steel-utils/src/saved_data.rs
```diff
@@ -18,6 +18,8 @@ pub mod names {
 
     /// Vanilla `TicketStorage.TYPE`, persisted as `data/chunk_tickets.toml`.
     pub const CHUNK_TICKETS: SavedDataName = SavedDataName::trusted("chunk_tickets");
+    /// Stronhold generation data, persisted as `data/stronghold_rings.toml`
+    pub const STRONGHOLD_RINGS: SavedDataName = SavedDataName::trusted("stronghold_rings");
```
coco875: a bit out of scope of this pr but why we store as toml and not binary ? to have faster read and write and we don't need that an admin or user to modify it

steel-worldgen/src/structure/placement.rs
```diff
@@ -164,7 +168,51 @@ where
...
+
+                        // sectionToBlockCoord(x, 8) = x * 16 + 8; snap result is blocks → >> 4.
+                        if let Some((sx, sz)) =
+                            snap(initial_x * 16 + 8, initial_z * 16 + 8, &mut forked)
+                        {
+                            ChunkPos::new(sx >> 4, sz >> 4)
+                        } else {
+                            ChunkPos::new(initial_x, initial_z)
+                        }
+                    })
```
coco875: the lambda function are the same as below, so we can factorize it.

coco875 (changes requested):
look good else just one comment on a lambda function (and another off-topic to the pr so you can ignore it)

steel-utils/src/saved_data.rs
DarkMrMelther: This file was only moved, and added sync functions, the toml decision was made by others (4lve I guess)

steel-worldgen/src/structure/placement.rs
DarkMrMelther: I'll take a look

steel-utils/src/saved_data.rs
coco875: I know it's have been move and I think that in steel-utils it's better, didn't make a comment on that. But yeah this question are for 4lve then for you or in link with the pr

4lve: Depends on how large the file contents are if it deserves a custom wincode binary format

DarkMrMelther: If I remember correctly, the file is an array of 128 vec2.

coco875: hmm I am in favor to micro-optimize that because it's startup time, so if add 1-2-50 files that are need to be toml read and parse it will take a big chunk at the end

steel-worldgen/src/structure/placement.rs
DarkMrMelther: Done, take a look

steel-worldgen/src/structure/placement.rs
```diff
@@ -130,6 +130,24 @@ where
         forked: LegacyRandom,
     }
 
+    fn map_candidate<F>(candidate: RingCandidate, snap: &F) -> ChunkPos
+    where
+        F: Fn(i32, i32, &mut LegacyRandom) -> Option<(i32, i32)> + Sync,
+    {
```
coco875, line 136: why F is generic ? because snap function are hte same for both no ?

coco875 (changes requested):

steel-worldgen/src/structure/placement.rs
DarkMrMelther: I need to do the generic to match the parent function signature, I've tried using the already made generic but it seems like even if is inside the other function they doesn't share the generics

coco875: oh didn't see that snap_biome are also a generic

coco875: then perfect

coco875 (approved):

steel-utils/src/saved_data.rs
4lve: Okay, changed it to a bin format now. toml is mostly chosen for files where it's intended that you might edit them

PR #255, opened 2026-07-13
feat(item, entity): fishing rod + fishing hook entity

tn-lorenz:
# Type of change
- [X] Item implementation
- [X] Entity implementation
- [X] New feature

# Description
This PR aims to implement the fishing rod item and the fishing hook entity. As we currently lack a few core components neccessary to implement the full fishing hook logic, I left TO-DO's for advancements and vibration.

# Checklist
- [x] Self-reviewed the diff
- [x] No leftover debug code / comments (I have left some comments that are not debug comments because I thought they might be helpful)

# Classes modified:
- FishingRodItem
- FishingHook

# How this was tested
- it compile
- in-game testing

## WORKING:

- [X] hooking and reeling in entitites
- [X] casting and reeling in rod
- [X] fish biting
- [X] sounds
- [X] drop loot
- [X] spawn `ExperienceOrb`
- [X] award stats
- [X] seems to be thread-safe now

## TODO:
- [X] address unused variables/fns
- [X] add missing docs
- [x] understand loot tables so a fish gets dropped
- [x] check the src to know where I need to change `f64` to `f32` (because I accidentally confused the meanings of `float` and `double` during the initial implementation process)
- [x] check if the meaning of the `const`s I added for seconds are sound everywhere (idk if the "magic" numbers outside of the `tick` fn represent ticks as well -- for now I just assumed they would)
- [x] don't forget to check if I re-added the one else-branch that might've gone missing during refactoring for the 1st and 2nd time
- [x] check if omitted fn's such as `can_hit_entity` are correctly represented by the default `Projectile` trait, or if I have to override
- [x] slime Clippy out


coco875:
fix lint and test x)

steel-core/src/behavior/items/fishing_rod.rs
```diff
@@ -0,0 +1,135 @@
...
+            ));
+
+            if let Some(owner) = world.players.get_by_uuid(&player.gameprofile.id) {
+                let owner: SharedEntity = owner;
+                hook.set_owner(&owner);
+            } else {
+                hook.set_owner_uuid(Some(player.gameprofile.id));
+                player.set_fishing_hook(&hook);
+            }
+            //hook.set_item_clamped(thrown_item);
```
coco875, line 70: ?

steel-core/src/entity/projectile/fishing_hook.rs
```diff
@@ -0,0 +1,756 @@
...
+
+        let inventory = owner.inventory.lock();
+
+        let mainhand_item = inventory.get_item_in_hand(InteractionHand::MainHand);
+        let offhand_item = inventory.get_offhand_item();
+
+        let mainhand_fishing = mainhand_item.is(&vanilla_items::FISHING_ROD);
+        let offhand_fishing = offhand_item.is(&vanilla_items::FISHING_ROD);
+
+        if (mainhand_fishing || offhand_fishing) && self.distance_to_sqr(owner.position()) <= 1024.0
```
coco875, line 114: magic number, like it's distance_to_sqr then 1024 = 32*32 so a const for 32

steel-core/src/entity/projectile/fishing_hook.rs
```diff
@@ -0,0 +1,756 @@
...
+                world.send_particles(
+                    ParticleData::simple(&FISHING),
+                    particle_pos,
+                    particle_count,
+                    particle_spread,
+                    0.2,
+                );
+
+                state.nibble = rng().random_range(20..=40);
+                // TODO: `fishing_hook_mut()` or `fishing_hook` ?
```
coco875, line 260: ?

steel-core/src/entity/projectile/fishing_hook.rs
```diff
@@ -0,0 +1,756 @@
...
+            match state.time_until_lured {
+                0..20 => {
+                    tease_chance += f64::from(20 - state.time_until_lured) * 0.05;
+                }
+                20..40 => {
+                    tease_chance += f64::from(40 - state.time_until_lured) * 0.02;
+                }
+                40..60 => {
+                    tease_chance += f64::from(60 - state.time_until_lured) * 0.01;
+                }
```
coco875, line 276: magic number

steel-core/src/entity/projectile/fishing_hook.rs
```diff
@@ -0,0 +1,756 @@
...
+                _ => {}
+            }
+
+            if rng().random::<f64>() < tease_chance {
+                // same reason to call this early in here as well: no need to calculate the rest if there is no world to spawn the particle in.
+                let Some(world) = self.level() else {
+                    return;
+                };
+
+                let angle: f64 = rng().random_range(0.0..=360.0) * std::f64::consts::PI / 180.0;
```
coco875, line 286: maybe add a global const for 360° ? (same for 180)

steel-core/src/entity/projectile/fishing_hook.rs
```diff
@@ -0,0 +1,756 @@
...
+                    );
+                }
+            }
+
+            if state.time_until_lured <= 0 {
+                state.fish_angle = rng().random_range(0.0..=360.0);
+                state.time_until_hooked = rng().random_range(20..=80);
+            }
+        } else {
+            state.time_until_lured = rng().random_range(100..=600);
```
coco875, line 312: magic number

steel-core/src/entity/projectile/fishing_hook.rs
```diff
@@ -0,0 +1,756 @@
...
+            if fluid_state.is_water() && fluid_state.is_source() && fluid_state.is_empty() {
+                OpenWaterType::InsideWater
+            } else {
+                OpenWaterType::Invalid
+            }
+        } else {
+            OpenWaterType::AboveWater
+        }
+    }
+    // fn is_open_water_fishing(){}
```
coco875, line 382: ?

steel-core/src/entity/projectile/fishing_hook.rs
```diff
@@ -0,0 +1,756 @@
...
+                            state.open_water = state.open_water
+                                && state.out_of_water_time < 10
+                                && self.calculate_open_water(pos);
+                        }
+
+                        if is_in_water {
+                            state.out_of_water_time = (state.out_of_water_time - 1).max(0);
+                            // TODO: `fishing_hook_mut()` or `fishing_hook` ?
+                            if *self.entity_data.lock().fishing_hook_mut().biting.get() {
+                                // TODO: -0.1 * this.syncronizedRandom.nextFloat() * this.syncronizedRandom.nextFloat()
```
coco875, line 589: Image
x)

steel-core/src/entity/projectile/fishing_hook.rs
```diff
@@ -0,0 +1,756 @@
...
+}
+
+impl Projectile for FishingHook {
+    fn projectile_base(&self) -> &ProjectileBase {
+        &self.projectile_base
+    }
+
+    /*fn can_hit_entity(&self, _entity: &dyn Entity) -> bool {
+        todo!()
+    }*/
```
coco875, line 644: ?

coco875 (changes requested):
some comment about code to be more clean

steel-core/src/entity/projectile/fishing_hook.rs
tn-lorenz: Idk which of these two functions is the correct one: both work

coco875: by default choose the non mut one

steel-core/src/entity/projectile/fishing_hook.rs
tn-lorenz: I just copied the java code: this is a spelling mistake from the java src

coco875: you are allowed to fix typos from vanilla naming x)

steel-core/src/entity/projectile/fishing_hook.rs
tn-lorenz: I think that's bs 🤣

steel-core/src/entity/projectile/fishing_hook.rs
tn-lorenz: Oh yeah I forgot to delete this; it's a useless function, even in vanilla

steel-core/src/entity/projectile/fishing_hook.rs
tn-lorenz: As these are clearly ranges of numbers, what would you suggest that I do?

steel-core/src/entity/projectile/fishing_hook.rs
tn-lorenz: I don't see how that would be useful here... maybe `const full_cirle` and `const half_circle`? But this could be more confusing as in the context, speaking about a circle doesn't make immediate sense

steel-core/src/entity/projectile/fishing_hook.rs
coco875: have const first_stage = 20, const second_stage = 40 etc, depending of what the value mean

steel-core/src/entity/projectile/fishing_hook.rs
coco875: I was more thinking of const DEGREE_360 and DEGREE_180 to make it clear that it's degree number that get after tranformed to radians

steel-core/src/entity/projectile/fishing_hook.rs
tn-lorenz: I indeed had to use the mut version

coco875: then it solve the todo x)

tn-lorenz:
Omg...me misspelling "spelling" to make the lint crash out is so ironic fml

steel-core/src/entity/entity/mod.rs
```diff
@@ -3474,6 +3474,23 @@ pub trait Entity: EntityEventSource + ErasedType + Send + Sync + 'static {
         };
         living.hurt_server(world, source, amount)
     }
+
+    // This already exists for structures and AABB whatever that might be, but I also need it for entities, I hope this is the right spot to put it.
```
coco875, line 3592: why have this comment alone sperate from doc ? (two / are for comment and three / are for doc)

steel-core/src/entity/projectile/fishing_hook.rs
coco875: still think that 1024 alone mean nothing and that we can't know that it's the max_distance squared of fishing before it stop

steel-core/src/entity/projectile/fishing_hook.rs
coco875: it's still no delete ?

steel-core/src/entity/projectile/fishing_hook.rs
```diff
@@ -0,0 +1,773 @@
...
+            let hooked_in = {
+                let hook_state = self.hook_state.lock();
+                hook_state.hooked_in.clone()
+            };
+
+            if let Some(hooked_in) = hooked_in {
+                self.pull_entity(&hooked_in);
+                // TODO: criteria triggers (advancements)
+                damage = if hooked_in.as_ref().is::<ItemEntity>() {
+                    3
```
coco875: magic number

steel-core/src/entity/projectile/fishing_hook.rs
```diff
@@ -0,0 +1,773 @@
...
+                } else {
+                    5
```
coco875: magic number

steel-core/src/entity/projectile/fishing_hook.rs
```diff
@@ -0,0 +1,773 @@
...
+                };
+            } else if self.hook_state.lock().nibble > 0 {
+                // TODO: Looting
+                // TODO: criteria triggers (advancements)
+                // TODO: award stat when catching fish
+            }
+
+            if self.base.on_ground() {
+                damage = 2;
```
coco875: magic number

steel-core/src/entity/projectile/fishing_hook.rs
```diff
@@ -0,0 +1,773 @@
...
+    fn tick(&self) {
+        if let Some(owner) = self.get_owner()
+            && let Some(player) = owner.as_player()
+            && !self.should_stop_fishing(player)
+        {
+            if self.on_ground() {
+                let should_remove = {
+                    let mut state = self.hook_state.lock();
+                    state.life += 1;
+                    state.life >= 1200
```
coco875: magic number

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
coco875, line 1: move it to steel-core/src/entity/entities/objects/projectiles

steel-utils/src/geometry.rs
```diff
@@ -29,6 +29,7 @@ impl Space for BlockLocal {
     const ZERO_SPAN_IS_EMPTY: bool = true;
 }
 
+/// Marker type for world-space AABBs.
```
coco875: ?

coco875 (changes requested):

steel-core/src/entity/entity/mod.rs
tn-lorenz: ...because this is not a doc comment...this is a comment concerning whoever implements the missing TODO's??/Revievs???

steel-utils/src/geometry.rs
tn-lorenz: I did not add this, this extra line came from your AI pr

coco875: I have read my two commit and didn't touch this file so you talk about what ?

coco875: https://github.com/Steel-Foundation/SteelMC/pull/255/changes/96e40408954e929cdfe22c7e3c0fea73d42aa781
https://github.com/Steel-Foundation/SteelMC/pull/255/changes/c401e670af42fa56d9f9cc7cb94dc09fc4b1bdf8
if you want to check

steel-core/src/entity/entity/mod.rs
coco875: I don't get what you mean by bot a doc comment, I have only make two commit that didn't touch this file

tn-lorenz: Yeah I misspelled, I meant that this is just a message to help whoever is going to do the to-do's later, and I don't consider this documentation

steel-core/src/entity/projectile/fishing_hook.rs
tn-lorenz: But then I would make 1024 the const, not 32? `const MAX_DISTANCE_SQR: i32 = 1024`, for example? Or to make it more clear: `const MAX_DISTANCE_SQR: i32 = 32 * 32`

coco875: I think 32*32 would be better

steel-core/src/entity/projectile/fishing_hook.rs
tn-lorenz: weird, I already deleted this line, somehow it did not get pushed

steel-core/src/behavior/items/fishing_rod.rs
```diff
@@ -0,0 +1,131 @@
...
+#[item_behavior]
+pub struct FishingRodItem;
+
+impl ItemBehavior for FishingRodItem {
+    fn use_item(&self, context: &mut UseItemContext) -> InteractionResult {
+        let player = context.player;
+        let world = context.world;
+        let infinite_materials = context.player.has_infinite_materials();
+
+        let pitch = 0.4 / (rng().random::<f32>() * 0.4 + 0.8);
```
coco875, line 25: like 0.4 are redundant make it a const to say what it mean

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,797 @@
...
+    )]
+    fn catching_fish(&self, pos: BlockPos, state: &mut FishingHookState) {
+        let mut fishing_speed = 1;
+        let above = pos.above();
+
+        let Some(world) = self.level() else {
+            return;
+        };
+
+        if rng().random::<f64>() < 0.25 && world.is_raining_at(above) {
```
coco875: magic number

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,797 @@
...
+            fishing_speed += 1;
+        }
+
+        if rng().random::<f64>() < 0.5 && world.can_see_sky(above) {
```
coco875: magic number

coco875 (changes requested):

steel-core/src/behavior/items/fishing_rod.rs
```diff
@@ -0,0 +1,131 @@
...
+            let (yaw, player_pitch) = player.rotation();
+            hook.shoot_from_rotation(player, player_pitch, yaw, 0.0, SHOOT_POWER, 1.0);
+
+            let entity: SharedEntity = hook;
+            if let Err(error) = world.try_add_entity(Arc::clone(&entity)) {
+                entity.set_removed(RemovalReason::Discarded);
+                log::debug!("failed to spawn fishing hook: {error}");
+                return InteractionResult::Fail;
+            }
+            // TODO: award stat
```
Laptop59: Now that stats are implemented, you should be able to award the stat.

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,797 @@
...
+                // TODO: criteria triggers (advancements)
+                damage = if hooked_in.as_ref().is::<ItemEntity>() {
+                    DMG_ITEM_ENTITY
+                } else {
+                    DMG_DEFAULT
+                };
+            } else if self.hook_state.lock().nibble > 0 {
+                // TODO: Looting
+                // TODO: criteria triggers (advancements)
+                // TODO: award stat when catching fish
```
Laptop59: Award a stat here as well.

Laptop59 (changes requested):

steel-core/src/behavior/items/fishing_rod.rs
tn-lorenz: I don't think everything needs to be named...why do I have to name literally EVERY number, when Purdze for example, didn't either? Look at the exact equivalent to this: items\ender_pearl.rs:31

coco875: after discussion, it look like to be similar in other situation so stay like this, and I will make a pr to uniform this

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
tn-lorenz: This one isn't possible yet, we lack the corresponding stat implementation

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
tn-lorenz: I think everybody can infer that this means that the probability is supposed to be less than 25% ... making this a Const feels redundant

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
tn-lorenz: Same like above, everybody would understand that; I don't feel making this a const would be necessary

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
coco875: the problem are more what this const mean like if it's below 25% what does it do, does it's re-use somewhere else ? Can I re-use somewhere else ? etc

tn-lorenz: No it's not used somewhere else 🥀

coco875: have a const allow to have some hint of what the if does and what chaging this value mean

coco875 (dismissed):

coco875 (dismissed):

steel-core/src/behavior/items/fishing_rod.rs
```diff
@@ -0,0 +1,144 @@
...
+                inventory.get_item_in_hand(hand).clone()
+            };
+
+            if rod.is(&vanilla_items::FISHING_ROD) {
+                let damage = fishing.retrieve(&rod);
+
+                context.inv.with_item(|item| {
+                    item.hurt_and_break(damage, infinite_materials);
+                });
+            }
```
CuzImClicks: in this section you are locking the player inventory three separate times.

steel-core/src/behavior/items/fishing_rod.rs
```diff
@@ -0,0 +1,144 @@
...
+                player.set_fishing_hook(&hook);
+            }
+
+            let (yaw, player_pitch) = player.rotation();
+            hook.shoot_from_rotation(player, player_pitch, yaw, 0.0, SHOOT_POWER, 1.0);
+
+            let entity: SharedEntity = hook;
+            if let Err(error) = world.try_add_entity(Arc::clone(&entity)) {
+                entity.set_removed(RemovalReason::Discarded);
+                log::debug!("failed to spawn fishing hook: {error}");
```
CuzImClicks: should this be elevated to a warn or error?

steel-core/src/behavior/items/fishing_rod.rs
```diff
@@ -0,0 +1,144 @@
...
+                return InteractionResult::Fail;
+            }
+
+            let inventory = player.inventory.lock();
+            let rod = inventory.get_item_in_hand(hand);
```
CuzImClicks: locking inventory again -> read my comment on should_stop_fishing

steel-core/src/behavior/items/fishing_rod.rs
```diff
@@ -0,0 +1,144 @@
...
+use std::sync::Arc;
+use steel_macros::item_behavior;
+use steel_protocol::packets::game::SoundSource;
+use steel_registry::sound_events::{ENTITY_FISHING_BOBBER_RETRIEVE, ENTITY_FISHING_BOBBER_THROW};
+use steel_registry::stat::vanilla_stat_types;
+use steel_registry::{vanilla_entities, vanilla_items};
+
+const SHOOT_POWER: f32 = 1.5;
+
+/// literally self-explanatory
```
CuzImClicks: just correct this xD

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,893 @@
...
+/// This struct holds entity specific state information per fishing hook entity.
+pub struct FishingHookState {
+    out_of_water_time: i32,
+    life: i32,
+    nibble: i32,
+    time_until_lured: i32,
+    time_until_hooked: i32,
+    fish_angle: f32,
+    open_water: bool,
+    current_state: FishHookState,
```
CuzImClicks: FishingHookState contains a FishHookState? try and find a better name for the current state

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,893 @@
...
+    hooked_in: Option<SharedEntity>,
```
CuzImClicks: maybe hooked_entity would be clearer

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,893 @@
...
+    luck: i32,
+    lure_speed: i32,
+}
+
+impl FishingHookState {
+    #[must_use]
+    /// returns a new `FishingHookState` with the given lure speed and luck values.
```
CuzImClicks:
```suggestion
    /// Returns a new `FishingHookState` with the given lure speed and luck values.
```

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,893 @@
...
+const THREE_SECONDS: i32 = 60;
+const FOUR_SECONDS: i32 = 80;
+const FIVE_SECONDS: i32 = 100;
+const THIRTY_SECONDS: i32 = 600;
+const ONE_MINUTE: i32 = 1200;
+
+impl FishingHookEntity {
+    /// Creates a fishing hook entity.
+    #[must_use]
+    pub(crate) fn new(
```
CuzImClicks: why pub crate? are plugins not allowed to create fishing hook entities? xD

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,893 @@
...
+    /// Sets the projectile owner and mirrors vanilla's `Player.fishing` update.
+    pub(crate) fn set_owner(self: &Arc<Self>, owner: &SharedEntity) {
+        self.set_owner_entity(Some(owner));
+        if let Some(player) = owner.as_player() {
+            player.set_fishing_hook(self);
+        }
+    }
+
+    /// Determines if the player should stop fishing and removes the entity if so.
+    fn should_stop_fishing(&self, owner: &Player) -> bool {
```
CuzImClicks: consider taking the main hand item and the offhand item as parameters instead of locking. this makes the caller handle the lock explicitly. this often prevents deadlocks.

CuzImClicks (commented):
this is all i have time for right now, thanks for your work and keep it up :)

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,893 @@
...
+            state.time_until_lured = rng().random_range(FIVE_SECONDS..=THIRTY_SECONDS);
+            state.time_until_lured -= state.lure_speed;
+        }
+    }
+
+    /// Calculates if the area the player is currently fishing in is open water.
+    fn calculate_open_water(&self, pos: BlockPos) -> bool {
+        let mut prev_layer = OpenWaterType::Invalid;
+
+        for _ in -1..=2 {
```
Joniii11: This is not vanilla behavviour Y loop is ignored so it checks y+2 4 times instead of 4 different layers also vansilla checks empty block collision shape not empty fluid state

steel-core/src/behavior/items/fishing_rod.rs
```diff
@@ -0,0 +1,144 @@
...
+                &ENTITY_FISHING_BOBBER_THROW,
+                SoundSource::Neutral,
+                player.position(),
+                0.5,
+                pitch,
+                None,
+            );
+
+            let player_pos = player.position();
+            let spawn_pos = DVec3::new(player_pos.x, player.get_eye_y() - 0.1, player_pos.z);
```
Joniii11, line 62: Hook launch is not vnilla FishingHook has own spawn offset and velocity formula but this uses generic projectile throwing also the trajectory is different on every cast

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,893 @@
...
+        id: i32,
+        position: DVec3,
+        world: Weak<World>,
+    ) -> Self {
+        Self {
+            base: EntityBase::new(id, position, entity_type.dimensions, world),
+            entity_type,
+            entity_data: SyncMutex::new(FishingBobberEntityData::new()),
+            projectile_base: ProjectileBase::new(),
+            hook_state: SyncMutex::new(FishingHookState::new(0, 0)),
```
Joniii11, line 121: lure and luck are ignored here because every hook starts with (0, 0) vanilla gets both values from the fishing rod when creating the hook maybe do a // todo or // fixme whatever if that foundation doesnt exist yet lol

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,893 @@
...
+            && let Some(player) = owner.as_player()
+            && !Self::should_stop_fishing(self, player)
+        {
+            let hooked_in = {
+                let hook_state = self.hook_state.lock();
+                hook_state.hooked_in.clone()
+            };
+
+            if let Some(hooked_in) = hooked_in {
+                self.pull_entity(&hooked_in);
```
Joniii11: Missing Vanilla entity event 31 after pulling hooked entity vnilla sends this event and client uses it for reel movement so packet flow is different

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,893 @@
...
+
+                        let removed = hooked.is_removed();
+                        let can_interact = hooked.can_interact_with_level();
+
+                        if !removed && can_interact {
+                            let pos = hooked.position();
+                            let height = hooked.bounding_box().height();
+
+                            self.try_set_position(DVec3::new(pos.x, pos.y + height * 0.8, pos.z))
+                                .expect("...");
```
Joniii11: Vanilla checks that hooked entity is still in same dimension and detaches otherwise Here this check is missing and expect() can also panic

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,893 @@
...
+                                && self.calculate_open_water(pos);
+                        }
+
+                        if is_in_water {
+                            state.out_of_water_time = (state.out_of_water_time - 1).max(0);
+                            if *self.entity_data.lock().fishing_hook().biting.get() {
+                                // If you don't find this random thing in the src, remove the "h", I had to correct this spelling mistake due to lint
+                                // TODO: -0.1 * this.synchronizedRandom.nextFloat() * this.synchronizedRandom.nextFloat()
+                                self.base.set_velocity(
+                                    self.base.velocity().add(DVec3::new(0.0, -0.1, 0.0)),
```
Joniii11: Bite movement is not vanilla uses two synchronized random floats here but Steel always applies fixed -0.1

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
tn-lorenz: I sometimes coin flip because idk what to choose 🤷‍♂️in this case it's really stupid I did this (I think I just copied it over) because _I know_ that I _will_ need to spawn a `FishingHookEntity` for a plugin I will make lmao ✌️💀

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,893 @@
...
+    fn entity_type(&self) -> EntityTypeRef {
+        self.entity_type
+    }
+
+    /// Responsible for all state-changes.
+    #[expect(
+        clippy::too_many_lines,
+        reason = "Logic that belongs together is being kept together."
+    )]
+    fn tick(&self) {
```
Joniii11, line 850: Missing projectile base tick here in vanilla calls super.tick() before FishingHook logic but this skips Steel projectile_base_tick()

Joniii11 (changes requested):

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
tn-lorenz: For the collision shape: I left a todo there because I literally don't know how to obtain it(from the state); do you have an idea?

steel-core/src/behavior/items/fishing_rod.rs
tn-lorenz: Yeah for the difference in rod trajectory: I left a TO-DO for this, because idk what the equivalent of `syncronizedRandom` is in steel: I asked this a few times on dc, but didn't get a satisfactory answer and I looked into it myself (to no avail) -- do you have an idea?

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
tn-lorenz: The problem is, I had it this way before, but it broke the automatically generated code, so I had to make it default 0,0 in `new`

steel-core/src/behavior/items/fishing_rod.rs
tn-lorenz: wdym? I stole this code

CuzImClicks: yea but still, i feel like this is a failure that should be documented in the logs with something more than just a debug? if it happens a lot then we should leave it at debug tho

tn-lorenz: I think this rarely happens ever, I haven't looked at what could fail in `try_add_entity` probably only world/dimension related stuff tho

steel-core/src/behavior/items/fishing_rod.rs
tn-lorenz: Do you mean I should take the item as a parameter? But this way I'd need to have to change the definition of `use_item` in the `ItemBehaviour` trait, wouldn't that be too much of a Mit dem Kopf durch die Wand kind of solution??

tn-lorenz: should I maybe get the `rod` from the `context ` (I noticed the `inv` field is an `InventoryAccess` now I gotta figure out what that is)

steel-core/src/behavior/items/fishing_rod.rs
tn-lorenz: > Hook launch is not vnilla FishingHook has own spawn offset and velocity formula but this uses generic projectile throwing also the trajectory is different on every cast

What you're saying doesn't seem to match what I find in: https://mcsrc.dev/2/26.2/net/minecraft/world/item/FishingRodItem
What am I missing?

tn-lorenz:
I'd be particularly interested in reviews regarding the new `fn`s I added (`shoot_from_player`, `update_owner_info`) and the use of `shoot_from_player` inside `fishing_rod.rs` (obtaining an `Arc<Player>` for the `fn` -> is this a good way to do it?). Of course, also a deeper look into vanilla parity.

steel-core/src/behavior/items/fishing_rod.rs
tn-lorenz: I did some changes to this section now; let me know it they are satisfactory.

kdcokenny (approved):

steel-core/src/behavior/items/fishing_rod.rs
```diff
@@ -0,0 +1,136 @@
...
+
+            if let Err(error) = world.try_add_entity(Arc::clone(&entity)) {
+                entity.set_removed(RemovalReason::Discarded);
+                log::error!("Failed to spawn fishing hook: {error}");
+                return InteractionResult::Fail;
+            }
+
+            player.award_stat(&vanilla_stat_types::ITEM_USED, item.item());
+
+            // TODO: vibration
```
coco875: add more precision to comment to better get why it's not done now

coco875 (dismissed):
look good, just one comment on a todo

coco875 (dismissed):

coco875 (dismissed):

steel-core/src/entity/entity/mod.rs
```diff
@@ -3615,6 +3615,33 @@ pub trait Entity: EntityEventSource + ErasedType + Send + Sync + 'static {
...
 
+    // This already exists for structures and AABB whatever that might be, but I also need it for entities, I hope this is the right spot to put it.
+    /// Calculates the squared Euclidean distance from this entity's position to the given position.
+    fn distance_to_sqr(&self, pos: DVec3) -> f64 {
+        let dx = self.position().x - pos.x;
+        let dy = self.position().y - pos.y;
+        let dz = self.position().z - pos.z;
+
+        dx * dx + dy * dy + dz * dz
+    }
```
JunkyDeveloper, line 3600: can this not be simplified like this?

Image

JunkyDeveloper:
in my commit I removed duplicate code from end cristal and raw entity

JunkyDeveloper (changes requested):
I really really want to approve but this does not work https://www.youtube.com/shorts/iUEbHPonGKI
the problem is, that if I go through the portal the hook will be triggert and I go to the enderperl

please fix, or investigate what this causes and create an issue, so it is documented that this wrong behavior exists.
@kdcokenny supports you here :)

tn-lorenz:
> I really really want to approve but this does not work https://www.youtube.com/shorts/iUEbHPonGKI
> 
> the problem is, that if I go through the portal the hook will be triggert and I go to the enderperl
> 
> 
> 
> please fix, or investigate what this causes and create an issue, so it is documented that this wrong behavior exists.
> 
> @kdcokenny supports you here :)

This should be normal, because the hook is supposed to despawn once the hooked entity and owner are no longer in the same dimension; maybe this only applies to living entities tho, need to check

steel-core/src/entity/entity/mod.rs
tn-lorenz: Why get_center?

Btw I'm also thinking about replacing all degree to radians conversions with `to_radians()` that should be cleaner, right?

tn-lorenz: > can this not be simplified like this?
> 
> Image

no btw because this is not supposed to use the positions block's center

tn-lorenz:
> I really really want to approve but this does not work https://www.youtube.com/shorts/iUEbHPonGKI the problem is, that if I go through the portal the hook will be triggert and I go to the enderperl
> 
> please fix, or investigate what this causes and create an issue, so it is documented that this wrong behavior exists. @kdcokenny supports you here :)

This is not a bug from my PR as Kenny found out, so not a blocker for this PR. I'm going to fix the bug I found myself, the one RedCommander found and do some more cleaning up of the code, then it should be ready

steel-core/src/entity/projectile/mod.rs
```diff
@@ -443,9 +451,13 @@ pub trait Projectile: Entity + ProjectileEventSource {
     /// Casts the move vector and returns the nearest block/entity hit (vanilla
     /// `ProjectileUtil.getHitResultOnMoveVector` with `this::canHitEntity`).
     fn get_hit_result_on_move_vector(&self) -> Option<ProjectileHit> {
+        const MAGIC_MINIMUM: f64 = 1.0e-12;
```
coco875: x) wasn't necessary and often it's EPSILON by memory, to designate a a very small number

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,1211 @@
...
+pub const MAX_OUT_OF_WATER_TIME: i32 = 10;
+const MAX_DISTANCE_SQR: f64 = 32.0 * 32.0;
+
+const DMG_DEFAULT: i32 = 5;
+const DMG_ITEM_ENTITY: i32 = 3;
+const DMG_ON_GROUND: i32 = 2;
+const DMG_CAUGHT: i32 = 1;
+
+const DEGREE_180: f32 = 180.0;
+const DEGREE_360: f32 = 360.0;
```
coco875, line 48: we don't have this const somewhere ?

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,1211 @@
...
+
+                let Some(world) = self.level() else {
+                    return;
+                };
+
+                let splash_block_state =
+                    world.get_block_state(BlockPos::containing(fish_x, fish_y - 1.0, fish_z));
+
+                if splash_block_state.get_block() == &vanilla_blocks::WATER {
+                    if rng().random::<f32>() < 0.15 {
```
coco875: maybe also a magic number but very not necessary, you can just my comment as solve if you don't want ot

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,1211 @@
...
+    fn entity_type(&self) -> EntityTypeRef {
+        self.entity_type
+    }
+
+    /// Responsible for all state-changes.
+    #[expect(
+        clippy::too_many_lines,
+        reason = "Logic that belongs together is being kept together."
+    )]
+    fn tick(&self) {
```
coco875: maybe to reduce function size split each match of bobber_state into their own function ? to make it a bit more atomic and avoid have one big function that do the whole tick logic

coco875 (changes requested):

steel-core/src/behavior/items/fishing_rod.rs
```diff
@@ -0,0 +1,173 @@
...
+                pitch,
+                None,
+            );
+
+            let player_pos = player.position();
+            let spawn_pos = DVec3::new(player_pos.x, player.get_eye_y() - 0.1, player_pos.z);
+
+            // FIXME: we don't have enchantments yet, so I can't get the current `luck` or `lure_speed` of `rod`
+            let luck = 0;
+            let lure_speed = 0;
```
CuzImClicks:
```rust
item.apply_unconditional_enchantment_value_effects(
    EnchantmentEffectComponent::FishingLuckBonus,
    0.0,
);


item.apply_unconditional_enchantment_value_effects(
    EnchantmentEffectComponent::FishingTimeReduction,
    0.0,
);
```

i think this is what you need here

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,1211 @@
...
+
+    /// Creates a fishing hook entity from saved base data.
+    #[must_use]
+    pub fn from_saved(entity_type: EntityTypeRef, load: EntityBaseLoad) -> Self {
+        Self {
+            base: EntityBase::from_load(load, entity_type.dimensions),
+            entity_type,
+            entity_data: SyncMutex::new(FishingBobberEntityData::new()),
+            projectile_base: ProjectileBase::new(),
+            // FIXME: `lure_speed` and `luck` are taken from the existing rod, but auto generation fails when doing this, refer to: https://mcsrc.dev/2/26.2/net/minecraft/world/entity/projectile/FishingHook#L75
```
CuzImClicks, line 190: is this still true?

CuzImClicks (dismissed):

steel-core/src/behavior/items/fishing_rod.rs
tn-lorenz: What you suggested looks like we're _setting_ the values to the rod, but what we actually want to do here is _obtaining_ them. I just set them to 0 by default here, because afaik we don't have enchantments yet anyways (that's why I added `FIXME`)

The Java equivalents are:

`int lureSpeed = (int)(EnchantmentHelper.getFishingTimeReduction(serverLevel, itemStack, player) * 20.0F);`

`int luck = EnchantmentHelper.getFishingLuckBonus(serverLevel, itemStack, player);`

tn-lorenz: Btw I can't just get the saved `luck` and `lure_speed` levels from the rod here, because that's not the same I think

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
tn-lorenz: yes

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
tn-lorenz: Wouldn't it make more sense, to take the whole `match` out then? I think having one `fn` per match would be unnecessary and less readable.

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
tn-lorenz: Doesn't look like it, but there are numerous instances where `180.0` is used, same for `360.0`. I think this confirms that I should make a separate PR where I add these `const`s to util/math and make the code more uniform this way.

coco875: then I let you fix it in another pr

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,1211 @@
...
+        if let Some(owner) = self.get_owner()
+            && let Some(player) = owner.as_player()
+        {
+            let can_fish = {
+                let inventory = player.inventory.lock();
+                let mainhand_item = inventory.get_item_in_hand(InteractionHand::MainHand);
+                let offhand_item = inventory.get_offhand_item();
+
+                !self.should_stop_fishing(player, mainhand_item, offhand_item)
+            };
```
coco875: hmm maybe create an inside function can_fish(self, player)

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,1211 @@
...
+                        state.life += 1;
+                        state.life >= ONE_MINUTE
+                    };
+
+                    if should_remove {
+                        self.set_removed(RemovalReason::Discarded);
+                    }
+                } else {
+                    self.hook_state.lock().life = 0;
+                }
```
coco875: hmm update_life ?

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
tn-lorenz: Unless changes to how generation works have been merged??

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,1211 @@
...
+                let mut liquid_height: f32 = 0.0;
+                let pos = BlockPos::from(self.base.position());
+
+                if let Some(world) = self.level() {
+                    let block_state = world.get_block_state(pos);
+                    let fluid_state = block_state.get_fluid_state();
+
+                    if fluid_state.is_water() {
+                        liquid_height = get_height(&world, pos, fluid_state);
+                    }
```
coco875: maybe something like that
liquid_height = { if fluid_state.is_water() { get_height(&world, pos, fluid_state) } else { 0.0} }

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,1211 @@
...
+
+                    let is_in_water = liquid_height > 0.0;
+
+                    let bobber_state = {
+                        let state = self.hook_state.lock();
+                        state.bobber_state
+                    };
```
coco875: get_bobber_state()

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,1211 @@
...
+                                    )));
+                                }
+
+                                self.catching_fish(pos, &mut state);
+                            } else {
+                                state.out_of_water_time =
+                                    (state.out_of_water_time + 1).min(MAX_OUT_OF_WATER_TIME);
+                            }
+                        }
+                    }
```
coco875: update_bobber ?

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,1211 @@
...
+
+                    let hooked_in = {
+                        let state = self.hook_state.lock();
+                        state.hooked_entity.is_some()
+                    };
```
coco875: is_hooked_in()

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,1211 @@
...
+                    self.move_entity(MoverType::SelfMovement, self.base.velocity());
+                    self.apply_effects_from_blocks();
+                    self.update_rotation();
+
+                    let should_stop = {
+                        let state = self.hook_state.lock();
+
+                        state.bobber_state == BobberState::Flying
+                            && (self.base.on_ground() || self.base.horizontal_collision())
+                    };
```
coco875: should_stop()

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
coco875: yes after reading more deeply it would be better but I also suggest other change

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
tn-lorenz: Yeah I was thinking about `tick_bobber_state`

coco875: hmm bobber_state aren't just the bobber ?

steel-core/src/behavior/items/fishing_rod.rs
CuzImClicks:
```rust
    #[must_use]
    pub fn apply_unconditional_enchantment_value_effects(
        &self,
        component: EnchantmentEffectComponent,
        input: f32,
    ) -> f32 {
        let Some(enchantments) = self.get_enchantments() else {
            return input;
        };

        let mut value = input;
        for (key, level) in enchantments.iter() {
            if *level == 0 {
                continue;
            }
            let Some(enchantment) = REGISTRY.enchantments.by_key(key) else {
                continue;
            };
            let level = *level as i32;

            for effect in enchantment.effects.value_effects(component) {
                if !effect.is_unconditional() {
                    continue;
                }
                if let Some(updated) = effect.effect.process_without_random(level, value) {
                    value = updated;
                }
            }

            let Some(effect) = enchantment.effects.single_value_effect(component) else {
                continue;
            };
            if let Some(updated) = effect.process_without_random(level, value) {
                value = updated;
            }
        }

        value
    }
```

this is exactly what you need i think. it gives you the values all of the enchantments add to your luck or lure stat

it has a weird name but it doesnt set anything, it computes what you need

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
tn-lorenz: why? idk if this is still vanilla

coco875: just to avoid let mut liquid_height, (it's only set once)

coco875 (approved):

steel-core/src/entity/entities/objects/projectiles/fishing_hook.rs
```diff
@@ -0,0 +1,1237 @@
...
+            base: EntityBase::new(id, position, entity_type.dimensions, world),
+            entity_type,
+            entity_data: SyncMutex::new(FishingBobberEntityData::new()),
+            projectile_base: ProjectileBase::new(),
+            hook_state: SyncMutex::new(FishingHookState::new(0, 0)),
+            synchronized_random: SyncMutex::new(LegacyRandom::from_seed(0)),
+        }
+    }
+
+    // TODO: as soon as we have enchantments: use `luck` and `lure_speed` as input
```
CuzImClicks: remove the todo since you implemented it

JunkyDeveloper (approved):

PR #257, opened 2026-07-14
chore: clean custom payload log

chocodev11:


4lve:
Could you also make the `player x used mainhand` log into a debug log instead of info?

PR #259, opened 2026-07-14
add more permission to clear

coco875:
like mentionned in the other I add the ability to add more permission to clear so

- minecraft.command.clear — clear any player.
- minecraft.command.clear.self — clear only yourself.
- minecraft.command.clear.group.<group> — clear players belonging to that group.

All effective target groups count, including default, assigned, and inherited groups. Permission is required for every effective group, so access to default does not permit clearing a player who is also op.

All selected targets are checked before inventories are changed, preventing partial execution.

more example:
- minecraft.command.clear.* — grants all target-scoped clear permissions.
- minecraft.command.clear.group.* — grants access to every group, but not .self.
- minecraft.command.clear.group.admin — grants only the admin group.
- minecraft.command.clear — remains the broad grant for every clear target.

### DIFF
--- a/steel-core/src/command/builtins/clear.rs
+++ b/steel-core/src/command/builtins/clear.rs
@@ -15,11 +15,10 @@ use super::super::{
     registration::CommandRegistration,
     target_permission::{can_target_player, player_target_access},
 };
-use crate::permission::PermissionKey;
+use crate::permission::{PermissionKey, PermissionKeyError};
 use crate::{entity::Entity as _, player::Player};
 
-pub(super) fn registration()
--> Result<CommandRegistration<CommandSource>, crate::permission::PermissionKeyError> {
+pub(super) fn registration() -> Result<CommandRegistration<CommandSource>, PermissionKeyError> {
     let root = PermissionKey::parse("minecraft.command.clear")?;
     Ok(
         CommandRegistration::new(Identifier::vanilla_static("clear"), |_| command())
@@ -155,7 +154,7 @@ fn clear_players(
     Ok(count)
 }
 
-fn invalid_permission(error: crate::permission::PermissionKeyError) -> CommandSyntaxError {
+fn invalid_permission(error: PermissionKeyError) -> CommandSyntaxError {
     CommandSyntaxError::dynamic(format!("Invalid clear target permission: {error}"))
 }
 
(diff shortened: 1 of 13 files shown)


coco875:
will do give once this one are merge to re-use AnyDescendant

4lve:
I'm not sure if I like the idea of permission groups becoming a way to target users. I think for now only clear.self and clear.others would be best

coco875:
I think it's better to add group also to allow for example moderator to clear somone inventory but not other moderator for example (like as a revenge or other so to limit damage)

4lve:
I think that usecase is too niche to be in core

coco875:
if it's not in the core then how it can be add it externally (even with plugin) without ovewrite clear command ?

coco875:
and else the other are just minecraft.command.clear.group.*

4lve:
Could prob add an event handler for vanilla commands. However I do not agree with using permissions groups as a way block certain actions like this. I think a better design would be another permission or meta

coco875:
? permission are always to grant authorisation to do action so I don't see why make permission to allow clear or not some other group make a problem ?

coco875:
I don't really get what you mean by "using permissions groups as a way block certain actions like this" and else it's still a relation so a permission would allow to by clearby that we put on the person that do the command or the person that can subice it

4lve:
I think a protected + bypass permission model would be cleaner than using groups.

Moderators could have a protected permission, and admins could have a bypass protected permission.

But I still think this is too specific to add only for the clear command. For now `clear.self` and `clear.others` should be enough.

coco875:
ok fine I am still not convince (especially that all the layer it like can we have multiple level of by pass to protect admin between them but allow the owner to still do it, etc)

Does I keep modification with AnyDescendant ?

4lve:
I think metadata would be a better way to support this in the future.

Groups are only an easier way to assign permissions to players. They should not have special meaning by themselves.

For example, a player could have metadata like `steel:protection_level: 3`, and commands could compare protection and bypass levels.

For now you remove `AnyDescendant` since it is no longer needed.

coco875:
hmm I should read doc about metadata then, else for protection_level the problem is change it can easily have side effect like if you put builder and moderator and helper on the same level but you want that moderator being able to change helper then you increase the level but it also mean that moderator can now affect builder, with group at least it's clear and direct I think personnally (maybe under the format of metadata).

ok then will do the change later

coco875:
also like you mention permission don't have clear hierachy so add level would add one indirectly without being the same for other kind of permission

PR #260, opened 2026-07-14
feat: Coral Block implementation

GreenedDev:


coco875 (approved):
(a bit surprising that the person who have done coral didn't do the block x)) else LGTM

PR #261, opened 2026-07-14
Implement all data components and use the new downcasting system

4lve:
Initially I intended to only switch to the new downcasting system. But then it ended up with me implementing all components since it's one of those things that would hinder future development and having stub implementations doesn't look that good.

I also switched from using a large generated Items struct to making each one a LazyLock. This was so item components can reference other items, like elytra with phantom membrane. It also makes it so Items don't need special treatment syntax wise compared to blocks.


steel-core/build/build.rs
```diff
@@ -72,10 +72,10 @@ pub fn main() {
     println!("cargo:rerun-if-changed={manifest_dir}/src/entity/entities");
 }
 
-/// Items use lowercase field names (`vanilla_items::ITEMS.stone`)
+/// Items use `SCREAMING_SNAKE_CASE` statics (`vanilla_items::STONE`)
```
coco875, line 75: hmm by memory it's call UPPER_CASE ? (on doc I found always different name for this kind of thing)

DarkMrMelther: UPPER_CASE it's any string with mayus SCREAMING_SNAKE_CASE is specifically snake case (this_one) + upper case (THIS ONE)

steel-core/src/command/execution/item.rs
```diff
@@ -184,44 +172,38 @@ pub(super) fn numeric_i32(tag: &NbtTag) -> Option<i32> {
...
+    if *key == vanilla_components::MINIMUM_ATTACK_CHARGE.key {
+        return value
+            .downcast_ref::<f32>()
+            .is_some_and(|value| value.is_finite() && !value.is_sign_negative() && *value <= 1.0);
+    }
+    if *key == vanilla_components::POTION_DURATION_SCALE.key {
+        return value
+            .downcast_ref::<f32>()
+            .is_some_and(|value| value.is_finite() && !value.is_sign_negative());
+    }
```
coco875: why not a match of *key ?

steel-registry/build_assets/data_components.json
coco875, line 1: you have modify the extractor ?

4lve: Yes, already commited. This one is only used in a test though so we can easily check for if new versions add new components

PR #262, opened 2026-07-14
Execute release workflows on tag push

DarkMrMelther:
Restricted some workflows to master branch only
Solved syntax error with tags preventing those workflows to run
Added dispatch to SteelDocs to re-deploy after release (See [SteelDocs#37](https://github.com/Steel-Foundation/SteelDocs/pull/37) for more information)


JunkyDeveloper (approved):

.github/workflows/release.yml
```diff
@@ -2,8 +2,10 @@ name: Build Release
 
 on:
   push:
+    branches:
+      - master
     tags:
```
coco875: for release, by memory we do the tag before doing pre-release no ? So it can make problem for download link update ?

JunkyDeveloper: I think we bumb the version and @4lve makes a git tag, that is what I think, and with the git tag the release will be build

coco875 (approved):
LGTM

4lve:
Thanks, how will this work with pushes to master that don't bump the version?

JunkyDeveloper (changes requested):
sorry need to revert, because of the comment of 4lve

DarkMrMelther:
> Thanks, how will this work with pushes to master that don't bump the version?

Both conditions need to be fullfilled to execute the workflow, so it only triggers on pushes that bump the version in master

4lve:
> > Thanks, how will this work with pushes to master that don't bump the version?
> 
> Both conditions need to be fullfilled to execute the workflow, so it only triggers on pushes that bump the version in master

Okay good

.github/workflows/release.yml
DarkMrMelther: Tags are assigned to a push, we do build, publish the release, and then notify the docs

JunkyDeveloper (approved):

PR #263, opened 2026-07-15
Feat: join/left message

chocodev11:
implement join and leave message. The join message look for cached player name, then change it if need, then send it. The leave message is sent on disconnect so theoretically they are online.


steel-core/src/server/mod.rs
```diff
@@ -4212,7 +4236,41 @@ impl Server {
...
             .into();
 
-        self.broadcast_to_online_with(|player| CSystemChat::new(&message, false, player));
+        self.broadcast_system_chat(&message);
+    }
+
+    /// Broadcasts a system chat message to every online player.
+    fn broadcast_system_chat(&self, message: &TextComponent) {
+        self.broadcast_to_online_with(|player| CSystemChat::new(message, false, player));
+    }
```
JunkyDeveloper: The function `broadcast_sprint_report` is also using `broadcast_to_online_with` and can be abstracted with that function, maybe search also for other occasions (don't have a good global search on github :/ )

JunkyDeveloper (changes requested):

steel-core/src/server/mod.rs
```diff
@@ -4212,7 +4236,41 @@ impl Server {
...
+
+    fn broadcast_player_join_message(&self, player: &Player, previous_name: Option<&str>) {
+        use steel_utils::translations;
```
JunkyDeveloper: I'm also a fan of having all use on top of the file because you have in two functions the same use. @4lve what is your thought about that?

steel-core/src/server/mod.rs
```diff
@@ -4212,7 +4236,41 @@ impl Server {
...
+            translations::MULTIPLAYER_PLAYER_JOINED_RENAMED
+                .message([display_name, TextComponent::plain(old_name.to_owned())])
+                .into()
+        };
+        let message = message.color(Color::Yellow);
+        self.broadcast_system_chat(&message);
+    }
+
+    fn broadcast_player_leave_message(&self, player: &Player) {
+        use steel_utils::translations;
```
JunkyDeveloper: second time

steel-core/src/server/mod.rs
4lve: Yea, always put our imports at the top. Exception for feature flagged functions

chocodev11: alr messrs

steel-core/src/server/mod.rs
chocodev11: what?

JunkyDeveloper: You can reuse your method again in the code. Because otherwise some places we use broadcast_to  direct and in your code we use your helper function. And I like to have it consistent. Better?

chocodev11: hmm ok, i just do it the way vanilla does

PR #267, opened 2026-07-15
fix: spawn main thread with larger stack to prevent Windows crash

AshuraV1:
fix : spawn main thread with larger stack to prevent windows crash
windows defaults to a 1 MB stack limit for the main thread. During startup, VanillaGenerator::new calls the transpiled spline density functions which have deep call chains in debug builds. This immediately overflows the stack and crashes the process on windows
Fixed by wrapping main() into steel_main() and running it inside a spawned thread with an explicit 8 MB stack limit (matching the rayon generator stack limit)
Also gated to cfg(all(windows, debug_assertions)) so other platforms and release builds call steel_main() directly


4lve:
Could you feature flag it so it only happens on dev builds with windows. Otherwise it will do the same for all builds on all architectures

chocodev11:
should we also flag the rayon stack limit for only windows?

4lve:
That one affects Linux as well

4lve:
Thanks for the PR! In the future make sure to create a new branch on your local repo when creating a PR. This makes it so I can make edits before merging if required

PR #268, opened 2026-07-17
feat: implement pumpkin block

Laptop59:
Implements the behavior of the pumpkin block, which includes the carving mechanic to obtain pumpkin seeds.
This PR also moves some code finding the horizontal direction of an entity to a new function `Entity::direction`, which reflects the Vanilla method `Entity.getDirection`.


steel-core/src/behavior/blocks/vegetation/sweet_berry_bush.rs
coco875, line 1: why there is sweet berry change ?

steel-core/src/entity/entity.rs
```diff
@@ -1685,6 +1685,14 @@ pub trait Entity: EntityEventSource + ErasedType + Send + Sync + 'static {
         self.base().set_rotation(rotation);
     }
 
+    /// Gets the nearest direction to the entity's yaw (horizontal rotation).
+    ///
+    /// This means that this function always returns a horizontal direction.
+    fn direction(&self) -> Direction {
+        let (yaw, _) = self.rotation();
+        Direction::from_yaw(yaw)
+    }
```
coco875: maybe name it direction_yaw because normally there is pitch so to do up down for stair etc

coco875 (changes requested):

steel-core/src/behavior/blocks/vegetation/sweet_berry_bush.rs
Laptop59: Sweet berries use the same Vanilla function `Block.dropFromBlockInteractLootTable` as the Pumpkin block to drop loot, so I did this to keep it consistent with Vanilla, as the function is being implemented here.

steel-core/src/entity/entity.rs
Laptop59: Maybe, but i'm not sure because Vanilla is using the name `getDirection` for its implementation.

Laptop59: But yeah, i think autocomplete will help developers find the function, now that I have renamed it.

coco875 (approved):
LGTM but maybe drop_from_block_interact_loot_table feel a bit empty compare to the number of argument

steel-core/src/behavior/blocks/vegetation/pumpkin_block.rs
```diff
@@ -0,0 +1,133 @@
...
+                    0.05,
+                    0.05 * z_offset + rng.random::<f64>() * 0.02,
+                ),
+                Arc::downgrade(world),
+            ));
+
+            if let Err(error) = world.try_add_entity(entity) {
+                log::warn!("Failed to drop item stack entity: {error}");
+            }
+        }
```
coco875, line 97: we don't have a common function for that ? because I think it will be shared across some block behavior

coco875: in the change I see world.pop_resource ? does it do what I ask/describe (I don't have the full code in front of my eye) ?

coco875 (changes requested):

steel-core/src/behavior/blocks/vegetation/pumpkin_block.rs
Laptop59: > we don't have a common function for that ? because I think it will be shared across some block behavior

I hardcoded this into the block because even `PumpkinBlock.java` hardcodes its exact constants in its behavior

> in the change I see world.pop_resource ? does it do what I ask/describe (I don't have the full code in front of my eye) ?

I'm pretty sure `pop_resource` already existed before my PR, and yeah, as its name suggests, it just pops out items from a given position in a certain way.

Laptop59: I believe we also don't want pickup delay for the pumpkin seeds dropped (from my testing)

coco875: hmm it's not because minecraft do that way that we should translate it 1:1 x) else if it's exactly the same code why not use pop_resource ? (does pumpkin pop_resource in a different way ?)

coco875: maybe try find a way add way to deduplicate code

Laptop59: > hmm it's not because minecraft do that way that we should translate it 1:1 x) else if it's exactly the same code why not use pop_resource ? (does pumpkin pop_resource in a different way ?)

Actually not the exact same code because the constants are a bit different, but maybe it does feel a bit duplicated

coco875: if it's just constant that are diffrent maybe make a generic that take this constant as argument so pop_resource() will be the same but pumpkin will have pop_ressource_generic<0.3, 0.4, etc>() (with pop_resource use the same pop_ressource_generic behind the scene) ?

Laptop59: > if it's just constant that are diffrent maybe make a generic that take this constant as argument so pop_resource() will be the same but pumpkin will have pop_ressource_generic<0.3, 0.4, etc>() (with pop_resource use the same pop_ressource_generic behind the scene) ?

Yeah that's an idea i should try

Laptop59: Okay, so I'll admit that I didn't try making a generic function for constants, but that's because I figured that it would be a better solution to just offload the item entity spawning parts to a different function `World.spawn_item_with_velocity` so that I can specify my own velocity and position while still keeping the same behavior with less duplicated code

coco875: I think it's better then a generic (that can be heavy to read)

coco875: the function are already being use in pop_ressource ?

Laptop59: > the function are already being use in pop_ressource ?

Looks like `World.pop_resource` calls `World.spawn_item`, which calls `World.spawn_item_with_velocity`.

coco875: ??? I guess that the whole call stack can be simplified (maybe open an issue to avoid to forget and avoid doing it in the same pr, or you can also doing it in the same pr)

Laptop59: > ??? I guess that the whole call stack can be simplified (maybe open an issue to avoid to forget and avoid doing it in the same pr, or you can also doing it in the same pr)

Not sure if it really needs to be simplified, because they are a bit different:
- `World.pop_resource` also checks for the `block_drops` gamerule and adds random values to the provided position of the item entity to create.
- `World.spawn_item` is basically just `World.spawn_item_with_velocity` but with its own random velocity values.

Laptop59: It also looks like the `block_drops` gamerule (the rule I mentioned in `World.pop_resource`) does not affect carving pumpkins, but affects drops from a sweet berry brush in Vanilla (with this PR it replicates this behavior as well)

coco875: by curiosity there is a mojira about that ? else the pr look good

coco875 (approved):

steel-core/src/behavior/blocks/vegetation/pumpkin_block.rs
Laptop59: > by curiosity there is a mojira about that ? else the pr look good

I searched "pumpkin" and did not find anything about block drops

coco875 (approved):

JunkyDeveloper (approved):
LGTM

PR #270, opened 2026-07-18
feat: Finalize Firefly Bush Block implementation

GreenedDev:
The animateTick function seems to be only clientside, so i haven't implemented it. the particles and the sound appear anyway.


coco875 (approved):
LGTM

Purdze (changes requested):
drop bonemeal_action_type as vanilla never overrides getType(), so firefly bush is a GROWER (only grass/netherrack/nylium are NEIGHBOR_SPREADER)

update_shape is still missing, so the bush floats when its support breaks,  use survival_update_shape like bush_block.rs does

is_bonemeal_success can go, the trait default is already true

PR #271, opened 2026-07-19
Harden foundation and fix race conditions

4lve:
## Overview

While looking into redstone, I found several foundational race conditions in packet handling and chunk scheduling. These issues can be incredibly rare and difficult to reproduce, but we still have to account for them.

This PR addresses them in three parts.

## Packet processing between game ticks

Previously, packet logic ran as soon as the server received a packet. This allowed packet handlers to run both alongside each other and alongside the game tick.

For example:

- Two players could use an empty bucket on the same water block at almost exactly the same time. Both handlers could observe the water, give their player a water bucket, and then replace the water with air.
- An entity could be killed by a packet while it was still ticking, leaving the tick code running with the assumption that the entity was alive.

Packet handlers now run between game ticks. When the next tick begins, the server drains the packets admitted for that phase and closes packet execution. Packets arriving after the cutoff remain queued until the next inter-tick phase.

This keeps packet handlers from overlapping the game tick without moving all packet-handling overhead onto the game tick like vanilla.

Each scheduled packet also has an explicit concurrency category:

- `PlayerLocal` handlers can run across players when they only touch player-owned or individually locked state.
- `Serialized` handlers run one at a time in submission order when they can mutate shared world or domain state. Movement belongs here because it can affect things like collisions and farmland.
- `Exclusive` handlers act as global barriers and run completely alone. This is currently used for operations such as combat where multiple pieces of shared state are involved.

Every new packet handler must now make an explicit concurrency decision. Hopefully, more handlers can move into `PlayerLocal` as their transactions are audited.

This also adds conditional block mutation helpers. Operations such as filling a bucket now only succeed if the observed block state was actually replaced, preventing two consumers from claiming the same block.

## Chunk lifecycle commits at game-tick boundaries

Chunk scheduling previously ran on its own tick loop outside the game tick. This was done because ticket propagation, chunk generation scheduling, and unload processing can be expensive.

However, it meant chunk lifecycle state could change halfway through a game tick. If a ticket disappeared after the game tick started but before the chunk scheduling tick, the scheduler could move the chunk into the unloading stage. The chunk still existed and could be recovered, but normal block access would treat it as unavailable and return air.

The expensive scheduler work still runs in the background, but lifecycle changes are now committed only at game-tick boundaries.

If a background scheduling epoch is not ready, the currently committed chunk state remains authoritative until a later boundary. This guarantees that a loaded chunk cannot move into unloading halfway through a game tick.

## Full-neighborhood ticking readiness

Keeping the currently loaded chunks stable is not enough on its own. A ticking chunk may interact with its neighbors—for example, when a piston pushes blocks across a chunk border.

Steel now follows the same general readiness model as vanilla:

- Block ticking requires the full radius-1 neighborhood, or 3×3 chunks, to be loaded.
- Entity ticking requires the full radius-2 neighborhood, or 5×5 chunks, to be loaded.
- Initial chunk sending is also gated on confirmed block-ticking readiness.

This readiness is tracked incrementally as full chunks are published or demoted, avoiding repeated neighborhood scans while preserving the guarantees needed for safe ticking.

## Additional changes

The rework also adds:

- Ordered per-player packet lanes.
- Bounded per-player and global packet admission.
- Safe handling of packets arriving during tick-boundary drains.
- Staged chunk revival preservation during unload processing.
- Persistent pending generation postprocessing.
- Queueing and execution of vanilla block events.
- Correct preservation of scheduled block and fluid tick phases.


steel-registry/src/world_clock.rs
```diff
@@ -30,7 +30,7 @@ impl WorldClockRegistry {
         Self {
             world_clocks_by_id: Vec::new(),
             world_clocks_by_key: FxHashMap::default(),
-            tags: FxHashMap::default(),
+            tags: crate::RegistryTags::default(),
```
coco875: why crate:: ? and just an use then just RegistryTags

package-content/config.toml
```diff
@@ -6,9 +6,9 @@
 server_port = 25565
 # Maximum number of players allowed on the server
 max_players = 20
-# Allow view_distance above vanilla's 32-chunk cap, up to 127.
+# Allow view_distance above vanilla's 32-chunk cap, up to 125.
```
coco875: by memore render distance that are send are an u8 so 127+2 should still be fine

steel-registry/src/world_clock.rs
4lve: Yea, will fix

package-content/config.toml
4lve: Will look into it, the reason of this change is that 126,127 is now reserved for block, entity ticking. So a view distance of 125 will be equivalent to 127 in steel

coco875: but the client will only see 125 chunks and not 127, it's view distance also so if we see simulation distance to 4 it would load more chunk normally

4lve: Will try to make it 127 chunk again then. Just note that it will cause 2000 more chunks to be loaded on the server :sweat_smile:

coco875: for render it shouldn't load that much no ?

coco875: because it's not simulation distance else it's a problem of server owner

4lve: Fixed, it's now 128

package-content/config.schema.json
```diff
@@ -225,11 +225,23 @@
...
               "type": "integer",
               "description": "Worker threads for the Rayon chunk generation pool",
               "minimum": 0,
               "default": 0
+            },
+            "chunk_encoding": {
+              "type": "integer",
+              "description": "Worker threads for the Rayon chunk encoding pool",
+              "minimum": 0,
+              "default": 0
```
coco875, line 244: we didn't have a bit too much worker that can over solicitate core of the cpu ?

coco875: and also I know that c2me also describe how the amount of core are choose might be good to also describe here to help select the correct amount for a server owner

steel-core/src/behavior/blocks/fluid/liquid_block.rs
```diff
@@ -257,7 +257,11 @@ impl BlockBehavior for LiquidBlock {
         }
 
         let air = REGISTRY.blocks.get_default_state_id(&vanilla_blocks::AIR);
```
coco875, line 259: we can't just do `vanilla_blocks::AIR.default_state()` like above ?

steel-core/src/chunk/chunk_access.rs
```diff
@@ -642,6 +646,21 @@ impl ChunkAccess {
         }
     }
 
+    pub(crate) fn set_block_state_if_unchanged(
+        &self,
+        pos: BlockPos,
+        expected_state: BlockStateId,
+        new_state: BlockStateId,
+        flags: UpdateFlags,
+    ) -> Option<LevelChunkBlockSetResult> {
```
coco875, line 655: a result or an enum wouldn't be better to know if was because the chunk are unloaded or other

steel-core/src/chunk/chunk_holder.rs
```diff
@@ -64,6 +65,49 @@ pub enum ChunkResult {
     Ok(ChunkStatus),
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(u8)]
+pub(crate) enum TickingReadiness {
```
coco875, line 70: weird wording and also why fix the repr ? normally the rust compiler are smart enough for that

steel-core/src/chunk/chunk_holder.rs
```diff
@@ -64,6 +65,49 @@ pub enum ChunkResult {
...
+    Unready,
+    BlockTicking,
+    EntityTicking,
+}
+
+/// Exact ticking-readiness generation captured by concurrent consumers.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct TickingReadinessSnapshot(u64);
```
coco875, line 78: why after use a full u64 to store two bit ??? a u32 would be better

coco875: also you can't do atomic enum ?

coco875: else to avoid the convertion between u64 and atomicU64 you can just this as U64 directly so like that you only do ticking_readiness.readiness() instead of using load, etc

steel-core/src/chunk/chunk_holder.rs
```diff
@@ -409,6 +485,47 @@ impl ChunkHolder {
...
+    pub(crate) fn is_full_status_initialized(&self) -> bool {
+        self.full_status_initialized.load(Ordering::Acquire)
+    }
+
+    pub(crate) fn transition_ticking_readiness(
+        &self,
+        target: TickingReadiness,
+    ) -> Option<TickingReadiness> {
+        let mut current = self.ticking_readiness.load(Ordering::Acquire);
+        loop {
```
coco875, line 503: it's active waiting ? a mutex or a semaphore wouldn't be better to lock instead of active waiting ?

steel-core/src/chunk/chunk_map.rs
```diff
@@ -46,11 +57,15 @@ use crate::chunk::{
...
-use crate::world::tick_scheduler::{BlockTick, FluidTick};
+use crate::world::tick_scheduler::{
+    BlockTick, FluidTick, advance_tick_containers, collect_ticks_to_run,
+};
 use crate::worldgen::{ChunkGeneratorType, WorldGenContext};
 use crate::{entity::Entity, player::Player};
 
 const GENERATION_THREAD_MULTIPLE: usize = 2;
+// Vanilla applies this limit independently to block ticks and fluid ticks.
+const MAX_SCHEDULED_TICKS_PER_TICK: usize = 65_536;
```
coco875, line 68: maybe in another pr add in config a way to raise this limit if we want to do some extreme testing

steel-core/src/chunk/chunk_map.rs
```diff
@@ -79,21 +94,53 @@ struct TickableChunk {
...
+    fn start(&self, tick: &BlockTick) {
+        self.world.start_scheduled_block_tick(tick);
+    }
+}
+
+impl Drop for BlockTickBatchGuard<'_> {
+    fn drop(&mut self) {
+        self.world.end_scheduled_block_tick_batch();
+    }
+}
```
coco875, line 116: why not make a trait instead of a structur that only have one field ?

coco875: (same for below)

steel-core/src/chunk/chunk_map.rs
```diff
@@ -1297,6 +1449,301 @@ impl ChunkMap {
...
+        self.apply_readiness_demotions(&candidates);
+        self.update_pending_readiness(&candidates);
+        Ok(())
+    }
+
+    fn reconcile_ticking_readiness(
+        &self,
+        changed_positions: &[ChunkPos],
+    ) -> Result<(), FullNeighborhoodError> {
+        let publications = self.full_publications.drain();
```
coco875, line 1496: I think there is a maybe a way to avoid create a vec

steel-core/src/chunk/chunk_map.rs
```diff
@@ -1503,23 +2044,82 @@ impl ChunkMap {
...
 
-    /// Sorts and executes all ready scheduled ticks, calling block/fluid behavior callbacks.
-    fn execute_scheduled_ticks(
-        world: &Arc<World>,
-        mut ready_block_ticks: Vec<BlockTick>,
-        mut ready_fluid_ticks: Vec<FluidTick>,
-    ) {
-        const MAX_TICKS: usize = usize::MAX; // Vanilla uses 65_536, the lion does not concern himself with vanilla hotpatching
+    /// Advances both active clocks and collects this tick's block batch.
+    fn collect_scheduled_block_ticks(tickable_chunks: &[TickableChunk]) -> Vec<BlockTick> {
```
coco875, line 2048: I think that in this function there is a bit too much allocation through collect function

steel-core/src/chunk/chunk_map.rs
```diff
@@ -1503,23 +2044,82 @@ impl ChunkMap {
...
+        };
+        for index in advanced_fluid_containers {
+            level_chunks[index].dirty.store(true, Ordering::Release);
+        }
+
+        collected.ticks
+    }
+
+    /// Collects this tick's fluid batch after block callbacks have run.
+    fn collect_scheduled_fluid_ticks(tickable_chunks: &[TickableChunk]) -> Vec<FluidTick> {
```
coco875, line 2093: same also I think a bit too much allocation

steel-core/src/chunk/chunk_map.rs
coco875, line 1: the file are too big, so in another pr also do what you have start to done so split code and test

steel-core/src/chunk/chunk_scheduler.rs
```diff
@@ -0,0 +1,308 @@
...
+
+/// Timing information for one background epoch and its boundary commit.
+#[derive(Debug, Default)]
+pub(crate) struct ChunkMapSchedulingTimings {
+    /// Time spent applying queued ticket operations and propagating their levels.
+    pub(crate) ticket_updates: Duration,
+    /// Time spent committing holder lifecycle changes at the game-tick boundary.
+    pub(crate) lifecycle_commit: Duration,
+    /// Time spent creating or updating chunk-generation tasks.
+    pub(crate) schedule_generation: Duration,
```
coco875, line 21: does this time are reported somewhere to allow to see in what condition what take the most time ?

steel-core/src/player/networking.rs
```diff
@@ -50,6 +51,261 @@ pub enum OutboundPacket {
...
+    /// tolerate cross-player execution order differing from packet submission order.
+    PlayerLocal,
+    /// The handler may overlap player-local work, but not another serialized handler. Serialized
+    /// handlers start in global packet submission order.
+    Serialized,
+    /// The handler is a global submission-order barrier and must not overlap scheduled work.
+    Exclusive,
+}
+
+enum ScheduledPlayPacketKind {
```
coco875, line 72: I think maybe with build.rs there is a cleaner way to do that and there is missing a bit of doc to explain for person that add other packet

steel-core/src/server/packet_processor.rs
```diff
@@ -0,0 +1,1609 @@
...
+        if became_ready {
+            Self::mark_ready(&mut state, sequence, key, execution);
+        }
+        match execution {
+            ScheduledPacketExecution::PlayerLocal => {}
+            ScheduledPacketExecution::Serialized => state.serialized.push(Reverse(sequence)),
+            ScheduledPacketExecution::Exclusive => state.exclusive.push(Reverse(sequence)),
+        }
+        let should_wake = became_ready && state.phase == PacketPhase::Open;
+        drop(state);
```
coco875, line 311: it's needed to manually drop here ?

steel-core/src/world/mod.rs
```diff
@@ -5152,13 +5242,25 @@ mod tests {
...
+    use crate::chunk::chunk_ticket_manager::{ChunkTicket, ChunkTicketLevel};
     use crate::entity::{EntityBase, entities::PigEntity};
     use crate::test_support::test_world;
 
     const FIRST_HALF: BlockLocalAabb = BlockLocalAabb::new(0.0, 0.0, 0.0, 0.5, 1.0, 1.0);
     const SECOND_HALF: BlockLocalAabb = BlockLocalAabb::new(0.5, 0.0, 0.0, 1.0, 1.0, 1.0);
     static SPLIT_BLOCK: &[BlockLocalAabb] = &[FIRST_HALF, SECOND_HALF];
 
+    fn advance_scheduling_until(world: &Arc<World>, mut ready: impl FnMut() -> bool) {
+        for _ in 0..10_000 {
```
coco875, line 5254: look like an active wait why not use a mutex ?

steel-core/src/world/tick_scheduler.rs
```diff
@@ -84,73 +86,220 @@ impl TickKey for FluidRef {
...
+/// The active-time clock advances only when this container's chunk is eligible
+/// for block ticks. A priority queue keeps future work ordered without scanning
+/// every pending tick each game tick.
 #[derive(Debug)]
 pub struct TickList<T: TickKey> {
-    ticks: Vec<ScheduledTick<T>>,
-    /// Dedup set keyed by `(BlockPos, TickKey::key())`.
-    scheduled: FxHashSet<(BlockPos, usize)>,
+    active_tick: i64,
+    ticks: BinaryHeap<QueuedTick<T>>,
```
coco875, line 235: why not just a sorted vec ?

steel-core/src/world/tick_scheduler.rs
```diff
@@ -159,55 +308,82 @@ impl<T: TickKey> TickList<T> {
...
-            }
+    /// Packs ticks as relative active-time delays in Vanilla saved-list order.
+    #[must_use]
+    pub(crate) fn pack(&self) -> Vec<SavedTick<T>> {
+        let mut ticks: Vec<_> = self.ticks.iter().collect();
+        ticks.sort_by(|a, b| {
+            a.tick
+                .sub_tick_order
+                .cmp(&b.tick.sub_tick_order)
+                .then_with(|| a.insertion_order.cmp(&b.insertion_order))
```
coco875, line 319: yeah a sorted vec would avoid that

coco875 (commented):
post mrege review

package-content/config.schema.json
4lve: The packet workers are just tokio tasks so they aren't real threads in that sense. The chunk encoding is another rayon runtime, here I need to take another look as using the rayon default might be overkill for larger systems.

As for how many threads to chose the defaults are supposed to provide that automatically, only reason why you would want less is if you run two servers on one machine for example, but those examples should live in Steel Docs

steel-core/src/behavior/blocks/fluid/liquid_block.rs
4lve: Good catch, would like to see if we can implement some clippy lint or something to be consistent here

steel-core/src/chunk/chunk_access.rs
4lve: Since this PR it's now impossible for chunks to unload while such an operation is performed. But until there is a reason as of why the caller would need to know the reason I don't think we should change it

steel-core/src/chunk/chunk_holder.rs
4lve: What about the wording do you find off? And rust defaults to pointer sized variant sizes iirc

coco875: Readiness

steel-core/src/chunk/chunk_holder.rs
4lve: Will look into the reason. Might be better to use an u8

steel-core/src/chunk/chunk_holder.rs
coco875: else being the size of a pointer aren't a problem because register are in 32-64 bit

4lve: > Readiness

Is describes if a chunk is ready to be ticked. So TickingReadiness makes sense

steel-core/src/chunk/chunk_map.rs
4lve: Yea, not sure if any vanilla contraptions use that many scheduled ticks though xD

steel-core/src/chunk/chunk_map.rs
4lve: Will take a look, iirc vanilla does some weird stuff here

steel-core/src/chunk/chunk_map.rs
4lve: Yea, might do a PR where I only split large files

steel-core/src/chunk/chunk_scheduler.rs
4lve: When ticks are slow it will log it **IF** you enabled log.extra in config.toml

steel-core/src/player/networking.rs
4lve: Yea, might be a better way to do it. The ScheduledPacketExecution part is intended to be a bit explicit so you actually have to think about which other packets your packet can overlap

steel-core/src/world/mod.rs
4lve: A bit misleading, only used in tests. The PR where I split files will come in handy here

steel-core/src/world/tick_scheduler.rs
4lve: BinaryHeaps are really cool. https://youtu.be/EF3Z4jdD1EQ?t=7829

4lve: Will double check though, a sorted vec might make more sense in some cases

coco875: by memory it's still a tree so it's sparse in memory while sorted vec also have O(log n) to search because of binary search etc but every element are next to each other so more cache friendly

PR #272, opened 2026-07-20
Dynamic config

JunkyDeveloper:
## Summary
The problem from pterodactyl that steel can't have an already existing config with missing settings. This PR allows to upload a minimal version like only a file with the port and it expands it to the full working config. This allows also for the future, if steel adds new settings to be automatically in configs from older versions.

I had two systems the first one, this version, which also adds the comments. The second version was a bit less code but the comments are not added which I think are helpful.
Also a log section will be added if not exists in the same code


prek.toml
coco875, line 1: what are prek for ?

coco875:
does it's also add comment ?

prek.toml
JunkyDeveloper: precommit, means it checks is typo running through are all tests running through, is clippy happy, is the format correct, adds an empty line to each file and more stuff. It is a git commit hook, so I don't need to do all of that things by hand for each commit so it does it automatically.

it already shows I'm not anymore alone who is using prek to have all of these checks. a commit not takes up to 2min :D I removed tests before someone else added it back, we will see.

Also my changes are doing tests and clippy in parallel and 2 other things in parallel

JunkyDeveloper:
it also adds the comments, for each settings it uses the config.toml in package-content as a reference

coco875:
yeah but if I want to cleanup config because I know then can I remove and stay remove but when it update I get new field with comment above so like that I can know what was added and also decide to remove the new comment or not

JunkyDeveloper:
yes and no, has your removed setting a default internal? than yes it stay away, does steel require this config without it doesn't boot it create that settings + the comment to it. your custom order will not preserved the order will be from package-content. If you only remove comments, nothing will happen

better understandable? I will add that to the PR description, was not thinking about that

coco875:
but then when it upgrade it reformat back on how default config are with your config then ? if you change order or remove it comment, all will be revert ?

JunkyDeveloper:
if a required config is missing, I think yes, but need to test it that's why it is in draft, I need to fully understand the code and maybe make same changes to it

JunkyDeveloper:
will restart

PR #274, opened 2026-07-21
Announce server-wide view distance instead of per-player min

Purdze:
Announce the server-wide view distance in CLogin and remove the per-player CSetChunkCacheRadius echo, so clients can discover the server's limit.


JunkyDeveloper:
This branch has conflicts, maybe you didn't have the newest version?

steel-crypto/src/key_store.rs
```diff
@@ -14,34 +16,36 @@ impl KeyStore {
...
         log::debug!("Creating encryption keys...");
-        let private_key = Self::generate_private_key();
+        let (private_key, public_key) =
+            rsa_utils::generate_key_pair().expect("Failed to generate a key");
 
-        let public_key = RsaPublicKey::from(&private_key);
-        let public_key_der =
-            crate::public_key_to_bytes(&public_key).expect("Failed to encode public key");
+        let public_key_der = rsa_utils::public_key_to_bytes(&public_key)
+            .expect("Failed to encode the public key as DER");
```
JunkyDeveloper: Sorry but it really looks like the current version maybe your branch was so outdated that your fix wasn't merged: https://github.com/Steel-Foundation/SteelMC/blob/master/steel-crypto/src/key_store.rs

JunkyDeveloper (changes requested):
please check the newest version of the main because your change looks like the same thing as your PR here

steel-crypto/src/key_store.rs
JunkyDeveloper: resolved

JunkyDeveloper (commented):
LGTM but not tested, so no approval

PR #275, opened 2026-07-22
removed some unused code

JunkyDeveloper:
a small clean up, what claude found, it still works and compiles

idk what mojang api is but looks like dead code


coco875 (approved):
Good find, did you also quickly try to see compilation time just in case it has decrease ?

JunkyDeveloper:
Not really is only 1 small dep and it was claude not me, it first deleted 10 files and later found only this one is possible XD

PR #276, opened 2026-07-22
remove duplicated code

JunkyDeveloper:
maybe I will delete the last two commits, because I don't see much benefit here.

Still all claude findings, it compiles and the CI passes also but As I said this PR still needs a lot of rework and cirtical thinking


JunkyDeveloper:
is not worth to fix

PR #277, opened 2026-07-22
cleanup steel folder

coco875:
I have use rustqual to see what improvement he think and ask ai to resolve most of them then I manually check.

more info : https://github.com/SaschaOnTour/rustqual

I have manually revert som SRP and IOSP violation because he check a bit too much manually trying to split struct that was already doing only one thing (like the CommandLogger) or complain because there is one if in a function where it was instruction. Outside of that he also suggest some very minor performance improvement (re-use allocation, avoid clone, etc) and IOSP that was useful to make the code more clear and function less long.

there is still a part where I have doubt it's in main.rs where it change output to result.


steel/src/config.rs
```diff
@@ -248,8 +248,10 @@ const fn default_log_file() -> bool {
     true
 }
 
+pub(crate) const DEFAULT_MAX_HISTORY: usize = 50;
+
 const fn default_max_history() -> usize {
```
4lve: Probably better to replace the function, now it causes duplication instead

steel/src/logger/state.rs
```diff
@@ -26,6 +26,11 @@ pub struct LogState {
     pub file: LogFile,
 }
 
+pub(crate) enum Delete {
+    BeforeCursor,
+    AfterCursor,
+}
```
JunkyDeveloper, line 32: does the whole delete thing makes any difference? like memory, etc?
Because I liked the syntax before more, so what was the problem what that tool had with that?

4lve (commented):
Haven't looked at everything yet but the parts I've seen look good

steel/src/config.rs
coco875: hmm true didn't take attention to that

steel/src/config.rs
```diff
@@ -530,47 +526,76 @@ fn load_or_create_groups(path: &Path) -> Result<PermissionGroupsConfig, String>
     Ok(config)
 }
 
+const MAX_VANILLA_VIEW_DISTANCE: u8 = 32;
+const MIN_COMPRESSION_THRESHOLD: u32 = 256;
+const MAX_COMPRESSION_LEVEL: i32 = 9;
```
JunkyDeveloper: maybe consts at the top of the file for easy to find for checking?

steel/src/config.rs
JunkyDeveloper: While thinking on other things but why do we validate our default config?

steel/src/main.rs
```diff
@@ -112,23 +114,38 @@ static ALLOC: dhat::Alloc = dhat::Alloc;
 #[global_allocator]
 static ALLOC: mimalloc::MiMalloc = mimalloc::MiMalloc;
 
+#[cfg(all(windows, debug_assertions))]
+const DEBUG_MAIN_THREAD_STACK_SIZE: usize = 8 * 1024 * 1024;
+
+#[cfg(feature = "deadlock_detection")]
+const DEADLOCK_CHECK_INTERVAL: Duration = Duration::from_secs(10);
```
JunkyDeveloper: also here consts maybe at the top of the file

steel/src/main.rs
```diff
@@ -165,20 +173,21 @@ fn steel_main() {
             .thread_name("chunk-worker")
             .enable_all()
             .build()
-            .unwrap(),
+            .map_err(|error| format!("failed to build chunk runtime: {error}"))?,
```
JunkyDeveloper, line 179: do we have here already the logger if yes maybe use that one here

steel/src/logger/suggestions.rs
```diff
@@ -16,6 +16,11 @@ use std::{
 };
 use steel_core::command::sender::CommandSender;
 
+const SUGGESTION_COLUMN_WIDTH: usize = 20;
+const MAX_SUGGESTION_ROWS: usize = 3;
+const SUGGESTION_ELLIPSIS_WIDTH: usize = 3;
+const PROMPT_RESERVED_ROWS: usize = 4;
```
JunkyDeveloper, line 22: also here to the top?

steel/src/config.rs
JunkyDeveloper: I really thought long term, but the default_max_history is used for serde or? That's why I didn't marked that one.

steel/src/logger/state.rs
coco875: it's deduplicate code see delete function below that replace pop_before and pop_after (and also think that delete are more meaning full then pop)

steel/src/logger/suggestions.rs
coco875: what do you mean ? Move this piece of code above ? (before use/include)

steel/src/config.rs
coco875: MAX_HISTORY are used for minecraft command by memory so to keep so how many message it keep but not sure but in any case it's a const (because of const)

steel/src/config.rs
coco875: because we don't trust ourself !!! (I don't know I think it's for double check just in case we don't have done a correct const DEFAULT_CONFIG: &str = include_str!("../../package-content/config.toml"); like we have nothing in the CI to check that package-content/config.toml are valid)

steel/src/main.rs
coco875: map_err don't panic or print nothing just tranform an option into result with a string for the error by memory so like there is ? it will be raise on the function above until main

steel/src/config.rs
JunkyDeveloper: to save cpu cycled maybe add it to the ci XD

steel/src/logger/state.rs
JunkyDeveloper: ok, I know the deleting, but I like the pop, it is also for sets and queues you pop and push. But it can be only a me problem

steel/src/main.rs
JunkyDeveloper: ahhh sorry, yeah mist it

steel/src/logger/suggestions.rs
JunkyDeveloper: sorry again, didn't saw it is line 19 already, I only remember sawing a } closing so I thought a method should be above (should have extended the view)

steel/src/config.rs
JunkyDeveloper: yes I know, my idea was, to comment the default max history and make the function pub and use it also there

coco875: it's used below as default in sede so can't remove the function

steel/src/config.rs
coco875: I think a test would be enough but will add it

coco875: ok there is already packaged_configs_parse who check it

steel/src/config.rs
JunkyDeveloper: as I said, remove the constant and make the function pub

JunkyDeveloper:
Lint is still failing :)

steel/src/config.rs
coco875: prefer for constante to use const variable then function personally

JunkyDeveloper: I know but that is not possible or? but at the end it is only one case

coco875:
now it work

steel/src/config.rs
coco875: I don't get what you mean ? you didn't ask just to remove the const and use the function instead ?

JunkyDeveloper: ignore my messages and be happy, it isn't that important to write more messages :)

JunkyDeveloper (dismissed):

JunkyDeveloper (approved):

PR #278, opened 2026-07-22
remove unused code

coco875:
similar to 277 but focus on deadcode and duplication, while doing it I found that most of them are because of unfinish system or future api, so I only remove what I found like non essential.

the rest are 
[rest.patch](https://github.com/user-attachments/files/30282180/rest.patch)


steel-registry/src/recipe/crafting.rs
```diff
@@ -19,18 +19,7 @@ pub enum CraftingCategory {
...
-    pub const fn parse_json(s: &str) -> Self {
-        match s {
-            "building" => Self::Building,
-            "redstone" => Self::Redstone,
-            "equipment" => Self::Equipment,
-            _ => Self::Misc,
-        }
-    }
-}
+impl CraftingCategory {}
```
JunkyDeveloper: Why is this needed?

JunkyDeveloper (dismissed):
LGTM, good deletion

JunkyDeveloper:
fix tests

coco875:
ok test was having a weid error so... it's why there is no commit to fix it

JunkyDeveloper (approved):
because I approved it already before

PR #283, opened 2026-07-23
Feat: dirt path

chocodev11:


JunkyDeveloper (dismissed):
LGTM

steel-core/src/behavior/blocks/vegetation/farmland_block.rs
```diff
@@ -102,7 +102,7 @@ impl FarmlandBlock {
     }
 
     /// Turns the farmland into dirt.
-    fn turn_to_dirt(
+    pub(crate) fn turn_to_dirt(
```
coco875: by memory common function like that are move to mod.rs, so maybe move it ?

coco875 (approved):

JunkyDeveloper (approved):
LGTM, block works, logic fits and is correct

PR #286, opened 2026-07-23
Menu Api

CuzImClicks:
this pr moves the wiring of the different components of the menu api into a central builder to improve the overall api from a user and developer standpoint. previously the wiring of the different behaviors would have to be done manually, which made it easy to accidentally write wrong code and was very boilerplate heavy.

here are a few explanations and examples for those who are new to menus 

the flow is as follows: 

>Only a `MenuBuilder` can create a `Menu` that wires a layout of `Slots` (that are views into `Containers`) to a `MenuKind` (behavior implemented by you). 

### impl Container
owns the actual `ItemStacks` and are wrapped by a `Shared<T>` (just a rename of `Arc<Mutex<T>>`; `.into_shared()`)

### impl Slot 
are views into containers; slots can choose whether or not to allow modification like placing or taking out. crucially this does not prevent on_slot_clicked from running, this enables armor slots and similar to chose which items are allowed to be placed and picked up. a nice side effect is that it also allows stuff like click guis where certain items become "buttons" that cannot be picked up.

### Menu
are the top level that wire the behavior, the layout and the kind together (can only be created by a MenuBuilder to enforce correctness).

### impl MenuKind
is where you actually can implement behavior in functions like on_tick and on_slot_clicked (...)

### MenuBuilder
only way to construct a menu. enforces as much correct behavior at build time as possible. and wires the different parts together automatically in the background. it prevents making invalid shift clicking routes, using the sections and regions with a menu they werent created for/by, etc.

---

this is a showcase using the excellent foundations of 4lve's containers, since both the target's inventory and the source's inventory are just containers, that are shared by arc mutexes, we can just create a menu and back it by the actual player inventory. 

image

```rust
fn invsee(container_id: u8, source: &Arc<Player>, target: &Arc<Player>) -> Menu {
    let mut b = MenuBuilder::new(&vanilla_menu_types::GENERIC_9X5, container_id);

    let target_ref = ContainerRef::from(target.inventory.clone());
    let target_inventory = b.player_inventory(&target.inventory); // place the other player's inventory in the first 4 rows

    let armor = b.custom_section(armor_slots(&target.inventory), [target_ref.clone()]); // place armor slots
    let offhand = b.custom_section( // place offhand slot
        [SlotType::Normal(NormalSlot::new(target_ref.clone(), 40))], // manually map it to slot 40
        [target_ref],
    );
    let crafting = b.section(target.crafting_container(), 4); // place the 2x2 crafting grid as a 1x4

    let viewer = b.player_inventory(&source.inventory); // place the normal player inventory
    b.route( // create a shift clicking all slots of the target inventory to the normal inventory 
        target_inventory.all(),
        [viewer.all()],
        FillDirection::Backward,
    );
    b.route( // then from armor offhand and crafting to the normal inventory
        [armor, offhand, crafting],
        [viewer.all()],
        FillDirection::Backward,
    );
    b.route( // then from the normal inventory to the three others
        viewer.all(),
        [target_inventory.all(), armor, offhand, crafting],
        FillDirection::Forward,
    );

    b.build(MenuKindType::custom(InvseeMenuKind {}))
}

#[derive(Debug)]
struct InvseeMenuKind {}

impl MenuKind for InvseeMenuKind {} // see second tutorial on how to use this
```

then this is a more complex example of the grid api to create the domain menu ive shared. there are two concepts at play here. 

painting => letting you place stuff in the background without having to provide a container.
placing => which lets you place containers in rectangles in the menu.

painted slots that do not get overwritten by a container automatically get placed into an anonymous container that only lives for the duration of the menu being open. the view into this container is using display slots only, which do not allow pickup up or placing under any circumstances

image

```rust
fn domain_menu(
    container_id: u8,
    player: Arc<Player>,
    current_world: &Arc<World>,
    source: &CommandSource,
) -> Menu {
    let mut b = MenuBuilder::new(&vanilla_menu_types::GENERIC_9X6, container_id);

    let server = source.server();

    let domain_names: Vec<String> = server
        .worlds
        .domain_names()
        .map(ToOwned::to_owned)
        .collect();

    let map: Vec<(Section, Vec<Arc<World>>)> = b.grid(6, |g| {
        g.paint_all(ItemStack::empty());
        g.paint(
            Rect::cols(..).rows(0),
            ItemStack::new(&vanilla_items::GRAY_STAINED_GLASS_PANE),
        );
        g.paint(
            Rect::cols(..).rows(5),
            ItemStack::new(&vanilla_items::GRAY_STAINED_GLASS_PANE),
        );
        g.paint(
            Rect::cols(8).rows(..),
            ItemStack::new(&vanilla_items::GRAY_STAINED_GLASS_PANE),
        );
        g.paint(
            Rect::cols(0).rows(1..5),
            ItemStack::new(&vanilla_items::GRAY_STAINED_GLASS_PANE),
        );

        domain_names
            .iter()
            .take(4)
            .enumerate()
            .map(|(i, domain_name)| {
                g.subgrid(Rect::cols(..8).rows(i + 1), |g| {
                    g.paint_all(ItemStack::empty());

                    let mut sign = ItemStack::new(&vanilla_items::OAK_SIGN);
                    sign.set(CUSTOM_NAME, domain_name.clone().into());

                    g.paint(Rect::cell(0, 0), sign);

                    let worlds = server.worlds.worlds_in_domain(domain_name);

                    let icons: Vec<ItemStack> =
                        worlds.iter().map(|w| icon(w, current_world)).collect();

                    let len = icons.len();

                    let container = SimpleContainer::from_items(icons).into_shared();

                    (
                        g.place(Rect::cols(1..(len + 1).min(6)).rows(0), container)
                            .display()
                            .region()
                            .single(),
                        worlds,
                    )
                })
            })
            .collect()
    });

    b.build(MenuKindType::custom(DomainMenuKind {
        map,
        server: server.clone(),
        player,
    }))
}
```

```rust
struct DomainMenuKind {
    map: Vec<(Section, Vec<Arc<World>>)>,
    server: Arc<Server>,
    player: Arc<Player>,
}

impl MenuKind for DomainMenuKind {
    fn on_slot_clicked(
        &mut self,
        _behavior: &mut MenuBehavior,
        _guard: &mut ContainerLockGuard, // handles all locks for us, meaning we dont have to lock any of the containers we added to the menu ourselves.
        click: Click, // the type of action coupled with the place where it happened
        _player: &Player,
    ) -> ClickOutcome { // the click outcome means whether or not the default vanilla behavior should run; crucially this also runs for the player inventory, since it is part of the "open menu". if we were to always return consume no vanilla behavior would ever run 
        let Some(index) = click.slot() else { // the slot affected
            return ClickOutcome::Fallthrough;
        };

        // --- everything below this line is just handling the domain switching
        let Some((section, worlds)) = self
            .map
            .iter()
            .find(|(section, _worlds)| section.contains(index))
        else {
            return ClickOutcome::Fallthrough;
        };

        if worlds.is_empty() {
            return ClickOutcome::Consume;
        }

        let world = &worlds[index - section.start()];

        if world.domain() == self.player.get_world().domain() {
            self.server.queue_world_change(
                self.player.clone(),
                WorldChangeRequest::WorldSpawn {
                    target_world: world.clone(),
                },
            );
        } else {
            match self
                .server
                .queue_domain_switch_to_world(self.player.clone(), world.clone())
            {
                Ok(()) => {}
                Err(e) => {
                    tracing::error!(e);
                }
            }
        }
        ClickOutcome::Consume
    }
}
```


tn-lorenz:
So cool, you're the goat for this!! I can't wait to build plugins using the new api 👀

notwouf:
that looks awesome, also your explanations are pretty clear, even I could get it. I think you would be of a great help to the documentoor guys ^^

JunkyDeveloper:
> that looks awesome, also your explanations are pretty clear, even I could get it. I think you would be of a great help to the documentoor guys ^^

He already promised it, so I only wait for that PR ;)

PR #287, opened 2026-07-24
Feat: organizing entitie

chocodev11:


steel-core/src/entity/entities/objects/explosives/end_crystal.rs
coco875, line 1: hmm end_crystal I would put in the same categorie of tnt, so maybe explosion like fireball, and reserve boss for guardian, ender dragon, wither, (forget the name of the blue mob)

steel-core/src/entity/entities/objects/items/experience_orb.rs
coco875, line 1: I wouldn't put it in item but don't know what other categorie to fit

steel-core/src/entity/entities/neutral/mod.rs
```diff
@@ -0,0 +1 @@
+//! Neutral entity implementations.
```
coco875: maybe give some example ? to help contributor help place their entity in the correct category

chocodev11: which angry only if you hit them first

coco875: (I mean add it in the comment to make it more clear)

chocodev11: they have to know huh, like every other? or i will add comments to every category?

coco875: it would be only to category that can be ambigious, so neutral and passive ?

chocodev11: wdym

coco875: add a small description of the categorie to neutral and passive (so neutral say that it only become agressif when hit and passive never get agressif)

JunkyDeveloper (dismissed):
I'm ok now with that order

JunkyDeveloper (dismissed):

JunkyDeveloper:
fix lint

coco875 (approved):
look good

JunkyDeveloper (approved):

PR #292, opened 2026-07-26
feat: add reverse search

teknostom:
Adds reverse search like reverse-i-search.

Also, CTRL+C now clears buffer when something is written, and only exits if buffer is empty.


steel/src/logger/input.rs
coco875, line 87: I didn't know we have that maybe add a doc page about it would be helpful if user want to use all keybind

steel/src/logger/search.rs
```diff
@@ -0,0 +1,168 @@
...
+        let from = search.current.as_ref().map_or(0, |found| found.entry + 1);
+        Self::update(state, from)
+    }
+
+    pub(super) fn push(state: &mut LogState, text: &str) -> Result<()> {
+        let Some(search) = &mut state.search else {
+            return Ok(());
+        };
+        search.query.push_str(text);
+        let from = search.current.as_ref().map_or(0, |found| found.entry);
```
coco875, line 51: there is code duplication with this line maybe make a function get from ?

coco875 (changes requested):
only one thing else it look mostly good

JunkyDeveloper (approved):

steel/src/logger/search.rs
JunkyDeveloper: my problem here is, in the middle of the function it has different logic. an option would doing it via an enum, but it will not look as good as here in my opinion, for me that is the best solution and there are only 2 lines before and other the different code duplicated.

PR #293, opened 2026-07-27
Chests, furnaces and harden loot tables alongside maps

4lve:
This PR adds vanilla-compatible chests and furnaces.

Chests and barrels now resolve world-generated loot tables using their supplied loot seed. Since some loot tables like shipwrecks generate treasure maps, this PR also adds the map functionality, including map creation, persistence, rendering, tracking, and client synchronization.


4lve:
This PR quickly went out of scope since I wanted fo fully implement chests which had many implications when it came to loot tables. However this is too much to review in a single PR and especially something I'm not comfortable which just quickly merging before the announcment. I will leave the basic_containers branch up as it could help people who want to implement maps, or harden our loot tables, but these should be done as seperate PRs.

PR #295, opened 2026-07-28
feat: implement statistics

Laptop59:
## Type of change

- [x] New feature

## Description

This PR implements the foundation for [statistics](https://minecraft.wiki/w/Statistics). Statistics are counters that track integers for different types of tasks.

- Implements a new registry `CustomStatRegistry`, which stores all the `CustomStat`s in the game (i.e. the stats in the **General** tab). The registry will be populated with entries built from an extractor JSON file.
- Implements a new registry `StatTypeRegistry`, which contains all the `StatType`s that can be used for a stat. Each stat type is associated with a particular registry whose values will be used for identifying each stat in that stat type.
- A `Stat` is composed of its type (`StatType`) and value (depends on the stat type, like block, item, entity, etc.)
- Allows the server to send stats to the client and the client to request stats from the server. The server also refreshes the stats in case the player switches to another domain.
- The only stat awarded by this PR is the `Games Quit` stat in **General**. Other stats have not been implemented yet, which will be done in another PR. This PR mainly focuses on implementing the foundation, not the usages.
- This PR also serializes & deserializes stats for each player and changes the structure of player data. More information is given in **Additional Notes**.

## How this was tested

1. Join the server
2. Enter the pause menu and click the Statistics button. The client is quickly provided the stats instead of downloading endlessly. For new players, the stats should be empty for now and the **Items** and **Mobs** tabs should be grayed out.
3. Quit the server once and rejoin. Now check the statistics again and see that the counter next to `Games Quit` has incremented, i.e. is now `1`.
4. Repeat step 3 a few more times and check to ensure that the counter increments properly.
5. Go to another dimension like The Nether and back if you want. The counter **should not** increment.
6. Stop the server and restart it. Join it and ensure that the player data and stats have been saved and loaded properly.

## Screenshots / logs

stats1
stats2

## Checklist

- [x] Self-reviewed the diff

## Additional notes

This PR also changes how player data is stored: the folder `saves/<domain>/players` now stores two folders instead of player `.dat` files, which are now stored in `saves/<domain>/players/data` instead. This is to put statistic `.toml` files in `saves/<domain>/players/stats`  for each player. Therefore, player data will probably not load correctly from previous runs from other PRs the first time the branch is run.

The extractor code required for this PR is located at [SteelExtractor PR #23](https://github.com/Steel-Foundation/SteelExtractor/pull/23).


JunkyDeveloper:
I tested your PR, we need to define one thing, is Quite Game is quite dimension or quite server?
because currently it is quite domain, so with `/domain <other>` I switch between domains and the counter goes up and up and aup

Laptop59:
> I tested your PR, we need to define one thing, is Quite Game is quite dimension or quite server? because currently it is quite domain, so with `/domain <other>` I switch between domains and the counter goes up and up and aup

Yeah, I did this on purpose, because the counter is separate for each domain. So it would make sense that if I switch from one domain to another, I technically quit the previous domain.

steel-core/src/player/player_data_storage/mod.rs
```diff
@@ -280,12 +282,44 @@ impl FilePlayerDataStorage {
...
+        &self,
+        domain: &str,
+        uuid: Uuid,
+        data: &PersistentPlayerData,
+    ) -> io::Result<()> {
+        let player_stats_file = PlayerStatsFile::from_persistent_stats(&data.stats)?;
+        if let Some(toml_string) = serialize_player_stats_file(&player_stats_file) {
+            let final_path = Self::player_stats_file(&self.domain_players_dir(domain), uuid);
+            let lock = self.file_lock(&final_path);
+            let _guard = lock.lock().await;
```
coco875, line 316: outside of pr scope but wonder if we can simplify this into one thing to avoid get_lock and lock it

steel-registry/build/build.rs
```diff
@@ -47,6 +47,7 @@ mod fluids;
 mod cat_sound_variants;
 mod chicken_sound_variants;
 mod cow_sound_variants;
+mod custom_stats;
```
coco875: why custom ? and not just stats ?

steel-registry/build_assets/custom_stats.json
coco875, line 1: it's extracted ?

steel-registry/src/registry/mod.rs
```diff
@@ -222,6 +224,8 @@ pub const PLACED_FEATURE_REGISTRY: Identifier =
 pub const STRUCTURE_REGISTRY: Identifier = Identifier::vanilla_static("worldgen/structure");
 pub const STRUCTURE_PROCESSOR_LIST_REGISTRY: Identifier =
     Identifier::vanilla_static("worldgen/processor_list");
+pub const CUSTOM_STAT_REGISTRY: Identifier = Identifier::vanilla_static("custom_stat");
+pub const STAT_TYPE_REGISTRY: Identifier = Identifier::vanilla_static("stat_type");
```
coco875: why two kind of stat ?

steel-registry/src/stat/registry.rs
```diff
@@ -0,0 +1,268 @@
...
+pub trait StatValueRegistryEntry: Send + Sync + 'static {
+    // The functions here are prefixed so that it doesn't conflict
+    // with those of RegistryEntry.
+    fn stat_value_key(&self) -> &Identifier;
+    fn stat_value_id(&self) -> usize;
+}
+
+impl<E> StatValueRegistryEntry for E
+where
+    E: RegistryEntry + Send + Sync,
```
coco875, line 57: it can't be simplified to have the registry with RegistryEntry ?

steel-registry/src/stat/custom.rs
```diff
@@ -0,0 +1,48 @@
+use rustc_hash::FxHashMap;
+use steel_utils::{DowncastType, DowncastTypeKey, Identifier};
+
+/// A custom stat definition.
+#[derive(Debug)]
+pub struct CustomStat {
+    pub key: Identifier,
+}
```
coco875, line 14: it's really useful ot make a struct for only one field ?

steel-registry/src/stat/vanilla_stat_types.rs
```diff
@@ -0,0 +1,99 @@
...
+    StatType::new(Identifier::vanilla_static("custom"));
+
+/// Registers all vanilla stat types.
+///
+/// IMPORTANT: The registration order MUST match vanilla's Stats.java exactly,
+/// as the component's network ID is determined by its registration order.
+pub fn register_vanilla_stat_types(registry: &mut StatTypeRegistry) {
+    // 0: mined
+    registry.register(&BLOCK_MINED, || &REGISTRY.blocks);
+
```
coco875: why a new line here ?

steel-registry/src/stat/custom.rs
coco875, line 1: maybe precise somewhere that custom stat are just a registry for identifier for the general onglet

steel-core/src/player/stats_counter.rs
```diff
@@ -0,0 +1,274 @@
...
+
+/// Manages the counters for every statistic for a particular player.
+/// Analogous to Vanilla's `ServerStatsCounter.java`.
+pub struct StatsCounter {
+    /// The map of values stored in the counters for each stat that is
+    /// currently being tracked.
+    pub(super) stats: FxHashMap<Stat, i32>,
+
+    /// Stats that have been modified which haven't been updated to the client yet.
+    pub(super) dirty: FxHashSet<Stat>,
```
coco875, line 19: why not a tuple ?

coco875 (changes requested):
something that I don't put is maybe also add todo for stat that are missing but don't know where to put it

steel-registry/build_assets/custom_stats.json
Laptop59: Yes.

steel-registry/src/stat/registry.rs
Laptop59: This was done because `E` needs to implement `Send + Sync` so that `Stat` can be safe to send across threads, which will be stored in `StatsCounter`, which will be stored in `Player`, which should be safe to send across threads. `RegistryEntry` has no such requirement

steel-registry/src/stat/custom.rs
Laptop59: Vanilla uses `Identifier`, but what I found was that implementing `RegistryEntry` for `Identifier` doesn't really work: trying to implement `try_id` for `Identifier` is not logical. So I don't think there's a better way

steel-registry/src/stat/vanilla_stat_types.rs
Laptop59: Thought it would be a good idea to separate block-related stat types, item-related stat types, entity type-related stat types and custom-stat-related stat types from each other with a new line in-between.

As they are not necessary, I will remove them.

steel-core/src/player/stats_counter.rs
Laptop59: Looks like using a tuple for the map is indeed faster than keeping a map and a set. I'll modify the counter to use only one map. I used this approach because Vanilla uses it and it works fine.

steel-core/src/player/player_data_storage/mod.rs
Laptop59: Now the `file_lock` directly gives a guard, so tell me what you think of the new changes.

steel-registry/src/stat/registry.rs
coco875: oh ok, I see

steel-registry/src/stat/custom.rs
coco875: hmm then it's fine this way

steel-core/src/player/player_data_storage/mod.rs
JunkyDeveloper, line 1: My only problem is, the length of this file, maybe separate them into stats and player profile to have smaller files?
you even created a file maybe move more functions into this.

JunkyDeveloper (commented):
Looks really good but I would also like to have this split @coco875 what do you think?

steel-core/src/player/player_data_storage/mod.rs
coco875: hmm in any case in another pr but I agree that this file start getting to big so a refactor to split it are needed at this point

coco875 (dismissed):

coco875 (dismissed):
yeah we are above 2k and very close to 3k x)

steel-core/src/player/player_data_storage/stats.rs
```diff
@@ -0,0 +1,97 @@
...
+        for (stat_type_key, stats_in_stat_type) in self.stats {
+            let stat_type_entry = REGISTRY.stat_types.by_key(&stat_type_key).ok_or_else(|| {
+                io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    format!("unknown stat type {stat_type_key}"),
+                )
+            })?;
+
+            for (stat_value_key, count) in stats_in_stat_type {
+                let stat_value = stat_type_entry.value_from_key(&stat_value_key).ok_or_else(
```
Joniii11: one unknown stat drops all valid stats here vanilla keeps partial stats in ServerStatsCounter::parse() via resultOrPartial maybe skip only invalid entries

steel-core/src/player/player_data_storage/stats.rs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        Ok(Self { stats })
+    }
+
+    pub(super) fn into_persistent_stats(self) -> io::Result<Vec<PersistentStat>> {
+        let mut stats = Vec::new();
+
+        for (stat_type_key, stats_in_stat_type) in self.stats {
+            let stat_type_entry = REGISTRY.stat_types.by_key(&stat_type_key).ok_or_else(|| {
```
Joniii11: one unknown stat drops all valid stats here vanilla keeps partial stats in ServerStatsCounter::parse() via resultOrPartial maybe skip only invalid entries

steel-registry/src/stat/registry.rs
```diff
@@ -0,0 +1,278 @@
...
+        };
+
+        let entry_ref = Box::leak(Box::new(entry));
+        let id = self.stat_types_by_id.len();
+
+        let locked_ref = stat_type.stat_type_entry_ref.get_or_init(|| entry_ref);
+        assert_eq!(entry_ref, *locked_ref);
+
+        self.stat_types_by_id.push(entry_ref);
+        self.stat_types_by_key.insert(stat_type.key.clone(), id);
```
Joniii11, line 260: should duplicate keys be rejected here otherwise vec keeps both but by_key/id points to the last one

steel-core/src/player/player_data.rs
```diff
@@ -450,5 +477,20 @@ impl PersistentPlayerData {
...
+        {
+            let mut stats = player.stats.lock();
+
+            // This clears all counters, but also sets their dirty flag. This is important
+            // because the player needs to get notified that these stats got set to zero, for example
+            // after switching domains where the previous domain's state is remembered by the client.
+            // This means stale data is overwritten for the newer data the next time stats are sent.
+            stats.clear();
+
+            for PersistentStat { stat, count } in &self.stats {
```
Joniii11, line 489: old domain stats stay here as zero entries and then get saved into next domain this can make stats files keep growing

steel-registry/src/stat/vanilla_stat_types.rs
```diff
@@ -0,0 +1,96 @@
...
+    #[test]
+    fn registry_matches_extracted_stat_types() {
+        let entries: Vec<ExtractedStatTypeEntry> =
+            serde_json::from_str(include_str!("../../build_assets/stat_types.json"))
+                .expect("extracted stat types should be valid");
+
+        let mut registry = StatTypeRegistry::new();
+        register_vanilla_stat_types(&mut registry);
+
+        assert_eq!(entries.len(), 9);
```
Joniii11: magic number lol

steel-core/src/player/player_data_storage/file_storage.rs
```diff
@@ -0,0 +1,444 @@
...
+        data: &PersistentPlayerData,
+    ) -> io::Result<()> {
+        let player_stats_file = PlayerStatsFile::from_persistent_stats(&data.stats)?;
+        if let Some(toml_string) = serialize_player_stats_file(&player_stats_file) {
+            let final_path = Self::player_stats_file(&self.domain_players_dir(domain), uuid);
+            let _guard = self.file_lock(&final_path).await;
+            Self::write_atomic_path_locked(&final_path, toml_string.as_bytes()).await?;
+            log::debug!("Saved player stats for {uuid} in domain {domain}");
+        } else {
+            log::debug!("Did not save empty player stats for {uuid} in domain {domain}");
```
Joniii11: if stats become empty this keeps old stats file around maybe overwrite/remove it instead vanilla writes empty stats too

steel-registry/src/stat/vanilla_stat_types.rs
Laptop59: Removed the constant length check entirely, because the extracted stats will validate that.

steel-core/src/player/player_data_storage/file_storage.rs
Laptop59: Made an empty TOML file get serialized when it happens.

steel-registry/src/stat/registry.rs
Laptop59: Added a check for duplicate key.

JunkyDeveloper (approved):

coco875 (approved):

PR #296, opened 2026-07-28
Damage command

hthug06:
Implement The damage command, 100% vanilla parity

Nothing really changed, just needed to implement the damageType argument.


steel-core/src/command/builtins/damage.rs
```diff
@@ -0,0 +1,107 @@
...
+        ));
+    };
+
+    if target.hurt(&target_world, &damage_source, amount) {
+        // When using format! with a f32, if it's 1.0, it will just display 1 and not 1.0
+        // We just correct that so it's like vanilla
+        let mut display_amount = format!("{amount}");
+        if !display_amount.contains('.') {
+            display_amount.push_str(".0");
+        }
```
coco875: there is no wayt to just convert amount to float and then format so like that it will add .0 automatically without weird if on string ?

coco875 (changes requested):
Look mostly good, only one thing

steel-core/src/command/builtins/damage.rs
hthug06: Oh you were right, using the debug formatter format a float with the decimal, and even better, it format it with a maximal of 7 decimals like in vanilla !

4lve: Good, just know that we don't require perfect parity in such cases like number formatting

coco875 (dismissed):

steel-core/src/command/builtins/damage.rs
```diff
@@ -0,0 +1,103 @@
...
+    let Some(amount) = context.float("amount") else {
+        return Err(CommandSyntaxError::dynamic(
+            "Parsed value for amount is missing from the command context",
+        ));
+    };
+
+    // The base damage type for this command is generic
+    let damage_type = context
+        .damage_type("amount")
+        .unwrap_or(&vanilla_damage_types::GENERIC);
```
JunkyDeveloper, line 60: why here amount and not `damate_type

JunkyDeveloper (changes requested):

steel-core/src/command/builtins/damage.rs
hthug06: Little mistake, fixing it asap

JunkyDeveloper (approved):

coco875 (approved):

PR #298, opened 2026-07-29
Implement shulker box

Alvsch:
Implements the shulker box block and block entity.

## Known issues
- Starting of shulker box animations is not yet implemented

## Other changes
- Added `fn apply_components_from_item` to `BlockEntity` trait
- Added `Option<BlockEntity>` to `BlockLootContext`
- Added `fn contract` to `Aabb<DVec3, World>` matching the vanilla equivelant
- Added `fn prevents_block_drops` to `Player` which returns instabreak ability
- Added `fn to_nbt_tag_ref` to `LockCode`
- Fix ordering issue where `BlockEntity` was retrieved after its removal

Looking for feedback on overall implementation design as I fix the current issues


Alvsch:
I'm wondering if the current known issues should be fixed in another PR as this behavior is consistently missing across all container implementations.

JunkyDeveloper:
which issues do you mean?

Alvsch:
Animations doesn't work/isn't implemented (same with barrels). Looking into it is not so hard to implement.

trait Container needs start_open and stop_open functions to be added

JunkyDeveloper:
as a maintainer I would like to have it seperated into another PR, so it is easier to review and you can merge that branch already here.

I personally would also have the breaking changes in another PR, because the ice block impl was adding there some things and it took some time to figure that out while reviewing the ice block in general.

Because you can give also a more detailed description why is that needed, which would made the search way easier :)

But that is not a requirement, but a feedback from my side, maybe the other maintainer see it different :)

DarkMrMelther:
I think 4lve said that in his chest pr the counter of players with the container open was implemented

Alvsch:
> I think 4lve said that in his chest pr the counter of players with the container open was implemented

If you're talking about the `basic_containers` branch, that was closed due to scope creep iirc. This branch is however based on master after CuzImClicks Menu API changes was merged.

@JunkyDeveloper I've update the description to include all the changes that touches other parts than shulker boxes for me and anyone to look at. If the general consensus is to create a new PR for those changes I could do that.

JunkyDeveloper:
> @JunkyDeveloper I've update the description to include all the changes that touches other parts than shulker boxes for me and anyone to look at. If the general consensus is to create a new PR for those changes I could do that.

Thanks, will not have time in the next 10 days. but I think it will also help other maintainers :) at first we need to internally discuss all the things who we want to have it, but for now that is more than fine!

steel-core/src/behavior/block/context.rs
```diff
@@ -125,6 +136,10 @@ impl<'a> BlockLootContext<'a> {
         self.entity
     }
 
+    pub(crate) const fn block_entity(&self) -> Option<&'a dyn BlockEntity> {
+        self.block_entity
+    }
```
coco875, line 141: how it was done before ?

ArSiJa8:
everytthing works fine rom what i can see with my testing. All features are implemented besides the start of the animation (which is a known issue mentioned above)

> Starting of shulker box animations is not yet implemented

steel-core/src/inventory/menu/kinds/shulker_box_menu.rs
```diff
@@ -0,0 +1,80 @@
...
+            vanilla_items::ORANGE_SHULKER_BOX.id(),
+            vanilla_items::PURPLE_SHULKER_BOX.id(),
+            vanilla_items::YELLOW_SHULKER_BOX.id(),
+            vanilla_items::MAGENTA_SHULKER_BOX.id(),
+            vanilla_items::LIGHT_BLUE_SHULKER_BOX.id(),
+            vanilla_items::LIGHT_GRAY_SHULKER_BOX.id(),
+        ]
+    });
+    SHULKERS.contains(&stack.item().id())
+}
```
JunkyDeveloper: why not check against the Tag?

Image

JunkyDeveloper (dismissed):

steel-core/src/behavior/block/context.rs
Alvsch: This was not implemented before, making it impossible to access the BlockEntity of a broken block. This is similar to how we already adapted the vanilla implementation.

steel-core/src/inventory/menu/kinds/shulker_box_menu.rs
Alvsch: These are fabric modding tags and not in vanilla. This is the best solution I found however I can dig around a little further.

https://mcsrc.dev/2/26.2/net/minecraft/world/item/BlockItem#L196

Alvsch:
I don't think 26.3 is necessary as it only affects the behavior of shulker entities, not shulker boxes

JunkyDeveloper:
Thanks for checking! as a thank you for checken, this PR will be now prio for me to get it merged soon!

steel-core/src/behavior/blocks/container/shulker_box_block.rs
```diff
@@ -0,0 +1,201 @@
...
+            && let Some(container_ref) = block_entity.container_ref()
+            && can_open(state, world, pos, block_entity)
+        {
+            let inventory = player.inventory.clone();
+            player.open_menu(
+                TextComponent::translated(translations::CONTAINER_SHULKER_BOX.msg()),
+                move |context| shulker_box(inventory, context.container_id, container_ref),
+            );
+
+            // TODO: Award stat OPEN_SHULKER_BOX
```
CuzImClicks:
```suggestion
            player.award_custom_stat(&vanilla_custom_stats::OPEN_SHULKER_BOX);
```

dont forget to add the import

steel-core/src/behavior/blocks/container/shulker_box_block.rs
```diff
@@ -0,0 +1,201 @@
...
+            .map_or(0, |container| {
+                calculate_redstone_signal_from_container(container)
+            })
+    }
+}
+
+/// Builds the item form of a placed, possibly-filled shulker box.
+/// Vanilla `ShulkerBoxBlockEntity.collectComponents()` +
+/// `BaseContainerBlockEntity.collectImplicitComponents(CONTAINER)`.
+fn shulker_box_as_item(state: BlockStateId, shulker_box: &ShulkerBoxBlockEntity) -> ItemStack {
```
CuzImClicks: please make the helper functions in this file pub and then put them into the impl ShulkerBoxBlock part so that everyone can use them

steel-core/src/inventory/menu/kinds/shulker_box_menu.rs
CuzImClicks: `ItemTag::SHULKER_BOXES`

idk why junky used the c_shulker_boxes tag but this should be usable here?

CuzImClicks (changes requested):
great pr :)

CuzImClicks (dismissed):

steel-core/src/behavior/block/context.rs
```diff
@@ -129,6 +133,13 @@ impl<'a> BlockLootContext<'a> {
         self
     }
 
+    /// Adds the entity responsible for destroying the block.
```
Purdze: this adds the block entity at the broken position, not the entity responsible for destroying the block

steel-core/src/behavior/blocks/container/shulker_box_block.rs
```diff
@@ -0,0 +1,189 @@
...
+        0.5,
+        DVec3::from(pos.get_bottom_center()),
+    )
+    .deflate(1.0E-6);
+
+    let collision = WorldCollisionProvider::new(world);
+    !collision.has_block_collision(&lid_open_bounding_box)
+}
+
+impl BlockBehavior for ShulkerBoxBlock {
```
Purdze, line 83: overrides are missing from this impl:

`getShape` > `Shapes.create(shulkerBoxBlockEntity.getBoundingBox(state))`, steels equivalent is `get_collision_shape`, without it the open lid has no collision
`getBlockSupportShape` > `SHAPES_OPEN_SUPPORT.get(facing.getOpposite())` when not closed
`affectNeighborsAfterRemoval` > `Containers.updateNeighboursAfterDestroy`, so comparators next to a broken box currently won't update

steel-core/src/block_entity/entities/shulker_box.rs
```diff
@@ -0,0 +1,405 @@
...
+                    do_neighbor_updates(world, pos, state);
+                }
+
+                if animation.progress >= ANIMATION_STEPS {
+                    animation.animation_status = AnimationStatus::Opened;
+                    animation.progress = ANIMATION_STEPS;
+                    do_neighbor_updates(world, pos, state);
+                }
+
+                self.move_collided_entities(world, pos, state);
```
Purdze, line 146: deadlock.., the guard bound at line 153 is a parking_lot::MutexGuard which has a Drop impl, so nll can't release it early and it's still held here move_collided_entities then re-locks the same mutex at line 191, and SyncMutex = parking_lot::Mutex (steel-utils/src/locks.rs:12) isn't reentrant first box to enter Opening hangs the tick thread this is all one comment for line 169?

Purdze (changes requested):
really great pr, thanks! 

`ShulkerBoxBlock` is missing `get_block_entity_ticker` (see `furnace_block.rs:142`) and a `trigger_event` override delegating to the block entity.., see `bell_block.rs:313`.. without those two the ticker never registers and block events never land, so the animation can't work at all

Alvsch:
This will not compile, I've been trying to get the shapes and boxes working but I'm getting nowhere so I'm leaving my changes for someone else to fix up. I might check back on this tomorrow if no one else has time.

edit: I've intentionally not hooked up shulker block animations as no other container block has it and I don't want to increase the scope of this PR, instead leaving it for another PR.

CuzImClicks (approved):

Purdze (approved):

PR #299, opened 2026-07-30
feat: add mud block

SpikeHD:
Hello! Saw the post on Reddit and I like the approach and wanted to contribute :)

This implements the mud block, which is very simple, as I wanted to start small to learn the codebase/patterns


steel-core/src/behavior/blocks/building/mud.rs
coco875, line 1: mud don't have system where it dry (and become clay) ? (so likely with random tick I think) if it's the case then add todo for the tracker to say that the block aren't finish totally yet

coco875 (changes requested):

steel-core/src/behavior/blocks/building/mud.rs
coco875: oh you have just done it in your other pr x) and aren't implemented in mud behavior

coco875 (approved):
LGTM

JunkyDeveloper (approved):
LGTM

PR #300, opened 2026-07-30
feat: stalactite fluid transfer

SpikeHD:
Implements fluid transfer from stalactite to cauldrons, as well as mud -> clay behaviour

Vanilla parity notes:
* Flowing water isn't counted
* Tick is scheduled on the cauldron rather than being handled by the stalactite

Demo (with fast tickrate/random tick speed ofc)

<video src="https://github.com/user-attachments/assets/303c6c91-c85a-4242-9d23-800d5d1017db" />


steel-core/src/behavior/blocks/building/cauldron_block.rs
```diff
@@ -100,6 +138,31 @@ impl BlockBehavior for LayeredCauldronBlock {
...
+        else {
+            return;
+        };
+        let Some(fluid) = vegetation::get_cauldron_fill_fluid_type(world, stalactite_pos) else {
+            return;
+        };
+
+        if fluid == &vanilla_fluids::WATER {
+            let level = state.get_value(&BlockStateProperties::LEVEL_CAULDRON);
+            if level < 3 {
```
coco875: maybe change the 3 with a const like MAX_LEVEL_COULDRON ?

coco875 (changes requested):
Look good, only one magick number x)

steel-core/src/behavior/blocks/building/cauldron_block.rs
SpikeHD: fixed!

steel-core/src/behavior/blocks/vegetation/pointed_dripstone_block.rs
```diff
@@ -647,6 +662,168 @@ impl SpeleothemBlockBehavior {
...
+        state: BlockStateId,
+        world: &Arc<World>,
+        pos: BlockPos,
+        random_value: f32,
+    ) {
+        let Some(fluid_info) = self.get_fluid_above_stalactite(world, pos, state) else {
+            return;
+        };
+
+        let transfer_probability = if fluid_info.fluid == &vanilla_fluids::WATER {
```
coco875: maybe to avoid doing multiple time fluid_info.fluid == &vanilla_fluids::WATER store the result in a bool and re-use it ? (else you can do nothing like we only do the check two time)

steel-core/src/behavior/blocks/vegetation/pointed_dripstone_block.rs
```diff
@@ -647,6 +662,168 @@ impl SpeleothemBlockBehavior {
...
+        }
+        None
+    }
+
+    fn is_fillable_cauldron(state: BlockStateId, fluid: FluidRef) -> bool {
+        let block = state.get_block();
+        if fluid == &vanilla_fluids::WATER {
+            block == &vanilla_blocks::CAULDRON
+                || (block == &vanilla_blocks::WATER_CAULDRON
+                    && state.get_value(&BlockStateProperties::LEVEL_CAULDRON) < 3)
```
coco875: MAX_LEVEL_CAULDRON ?

coco875 (changes requested):
very small adjustement else it look good

steel-core/src/behavior/blocks/vegetation/pointed_dripstone_block.rs
SpikeHD: good point! ive extracted both lava and water states to `is_water` and `is_lava` since i find those more readable at a glance

steel-core/src/behavior/blocks/vegetation/pointed_dripstone_block.rs
SpikeHD: i found the cauldron max level in blockstateproperties and modified both instances to use that instead :+1:

coco875 (approved):
LGTM

steel-core/src/behavior/blocks/building/cauldron_block.rs
```diff
@@ -58,6 +63,39 @@ impl BlockBehavior for CauldronBlock {
...
+        } else if fluid == &vanilla_fluids::LAVA {
+            let new_state = vanilla_blocks::LAVA_CAULDRON.default_state();
+            world.set_block(pos, new_state, UpdateFlags::UPDATE_ALL);
+            world.game_event(
+                &vanilla_game_events::BLOCK_CHANGE,
+                pos,
+                &GameEventContext::new(None, Some(new_state)),
+            );
+            world.level_event(level_events::SOUND_DRIP_LAVA_INTO_CAULDRON, pos, 0, None);
+        }
```
JunkyDeveloper, line 97: I think a match would be here not better? if minecraft ever will add another fluid, or for modding support?

steel-core/src/behavior/blocks/building/cauldron_block.rs
```diff
@@ -100,6 +138,31 @@ impl BlockBehavior for LayeredCauldronBlock {
...
+                let new_state = state.set_value(&BlockStateProperties::LEVEL_CAULDRON, level + 1);
+                world.set_block(pos, new_state, UpdateFlags::UPDATE_ALL);
+                world.game_event(
+                    &vanilla_game_events::BLOCK_CHANGE,
+                    pos,
+                    &GameEventContext::new(None, Some(new_state)),
+                );
+                world.level_event(level_events::SOUND_DRIP_WATER_INTO_CAULDRON, pos, 0, None);
+            }
+        }
```
JunkyDeveloper, line 164: same as here

steel-core/src/behavior/blocks/vegetation/pointed_dripstone_block.rs
```diff
@@ -647,6 +662,170 @@ impl SpeleothemBlockBehavior {
...
+        let is_water = fluid_info.fluid == &vanilla_fluids::WATER;
+        let is_lava = fluid_info.fluid == &vanilla_fluids::LAVA;
+
+        let transfer_probability = if is_water {
+            WATER_TRANSFER_PROBABILITY_PER_RANDOM_TICK
+        } else if is_lava {
+            LAVA_TRANSFER_PROBABILITY_PER_RANDOM_TICK
+        } else {
+            return;
+        };
```
JunkyDeveloper, line 685: I think also a match looks here better :)

JunkyDeveloper (commented):
Please give a comment, what do you think about my requested changes

steel-core/src/behavior/blocks/building/cauldron_block.rs
coco875: for modding more then a match are needed but yeah it would be cleaner

SpikeHD:
@JunkyDeveloper @coco875  I would agree matches look better, but `WATER` and `LAVA` are statics, not enums/constants, so the way to `match` would be something like:

```rs
let transfer_probability = match fluid_info.fluid {
    fluid if fluid == &vanilla_fluids::WATER => WATER_TRANSFER_PROBABILITY_PER_RANDOM_TICK,
    fluid if fluid == &vanilla_fluids::LAVA => LAVA_TRANSFER_PROBABILITY_PER_RANDOM_TICK,
    _ => return,
};
```
which just feels a bit awkward compared to a direct comparison. Would this still be preferred?

JunkyDeveloper (approved):

PR #302, opened 2026-07-30
feat: implement /say, /me, and /msg chat commands

nglmercer:
## Summary

Implements the vanilla chat messaging commands:

- `/say <message>` — operator-only, broadcasts to all players and logs to the server console.
- `/me <message>` — available to all players, broadcasts an emote.
- `/msg <targets> <message>` — available to all players, with `/tell` and `/w` aliases. Sends an incoming whisper to each target and an outgoing echo back to the sender.

Command-generated chat is delivered as unsigned disguised chat (`CDisguisedChat`) bound to the matching vanilla chat type (`say_command`, `emote_command`, `msg_command_incoming`, `msg_command_outgoing`), so the client applies the standard decorations (`[%s] %s`, `* %s %s`, `%s whispers to you: %s`, `You whisper to %s: %s`).

## Changes

- `steel-core/src/command/builtins/message.rs` — new module with the three commands and graph-shape tests.
- `steel-core/src/command/execution/text.rs` — adds `CommandTextResolver::for_source` for resolving a message once against the command source.
- `steel-core/src/command/execution/source.rs` — removes the now-used `dead_code` expectation on `CommandSource::sender()`.
- `steel-core/src/command/builtins/mod.rs` — registers the commands and extends the command graph-shape test.

## Verification

- `cargo check -p steel-core` — pass
- `cargo fmt --all --check` — clean
- `cargo clippy -p steel-core --all-targets --all-features` — 0 warnings
- `cargo test -p steel-core --lib command::builtins` — 72 passed, including 3 new tests and the updated graph-shape test

## Note on vanilla verification

The chat-type bindings and decoration parameters were taken from the repository's generated data (`vanilla_chat_types.rs` and the generated translations) rather than transcribed from memory. `minecraft-src/` was not generated locally, so the `/msg` outgoing-echo binding follows stable vanilla behavior corroborated by the `msg_command_outgoing` parameter data (`[target, content]`); it can be confirmed against `MessageCommand.java` once the vanilla source is generated.


nglmercer:
no existing handling (or test) for multi-word component arguments

nglmercer:
is added a new multi-word component arguments `starts_structured_component` and `parse_structured_component`.

steel-core/src/command/execution/argument.rs
```diff
@@ -1173,16 +1193,21 @@ fn parse_component(reader: &mut StringReader<'_>) -> Result<TextComponent, Comma
             "Invalid text component cursor",
         ));
     }
-
-    let component = TextComponent::try_from_nbt(&tag).map_err(|error| {
+    TextComponent::try_from_nbt(&tag).map_err(|error| {
         reader.restore(start);
         invalid_component(reader, error.to_string())
-    })?;
-    validate_component_syntax(&component).map_err(|error| {
```
coco875, line 1232: why have remove the validate ?

coco875 (changes requested):

steel-core/src/command/execution/argument.rs
nglmercer: "parse_component was not removed and remains unchanged. Since I introduced parse_structured_component, we can add validate_component_syntax there instead. What do you think?"

coco875: hmm ok not really familiar with it, will try do a more deep review later

JunkyDeveloper:
Please fix lint

nglmercer:
> Please fix lint

ready

steel-core/src/command/execution/argument.rs
JunkyDeveloper: @coco875 did you had time for the deeper review?

coco875: not really an expert in command but can try

JunkyDeveloper: same for me

JunkyDeveloper: @hthug06 if you have time and want to help, I know you did commands, so I think you are more expert than us maybe you can find something

steel-core/src/command/builtins/message.rs
```diff
@@ -0,0 +1,323 @@
...
+fn me_command() -> CommandNodeBuilder<CommandSource, SteelCommandRuntime> {
+    literal("me").then(argument("message", SteelArgumentType::component()).executes(me))
+}
+
+fn msg_command() -> CommandNodeBuilder<CommandSource, SteelCommandRuntime> {
+    literal("msg").then(
+        argument("targets", SteelArgumentType::players())
+            .then(argument("message", SteelArgumentType::component()).executes(msg)),
+    )
+}
```
Laptop59, line 62: I'm pretty sure the wrong argument type is being used. It should be some kind of `MessageArgument` (from Vanilla), not a text component.

coco875 (dismissed):

Laptop59 (changes requested):

steel-core/src/command/builtins/message.rs
nglmercer: > I'm pretty sure the wrong argument type is being used. It should be some kind of `MessageArgument` (from Vanilla), not a text component.

ready new 'message' in SteelArgumentType

steel-core/src/command/execution/argument.rs
```diff
@@ -1163,26 +1178,68 @@ fn selected_clock(
...
+fn parse_message(reader: &mut StringReader<'_>) -> Result<TextComponent, CommandSyntaxError> {
+    let length = reader.remaining().len();
+    if length > 256 {
+        let message = translations::ARGUMENT_MESSAGE_TOO_LONG
+            .message([length.to_string(), "256".to_owned()])
+            .component();
+        return Err(reader.error(CommandSyntaxErrorKind::Dynamic(Box::new(message))));
+    }
+    Ok(TextComponent::plain(reader.read_remaining().to_owned()))
+}
```
Laptop59: Vanilla also resolves selectors to the names of the entities referenced by them, separated by commas, if the sender is allowed to use them.
The current code for parsing here does not do this and simply just reads the remaining string.

steel-core/src/server/tests.rs
```diff
@@ -1246,6 +1251,30 @@ fn assert_default_domain_data(player: &Player) {
...
+        }
+        if tick.is_multiple_of(100) {
+            sleep(Duration::from_millis(1)).await;
+        } else {
+            yield_now().await;
+        }
+    }
+    panic!("server jobs did not finish within 1_000_000 ticks");
+}
+
```
Laptop59: Why do a chunk scheduling change in a command PR? Isn't it out of scope?

Laptop59 (commented):
I actually tested this PR, and I found out that when trying to run `/say`, `/tell`, or `/me` as a player in the server, nothing seems to happen. This is because the server does not currently know how to handle the packet `SChatCommandSigned`. The packet's `ReadFrom` implementation is already there, so it just needs to be handled when caught, which is currently not done in SteelMC, as indicated by

> [Info] play packet id 8 is not known

Seems like on the server terminal, the commands work as expected, except for the selectors resolving to entity names.

coco875:
@Laptop59 thanks for the help to review the pr

steel-core/src/server/tests.rs
JunkyDeveloper: it isn't chunk scheduling only test, but also don't see much benefit

JunkyDeveloper (changes requested):

steel-core/src/command/execution/argument.rs
```diff
@@ -519,6 +526,76 @@ argument_value_wrapper!(
...
+            let mut names = source
+                .selector_display_names(&self.text[part.start..part.end])?
+                .into_iter();
+            if let Some(first) = names.next() {
+                result.children.push(first);
+                for name in names {
+                    result.children.push(separator.clone());
+                    result.children.push(name);
+                }
+            }
```
Laptop59, line 585: Maybe add utility functions for joining text components with a separator or default in `TextComponent` like Vanilla's `ComponentUtils.formatList`? Other commands in Vanilla also use `ComponentUtils.formatList` to format components into a list.

Also, the default separator should be colored gray.

Laptop59 (commented):
Looks like selector resolution in the server console works well. However, the server still does not recognize the signed `SChatCommandSignedPacket` from the client when sent. Without handling this packet, the implemented commands (`/say`, `/tell`, `/me` ) will not be able to do anything if ran by a player.

coco875:
any update ?

steel-core/src/command/builtins/message.rs
nglmercer: Fixed. `/say`, `/me`, and `/msg` now use a dedicated `SteelArgumentType::message()` backed by a new `MessageParser` that mirrors vanilla `MessageArgumentType`. The generic `component()` argument is back to exactly its `master` behaviour.

steel-core/src/command/execution/argument.rs
nglmercer: Fixed. `MessageParser` now scans the message for selectors at parse time (only when the source may use them), stores their spans, and `MessageValue::resolve` substitutes each one with the matched entities' display names at delivery time — matching vanilla, which defers resolution to `MessageArgument.Message.toComponent`. Unknown/missing selector types stay literal, real selector syntax errors propagate.

steel-core/src/server/tests.rs
nglmercer: You're right — reverted. `steel-core/src/server/tests.rs` is now identical to `master`.

steel-core/src/command/execution/argument.rs
nglmercer: Fixed by reusing what already exists: `command/execution/text.rs` already had `join_components` (vanilla's `ComponentUtils.formatList`), so it is now `pub(super)` and `MessageValue::resolve` calls it with `Resolvable::entity_separator()` — vanilla's gray `ComponentUtils.DEFAULT_SEPARATOR` — instead of a hand-rolled plain ", ". There's a test asserting the separator styling.

coco875:
because we suppect that the account are manage by a bot using heavily ai, we will close it to let a real person learn our codebase, instead of a blind contribution

PR #304, opened 2026-07-31
feat: impl AirBlock

shdctx:
added the AirBlock and the behavior


Joniii11:
Uh idk if this is even needed. The default impl should already do that.

4lve:
As the comment in the code explains this block provides no extra behavior in steel. It should rather be fixed in the tracker to be Implemented

PR #306, opened 2026-08-01
feat: mob complexity reduction

SpikeHD:
Makes two changes:

1. Reduces mob complexity by hoisting repeat boilerplate to the parent trait
2. Fixes a bug where uniform int RNG was not inclusive (which I discovered via killing many pigs and never getting 3 porkchops as a drop)

With this, writing mobs should be slightly easier and more focused on the individual implementations!

### DIFF
--- a/steel-core/src/behavior/blocks/building/powder_snow_block.rs
+++ b/steel-core/src/behavior/blocks/building/powder_snow_block.rs
@@ -4,7 +4,10 @@ use glam::DVec3;
 use steel_macros::block_behavior;
 use steel_registry::blocks::{BlockRef, block_state_ext::BlockStateExt as _, shapes::VoxelShape};
 use steel_registry::sound_event::SoundEventRef;
-use steel_registry::{vanilla_entities, vanilla_game_rules};
+use steel_registry::vanilla_entity_type_tags::EntityTypeTag;
+use steel_registry::{
+    REGISTRY, TaggedRegistryExt, vanilla_entities, vanilla_game_rules, vanilla_items,
+};
 use steel_utils::{BlockLocalAabb, BlockPos, BlockStateId};
 
 use crate::{
@@ -13,7 +16,8 @@ use crate::{
         EntityFallOnContext,
     },
     entity::ai::path::PathComputationType,
-    entity::{Entity, InsideBlockEffectCollector, InsideBlockEffectType, LivingEntity},
+    entity::{Entity, InsideBlockEffectCollector, InsideBlockEffectType},
+    inventory::equipment::EquipmentSlot,
     world::{LevelReader, World},
 };
 
@@ -40,6 +44,25 @@ impl PowderSnowBlock {
     #[must_use]
     pub const fn new(block: BlockRef) -> Self {
         Self { block }
+    }
+
+    /// Returns vanilla `PowderSnowBlock.canEntityWalkOnPowderSnow`.
+    pub(crate) fn can_entity_walk_on_powder_snow<E: Entity + ?Sized>(entity: &E) -> bool {
+        if REGISTRY.entity_types.is_in_tag(
+            entity.entity_type(),
+            &EntityTypeTag::POWDER_SNOW_WALKABLE_MOBS,
+        ) {
+            return true;
+        }
+
+        let Some(living) = entity.as_living_entity() else {
+            return false;
+        };
+        let mut has_leather_boots = false;
+        living.with_equipment_slot(EquipmentSlot::Feet, &mut |item_stack| {
+            has_leather_boots = item_stack.is(&vanilla_items::LEATHER_BOOTS);
+        });
+        has_leather_boots
     }
 
     #[must_use]
@@ -93,17 +116,13 @@ impl BlockBehavior for PowderSnowBlock {
         pos: BlockPos,
         entity: &dyn Entity,
     ) -> VoxelShape {
-        let can_walk_on_powder_snow = entity.as_living_entity().map_or_else(
-            || entity.default_can_walk_on_powder_snow(),
-            LivingEntity::can_walk_on_powder_snow,
-        );
         let collision_shape = self.get_collision_shape(
             state,
             world,
             pos,
             BlockCollisionContext::entity(entity.position().y, entity.is_descending())
                 .with_fall_distance(entity.fall_distance())
-                .with_can_walk_on_powder_snow(can_walk_on_powder_snow)
+                .with_can_walk_on_powder_snow(Self::can_entity_walk_on_powder_snow(entity))
                 .with_falling_block(entity.entity_type() == &vanilla_entities::FALLING_BLOCK),
         );
         if collision_shape.is_empty() {
--- a/steel-core/src/behavior/blocks/building/powder_snow_block.rs
+++ b/steel-core/src/behavior/blocks/building/powder_snow_block.rs
@@ -13,7 +13,7 @@ use crate::{
         EntityFallOnContext,
     },
     entity::ai::path::PathComputationType,
-    entity::{Entity, InsideBlockEffectCollector, InsideBlockEffectType},
+    entity::{Entity, InsideBlockEffectCollector, InsideBlockEffectType, LivingEntity},
     world::{LevelReader, World},
 };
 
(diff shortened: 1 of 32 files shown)


Joniii11 (approved):
i like it LGTM

steel-core/src/entity/entities/pig/mod.rs
```diff
@@ -143,50 +140,6 @@ impl PigEntity {
...
-    #[must_use]
-    pub fn is_age_locked(&self) -> bool {
-        AgeableMob::is_age_locked(self)
-    }
-
-    /// Sets the synchronized vanilla age-lock flag.
-    pub fn set_age_locked(&self, age_locked: bool) {
-        AgeableMob::set_age_locked(self, age_locked);
-    }
-
```
coco875, line 189: oh, this function was in impl Pig... yeah it can be just removed, good thing

coco875 (changes requested):
personally I think it make some problem because we go from just a dyn Entity where the compiler can optimise to inline function to an dyn Entity that try to get a dyn LivingEntity, dyn Ageable, dyn Animal, etc so we double the cost of vtable. I think there is more optimised way to do it

4lve:
> personally I think it make some problem because we go from just a dyn Entity where the compiler can optimise to inline function to an dyn Entity that try to get a dyn LivingEntity, dyn Ageable, dyn Animal, etc so we double the cost of vtable. I think there is more optimised way to do it

Yea, functions like these are usually not in hot loops though so it should be fine

4lve:
Being able to somehow get rid of all the boilerplate variant functions would be nice, maybe just like a sound_variant_registry getter function or something?

steel-core/src/entity/entity/mod.rs
```diff
@@ -1027,30 +1041,38 @@ pub trait Entity: EntityEventSource + ErasedType + Send + Sync + 'static {
...
+        self.as_living_entity()
+            .map_or_else(Vec::new, LivingEntity::pack_living_equipment)
     }
 
     /// Drains equipment slots that changed since the last tracker sync.
     fn drain_dirty_equipment(&self) -> Vec<EquipmentSlotItem> {
-        Vec::new()
+        self.as_living_entity()
+            .map_or_else(Vec::new, LivingEntity::drain_dirty_living_equipment)
     }
```
coco875: this a good example of what I mean, before we just have the entity Overwrite/replace this function so (dyn Entity).drain_dirty_equipment who be the function of the entity. But now (dyn Entity).drain_dirty_equipment call (dyn LivingEntity).drain_dirty_living_equipment then the function

coco875:
> Yea, functions like these are usually not in hot loops though so it should be fine

there is tick who are affected by that so it's in a hot loop

steel-core/src/entity/entity/mod.rs
```diff
@@ -844,11 +848,11 @@ pub trait Entity: EntityEventSource + ErasedType + Send + Sync + 'static {
...
-    /// The caller handles post-tick dirty data sync.
-    ///
-    /// Steel keeps the fallback empty because many vanilla subclasses override
-    /// tick without calling `super.tick()`.
-    fn tick(&self) {}
+    fn tick(&self) {
+        if let Some(living) = self.as_living_entity() {
+            living.tick_living_entity();
+        }
+    }
```
coco875: here

steel-core/src/entity/entity/mod.rs
```diff
@@ -866,7 +870,13 @@ pub trait Entity: EntityEventSource + ErasedType + Send + Sync + 'static {
...
     /// subclasses override tick without calling `super.tick()`.
     fn base_tick(&self) {
-        self.entity_base_tick();
+        if let Some(mob) = self.as_mob() {
+            mob.base_tick_mob();
+        } else if let Some(living) = self.as_living_entity() {
+            living.base_tick_living_entity();
+        } else {
+            self.entity_base_tick();
+        }
```
coco875: and also here

steel-core/src/entity/entity/mod.rs
4lve: Yea, not sure if there is any way around this. Maybe [try_as_dyn](https://doc.rust-lang.org/nightly/core/any/fn.try_as_dyn.html) already optimizes this case, but doubt that

coco875: I think it's better to designate it as a stack of trait applied ? so Pig = Entity + Ageable + LivingEntity + etc
if we can do that in rust then have a manual vtable where whe pass in function to applied function ?

4lve: Looked into it and seems like this is the cleanest way to do it and one vtable pointer lookup isn't that expensive. Feel free to look into it though, we just don't want unnecessary boilerplate

steel-core/src/behavior/blocks/building/powder_snow_block.rs
```diff
@@ -93,13 +93,17 @@ impl BlockBehavior for PowderSnowBlock {
...
+            LivingEntity::can_walk_on_powder_snow,
+        );
         let collision_shape = self.get_collision_shape(
             state,
             world,
             pos,
             BlockCollisionContext::entity(entity.position().y, entity.is_descending())
                 .with_fall_distance(entity.fall_distance())
-                .with_can_walk_on_powder_snow(entity.can_walk_on_powder_snow())
+                .with_can_walk_on_powder_snow(can_walk_on_powder_snow)
```
coco875: I think it's worst then have entity.can_walk_on_powder_snow for code maintanability

steel-core/src/entity/manager/mod.rs
```diff
@@ -1470,7 +1470,11 @@ impl WorldEntityManager {
     ) {
         snapshot_old_pos_and_rot_for_tick(entity.as_ref());
         entity.advance_tick_count();
-        entity.tick();
+        if let Some(living) = entity.as_living_entity() {
+            living.tick_living_entity();
+        } else {
+            entity.tick();
+        }
```
coco875: why ? it wasn't like this before ?

steel-core/src/behavior/blocks/building/powder_snow_block.rs
coco875: I think revert that are also better

steel-core/src/entity/entities/pig/mod.rs
```diff
@@ -567,7 +332,7 @@ impl LivingEntity for PigEntity {
     }
 
     fn can_use_slot(&self, slot: EquipmentSlot) -> bool {
-        slot != EquipmentSlot::Saddle || self.can_use_saddle_slot()
+        slot != EquipmentSlot::Saddle || (Entity::is_alive(self) && !AgeableMob::is_baby(self))
```
coco875, line 335: now that I think of it we can simplify it to just self.is_baby instead of precesing it's AgeableMob::is_baby ?

coco875: (by memory the problem before was the ambigous but like we only have one is_baby it should be fine now ?)

steel-core/src/entity/entity/mod.rs
```diff
@@ -855,7 +859,11 @@ pub trait Entity: EntityEventSource + ErasedType + Send + Sync + 'static {
     /// Mirrors vanilla `Entity.rideTick`.
     fn ride_tick(&self) {
         self.set_velocity(DVec3::ZERO);
-        self.tick();
+        if let Some(living) = self.as_living_entity() {
+            living.tick_living_entity();
+        } else {
+            self.tick();
+        }
```
coco875: it's necessary because it seems less efficient ?

coco875 (changes requested):

steel-core/src/behavior/blocks/building/powder_snow_block.rs
4lve: Putting that behavior in the block now instead

steel-core/src/entity/manager/mod.rs
4lve: Fixing that, moving the check into the default impl instead of at call sites. Having to override self.tick_living_entity() for each entity impl is too much boilerplate, but if we have performance concerns we could benchmark it

steel-core/src/entity/entities/pig/mod.rs
4lve: Nope, both living entity and AgeableMob implement is_baby. Also I don't mind fully qualified syntax, I think it's quite nice

4lve (approved):

coco875 (approved):
I don't have anything to add, most of my issue have been reduced or addressed

PR #307, opened 2026-08-01
Merge ProtoChunk, LevelChunk, and ChunkAccess into one Chunk struct

4lve:
Initially, we followed vanilla’s chunk design because we didn’t know what worldgen would need later. Now that worldgen is mostly finished, we can step back and simplify it.

`ProtoChunk` and `LevelChunk` stored a lot of the same data, while the current chunk status was already published by `ChunkHolder`. Because the holder had to switch between `Unloaded`, `Proto`, and `Full`, its data had to be stored behind an `RwLock`. This meant every chunk read had to acquire a read lock.

Chunks are now created once inside a `OnceLock` and promoted to Full in place. Full-only methods are still kept behind `FullChunkRef`, but the chunk itself no longer needs to be replaced. Normal reads now only need a couple of atomic checks instead of touching a shared read lock.

This also let us keep heightmaps, lighting, scheduled ticks, block entities, and staged entities in the same chunk throughout generation. Worldgen now uses typed access for the Noise, Surface, and Carvers stages, making it harder to use operations during the wrong stage. I also fixed cached worldgen reads using an old chunk status and made the vanilla generator reuse its aquifer state between stages.

I added a read-only helper for reading blocks across multiple chunks and sections. It acquires each section lock once in a fixed order, which avoids repeatedly locking the same sections. These reads are limited to small regions, with larger searches falling back to normal block reads. Writing is not supported yet because block writes can cause recursive updates and other side effects.

World ticking now keeps one worker thread per world instead of spawning a new blocking task every tick. This removes some repeated task setup while still allowing worlds to tick in parallel.

Chunk loading was also hardened. If a chunk payload is corrupt, it is removed only if it has not changed since the load started, and the chunk can then regenerate normally instead of getting stuck permanently. Errors in the region table or incomplete file reads are still kept as errors so we don’t throw away data that might still be recoverable.


4lve:
Also forgot to mention in the description but my chunk generation speeds on non-native PREGEN_SIZE=301 went from ~2400 to ~2500

PR #309, opened 2026-08-01
refactor command

coco875:
add more then function to reduce the number of tabulation.
reduce the number of function in command


JunkyDeveloper (dismissed):

steel-core/src/command/builtins/teleport.rs
```diff
@@ -214,17 +168,84 @@ fn teleport_to_entity(
...
+}
+
+fn teleport_to_position(
+    context: &SteelCommandContext<CommandSource>,
+    variant: PositionTeleport,
 ) -> Result<i32, CommandSyntaxError> {
     let source = context.source();
+    let destination = required_coordinates(context, "location")?;
+    let resolved = variant.resolve(context)?;
+    let targets = resolved.targets.resolve(context)?;
```
Purdze: **bug:** this resolves location/rotation/facing before the `targets` selector. vanilla (`TeleportCommand.java:84-103`) evaluates targets first, and base matched that, so the wrong error surfaces when multiple args are bad: `/tp @e[type=zombie] 0 64 0 facing entity @a` with no zombies + 2 players now says "Only one entity is allowed" instead of "No entity was found".

root cause is that `ResolvedPositionTeleport.targets` holds an *unresolved* selector despite the name. making it carry real resolved targets (resolved first, before the facing/rotation reads) fixes both the name and the ordering.

steel-core/src/command/brigadier/builder.rs
```diff
@@ -150,6 +150,26 @@ where
         self
     }
 
+    /// Adds a linear descendant path in declaration order.
+    #[must_use]
+    pub(crate) fn then_chain(self, descendants: impl IntoIterator<Item = Self>) -> Self {
+        let mut descendants = descendants.into_iter().collect::<Vec<_>>();
+        let Some(mut chain) = descendants.pop() else {
+            return self;
+        };
```
Purdze: two things on `then_chain`:

- an empty iterator silently returns `self`, dropping the whole branch with no error, and nothing at registration time rejects a childless non-executable node. all current call sites pass literal arrays so nothing breaks today, but a `debug_assert!(!descendants.is_empty())` would close it for future dynamic callers.
- `then_chain` and `then_all` have identical signatures but opposite tree semantics, so substituting one for the other always compiles and silently restructures the command surface. names that state the relationship (`then_path`/`then_siblings`?) would make a swap visible in review.

steel-core/src/command/brigadier/builder.rs
```diff
@@ -150,6 +150,26 @@ where
...
+        while let Some(parent) = descendants.pop() {
+            chain = parent.then(chain);
+        }
+        self.then(chain)
+    }
+
+    /// Adds sibling children while preserving registration order.
+    #[must_use]
+    pub(crate) fn then_all(mut self, children: impl IntoIterator<Item = Self>) -> Self {
+        self.children.extend(children);
```
Purdze: `then_all` extends `self.children` directly instead of routing through `then`, which forks the single child-insertion path. any future invariant added to `then` (duplicate-name assert, redirect check, modding hook) gets silently skipped by every `then_all` call site. `children.into_iter().fold(self, Self::then)` keeps one path.

Purdze (changes requested):
**duplication:** the `Source`/`Argument` + `ResolvedX` + `resolve()` scaffold is copy-pasted in 5 files (`clear.rs:37`, `fly.rs:56`, `gamemode.rs:57`, `kill.rs:32`, `teleport.rs:72`, with clear/fly identical apart from the name). that's ~150 lines that one shared helper on `SteelCommandContext` (`execution/runtime.rs:373`) would replace

**public api:** the public `CommandNode` (`api.rs:154`) didn't get the new builder methods, so extension authors are left on exactly the chained `.then()` shape this pr set out to fix. forwarding both is a few mechanical lines

steel-core/src/command/builtins/teleport.rs
coco875: I think I have fix it correctly not sure

JunkyDeveloper:

JunkyDeveloper:
will this refactor still important with the change from yellow?

PR #310, opened 2026-08-02
force lf

coco875:


JunkyDeveloper (dismissed):
LGTM

Joniii11 (approved):
lgtm

chocodev11:
i love windows

Purdze (changes requested):
i tested this and the UTF-8 half doesn't actually do anything as git deliberately skips conversion/validation when working-tree-encoding equals its internal default

the utf8 macro and the pattern lines reduce to text eol=lf, which line 1 already applies globally so this file could just be * text=auto eol=lf plus *.png binary

verified with git 2.51.1

coco875:
like that ?

Purdze:
> like that ?

lgtm after one fix: add *.webp binary (sunset.webp currently falls through to text=auto eol=lf). also maybe retitle since there's no utf-8 enforcement left, it's lf normalization only. existing clones will need git rm -r --cached . && git checkout . to refresh stale crlf files

other than that all good

coco875:
push, nothing else ?

Purdze (approved):

chocodev11:
yay, merge plz

JunkyDeveloper (approved):

PR #311, opened 2026-08-03
Feat: redstone signal optimization

chocodev11:
This PR purpose is porting 2 optimization concept i saw in lithium for redstone and it should be preserve vanilla behaviour, may test with flint idk.
- It read vertical states once, then visit each horizontal neighbor once and calculate its non-wire input and possible wire input together.
- It also reuses the above-block conductor result, dispatches no behavior for air/wire direct-signal candidates, returns at strength 15, and stops checking wires once non-wire input is already 14 because attenuated wire cannot exceed 14.


steel-core/src/behavior/blocks/redstone/wire/evaluator.rs
```diff
@@ -49,37 +48,158 @@ impl DefaultRedstoneWireEvaluator {
     }
 
     fn calculate_target_strength(&self, level: &dyn LevelReader, pos: BlockPos) -> i32 {
```
coco875, line 50: hmm the function look very long, I check on lithium are function are more well organize

coco875:
I have add comment and rename function to help make the parallel with lithium and vanill and it look good, can fix merge conflict ?

coco875 (approved):

JunkyDeveloper (approved):
gave my best to understand it and looks good for me, because will not wait now more weeks until flint is done...

PR #313, opened 2026-08-05
Add a headless pregeneration benchmark

fionera:
Headless pregen benchmark so we can measure the whole pipeline (tickets, scheduling, generation, unload, saves) without running the server binary and fighting over port 25565 and `saves/`.

Stacked on #314, so it shows both commits for now - only the second one is this PR.

### DIFF
--- a/steel-core/benches/pregen.rs
+++ b/steel-core/benches/pregen.rs
@@ -1,0 +1,381 @@
+//! End-to-end pregeneration throughput, without a server.
+//!
+//! The other benchmarks in this crate drive `GENERATION_PYRAMID` steps against
+//! hand-built holders, which measures generation compute. This one runs the
+//! production pregeneration driver against a bare `World`, so it covers the
+//! whole pipeline: tickets, scheduling epochs, the generation pool, unload and
+//! region saves.
+//!
+//! ```text
+//! cargo bench -p steel-core --bench pregen --features benchmark-support -- --size 301 --reps 3
+//! ```
+use std::env;
+use std::fmt::Display;
+use std::fs;
+use std::num::NonZero;
+use std::path::PathBuf;
+use std::process;
+use std::str::FromStr;
+use std::sync::Arc;
+use std::thread;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+use steel_core::bootstrap::init_globals_once;
+use steel_core::config::WorldStorageConfig;
+use steel_core::level_data::WorldGenerationSettings;
+use steel_core::server::pregen::pregen_area_for_benchmark;
+use steel_core::world::{World, WorldConfig};
+use steel_core::worldgen::WorldGeneratorRegistry;
+use steel_registry::vanilla_dimension_types;
+use steel_utils::threading::worker_threads_for_available;
+use steel_utils::types::{Difficulty, GameType};
+use steel_utils::{ChunkPos, Identifier};
+use tokio::runtime::{Builder, Runtime};
+use tokio_util::sync::CancellationToken;
+use toml::Value;
+use toml::map::Map;
+
+const DEFAULT_SEED: i64 = -9_091_483_014_810_473_238;
+const DEFAULT_SIZE: i32 = 301;
+const DEFAULT_REPS: usize = 3;
+
+#[global_allocator]
+static ALLOC: mimalloc::MiMalloc = mimalloc::MiMalloc;
+
+struct Options {
+    size: i32,
+    reps: usize,
+    seed: i64,
+    generation_threads: usize,
+    chunk_workers: usize,
+    main_workers: usize,
+    encoding_threads: usize,
+    window_size: Option<i32>,
+    ram_only: bool,
+    keep_storage: bool,
+}
+
+impl Options {
+    fn parse() -> Result<Self, String> {
+        let available = thread::available_parallelism().map_or(4, NonZero::get);
+        let mut options = Self {
+            size: DEFAULT_SIZE,
+            reps: DEFAULT_REPS,
+            seed: DEFAULT_SEED,
+            generation_threads: available,
+            chunk_workers: worker_threads_for_available(None, available),
+            main_workers: worker_threads_for_available(None, available),
+            encoding_threads: available,
+            window_size: None,
+            ram_only: false,
+            keep_storage: false,
+        };
+
+        // `cargo bench` passes libtest's own flags through.
+        let mut args = env::args().skip(1).peekable();
+        while let Some(arg) = args.next() {
+            let mut value = || args.next().ok_or_else(|| format!("{arg} requires a value"));
+            match arg.as_str() {
+                "--size" => options.size = parse(&value()?, "--size")?,
+                "--reps" => options.reps = parse(&value()?, "--reps")?,
+                "--seed" => options.seed = parse(&value()?, "--seed")?,
+                "--gen-threads" => options.generation_threads = parse(&value()?, "--gen-threads")?,
+                "--chunk-workers" => options.chunk_workers = parse(&value()?, "--chunk-workers")?,
+                "--main-workers" => options.main_workers = parse(&value()?, "--main-workers")?,
+                "--encode-threads" => {
+                    options.encoding_threads = parse(&value()?, "--encode-threads")?;
+                }
+                "--window" => options.window_size = Some(parse(&value()?, "--window")?),
+                "--ram-only" => options.ram_only = true,
+                "--keep-storage" => options.keep_storage = true,
+                "--bench" | "--test" => {}
+                "--help" | "-h" => {
+                    print_usage();
+                    process::exit(0);
+                }
+                other => return Err(format!("unknown argument {other} (try --help)")),
+            }
+        }
+
+        if options.reps == 0 {
+            return Err("--reps must be at least 1".to_owned());
+        }
+        Ok(options)
+    }
+}
+
+fn parse<T: FromStr>(value: &str, flag: &str) -> Result<T, String>
+where
+    T::Err: Display,
+{
+    value
+        .parse()
+        .map_err(|error| format!("{flag} takes a number: {error}"))
+}
+
+fn print_usage() {
+    println!(
+        "\
+Headless pregeneration benchmark.
+
+  --size N           chunk side length, odd (default {DEFAULT_SIZE})
+  --reps N           runs to perform (default {DEFAULT_REPS})
+  --seed N           world seed (default {DEFAULT_SEED})
+  --gen-threads N    generation pool threads (default: the server's)
+  --chunk-workers N  chunk runtime workers (default: the server's)
+  --main-workers N   main runtime workers (default: the server's)
+  --encode-threads N encoding pool threads (default: the server's)
+  --window N         pregen window side length (default: the server's)
+  --ram-only         skip region-file persistence entirely
+  --keep-storage     do not delete the per-run storage directory
+"
+    );
+}
+
+/// Current and peak resident set size in MiB.
+///
+/// Peak is process-lifetime and does not reset between reps, so use `--reps 1`
+/// when comparing it. Linux only; elsewhere the rate is printed on its own.
+#[cfg(target_os = "linux")]
+fn rss_mib() -> Option<(u64, u64)> {
+    let status = fs::read_to_string("/proc/self/status").ok()?;
+    let field = |name: &str| {
+        let line = status.lines().find(|line| line.starts_with(name))?;
+        let kib: u64 = line.split_whitespace().nth(1)?.parse().ok()?;
+        Some(kib / 1024)
+    };
+    Some((field("VmRSS:")?, field("VmHWM:")?))
+}
+
+#[cfg(not(target_os = "linux"))]
+fn rss_mib() -> Option<(u64, u64)> {
+    None
+}
+
+/// Unique per run and removed afterwards. Under `target/` because a 601x601 run
+/// writes gigabytes of region files.
+fn storage_root(rep: usize) -> PathBuf {
+    let unique = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map_or(0, |elapsed| elapsed.as_nanos());
+    PathBuf::from(env!("CARGO_MANIFEST_DIR"))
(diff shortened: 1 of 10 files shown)


colonthreeing:
You should stop having AI write your pull requests so it looks at least a little bit like you did something in the process

colonthreeing:
Also do note the Steel README

> AI may be used as a tool, but contributors must understand and be able to explain every line they submit. **Fully autonomous pull requests are not accepted.**

(emphasis added)

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,407 @@
...
+            match arg.as_str() {
+                "--size" => options.size = parse(&value()?, "--size")?,
+                "--reps" => options.reps = parse(&value()?, "--reps")?,
+                "--seed" => options.seed = parse(&value()?, "--seed")?,
+                "--gen-threads" => options.generation_threads = parse(&value()?, "--gen-threads")?,
+                "--chunk-workers" => options.chunk_workers = parse(&value()?, "--chunk-workers")?,
+                "--main-workers" => options.main_workers = parse(&value()?, "--main-workers")?,
+                "--encode-threads" => {
+                    options.encoding_threads = parse(&value()?, "--encode-threads")?;
+                }
```
coco875, line 88: for generation you don't need this thread

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,407 @@
...
+                "--window" => options.window_size = Some(parse(&value()?, "--window")?),
+                "--ram-only" => options.ram_only = true,
+                "--keep-storage" => options.keep_storage = true,
```
coco875, line 91: just simplify by doing only ram storage, I don't see any case where we would need it

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,407 @@
...
+}
+
+fn ensure_globals() {
+    INIT.call_once(|| {
+        let mut registry = Registry::new_vanilla();
+        registry.freeze();
+        let _ = REGISTRY.init(registry);
+        init_behaviors();
+        init_block_entities();
+        init_entities();
```
coco875: normally we can generate without this

steel-core/benches/pregen.rs
coco875, line 1: it would be better to extend bench_overworld_full_chunk_concurrent then create a whole new test that do the same

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,407 @@
...
+/// A storage directory unique to this run, removed when the run ends. Under
+/// `target/` because a 601x601 run writes gigabytes of region files.
+fn storage_root(rep: usize) -> PathBuf {
+    let unique = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map_or(0, |elapsed| elapsed.as_nanos());
+    PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("../target/bench-pregen")
+        .join(format!("{}-{rep}-{unique}", process::id()))
+}
```
coco875: don't store if it's serverless

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,407 @@
...
+            .build()
+            .map_err(|error| format!("generation pool should start: {error}"))?,
+    );
+    let encoding_pool = Arc::new(
+        rayon::ThreadPoolBuilder::new()
+            .num_threads(options.encoding_threads)
+            .thread_name(|index| format!("rayon-chunk-enc-{index}"))
+            .build()
+            .map_err(|error| format!("encoding pool should start: {error}"))?,
+    );
```
coco875: why you need encoding ?

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,407 @@
...
+            .enable_all()
+            .build()
+            .map_err(|error| format!("chunk runtime should start: {error}"))?,
+    );
+    let main_runtime = Builder::new_multi_thread()
+        .worker_threads(options.main_workers)
+        .thread_name("main-worker")
+        .enable_all()
+        .build()
+        .map_err(|error| format!("main runtime should start: {error}"))?;
```
coco875, line 189: shouldn't be necessary

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,407 @@
...
+  --window N         pregen window side length (default: the server's)
+  --ram-only         skip region-file persistence entirely
+  --keep-storage     do not delete the per-run storage directory
+"
+    );
+}
+
+/// Current and peak resident set size in MiB.
+///
+/// A scheduler change's failure mode is often memory rather than speed. Peak is
```
coco875: can you poiny me where in https://docs.kernel.org/scheduler/index.html this information come from ? (a bit skeptical)

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,407 @@
...
+
+impl Options {
+    fn parse() -> Result<Self, String> {
+        let available = thread::available_parallelism().map_or(4, NonZero::get);
+        let mut options = Self {
+            size: DEFAULT_SIZE,
+            reps: DEFAULT_REPS,
+            seed: DEFAULT_SEED,
+            // The server's defaults: rayon's own width for both pools, which is
+            // what the server uses when the config leaves them unset.
```
coco875: useless comment

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,407 @@
...
+use steel_utils::threading::worker_threads_for_available;
+use steel_utils::types::{Difficulty, GameType};
+use steel_utils::{ChunkPos, Identifier};
+use tokio::runtime::{Builder, Runtime};
+use tokio_util::sync::CancellationToken;
+use toml::Value;
+use toml::map::Map;
+
+/// Pinned: terrain shape drives how much work a chunk is, so two seeds are two
+/// different workloads.
```
coco875: useless comment

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,407 @@
+//! End-to-end pregeneration throughput, without a server.
+//!
+//! The other benchmarks in this crate drive `GENERATION_PYRAMID` steps against
+//! hand-built holders, which measures generation compute. This one runs the
+//! production pregeneration driver against a bare `World`, so it covers the
+//! whole pipeline: tickets, scheduling epochs, the generation pool, unload and
+//! region saves.
```
coco875, line 7: you want to include unload and region save ??? not just generation

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,407 @@
...
+//!
+//! ```text
+//! cargo bench -p steel-core --bench pregen --features benchmark-support -- --size 301 --reps 3
+//! ```
+//!
+//! Not a criterion benchmark: a run is seconds of wall clock and warms up
+//! internally as the pipeline fills, so its own reps say more than resampling
+//! would.
```
coco875: can be remove

coco875 (changes requested):
hmm honestly the code don't feel well integrated to other benchmark or even also use steel correctly

steel-core/benches/pregen.rs
fionera: Yes, I bascially want the full pregen benchmark we currently do just as cargo benchmark to have the full e2e codepath that would run in reality

coco875: by memory in samply that I have done this two are very small so close to no relevant and would more contribute to add noise

coco875: also would prefer if we separate the compute from the io

coco875: also if we want the full full pipeline we can just use PREGEN_SIZE and run it to also have network etc that are left unused until it boot

steel-core/benches/pregen.rs
fionera: No I want to also have it exercise the storage path as that can produce backpressure

steel-core/benches/pregen.rs
fionera: Its not about the kernel scheduler but in the steel scheduling, which when done wrong will fill the memory with to-be-unloaded chunks

steel-core/benches/pregen.rs
fionera: Doesnt seem like it https://github.com/Steel-Foundation/SteelMC/blob/master/steel-core/benches/worldgen.rs#L93

coco875: hmm ok didn't see it correctly

steel-core/benches/pregen.rs
fionera: described above

coco875:
@fionera you can just ignore comment and do like I mention so with a refactor to re-use part of initialize the server, it would be better

coco875:
I made comment when I didn't fully understand the purpose of it

steel-core/benches/pregen.rs
coco875: hmm I didn't read steel scheduling yet

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,381 @@
...
+            .build()
+            .map_err(|error| format!("generation pool should start: {error}"))?,
+    );
+    let encoding_pool = Arc::new(
+        rayon::ThreadPoolBuilder::new()
+            .num_threads(options.encoding_threads)
+            .thread_name(|index| format!("rayon-chunk-enc-{index}"))
+            .build()
+            .map_err(|error| format!("encoding pool should start: {error}"))?,
+    );
```
coco875: I think pool initialisation can be share code with server startup

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,381 @@
...
+        match storage {
+            Some(storage) if keep_storage => {
+                println!("  storage kept at {}", storage.display());
+            }
+            Some(storage) => {
+                let _ = fs::remove_dir_all(storage);
+            }
+            None => {}
+        }
+    }
```
coco875, line 294: same

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,381 @@
...
+                compression: None,
+                is_flat,
+                sea_level,
+                default_gamemode: GameType::Survival,
+                difficulty: Difficulty::Normal,
+            },
+            generation_pool,
+            encoding_pool,
+        ))
+        .map_err(|error| format!("bench world should initialize: {error}"))?;
```
coco875, line 248: maybe also it ?

coco875 (changes requested):

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,363 @@
...
+    chunk_workers: usize,
+    main_workers: usize,
+    encoding_threads: usize,
+    window_size: Option<i32>,
+    ram_only: bool,
+    keep_storage: bool,
+}
+
+impl Options {
+    fn parse() -> Result<Self, String> {
```
coco875, line 60: We don't have a lib for cli argument ?

steel-core/benches/pregen.rs
```diff
@@ -0,0 +1,363 @@
...
+"
+    );
+}
+
+/// Current and peak resident set size in MiB.
+///
+/// Peak is process-lifetime and does not reset between reps, so use `--reps 1`
+/// when comparing it. Linux only; elsewhere the rate is printed on its own.
+#[cfg(target_os = "linux")]
+fn rss_mib() -> Option<(u64, u64)> {
```
coco875, line 141: Maybe it can also work on macos, like it's unix

steel-core/src/server/mod.rs
```diff
@@ -3,7 +3,14 @@ mod broadcasting;
...
 pub mod jobs;
 mod packet_processor;
+#[cfg(not(feature = "benchmark-support"))]
+mod pools;
+#[cfg(feature = "benchmark-support")]
+pub mod pools;
+#[cfg(not(feature = "benchmark-support"))]
 mod pregen;
+#[cfg(feature = "benchmark-support")]
+pub mod pregen;
```
coco875, line 13: ? weird to include it if a flag are enable or not, you can just put it in always pub ?

steel-core/src/server/mod.rs
```diff
@@ -3,7 +3,14 @@ mod broadcasting;
 /// Tick-polled server jobs.
 pub mod jobs;
 mod packet_processor;
+#[cfg(not(feature = "benchmark-support"))]
+mod pools;
+#[cfg(feature = "benchmark-support")]
+pub mod pools;
```
coco875, line 9: same

steel-core/src/server/pregen.rs
```diff
@@ -215,6 +215,42 @@ impl Server {
...
+/// over. Everything below it is the production path, unchanged.
+///
+/// `window_size` defaults to the server's value when `None`. Returns the wall
+/// time of the pass, or `None` if it was cancelled.
+///
+/// # Errors
+/// Returns an error if `side_length` is not a positive odd integer, or if the
+/// window size exceeds the unload-backpressure budget.
+#[cfg(feature = "benchmark-support")]
+pub async fn pregen_area_for_benchmark(
```
coco875, line 234: not sure about the place of this function

steel-core/src/world/mod.rs
```diff
@@ -333,6 +333,33 @@ impl World {
...
+            chunk_runtime,
+            key,
+            dimension_type,
+            seed,
+            config,
+            generation_pool,
+            chunk_encoding_pool,
+        )
+        .await
+    }
```
coco875, line 360: why not just call new_with_config_and_encoding_pool_impl instead of wrap it ?

steel-core/src/world/mod.rs
```diff
@@ -341,6 +368,27 @@ impl World {
         config: WorldConfig,
         generation_pool: Arc<rayon::ThreadPool>,
         chunk_encoding_pool: Arc<rayon::ThreadPool>,
+    ) -> io::Result<Arc<Self>> {
+        Self::new_with_config_and_encoding_pool_impl(
```
coco875, line 372: you do the same here ??? what is the difference ?

coco875 (changes requested):

PR #316, opened 2026-08-06
feat: campfire cooking

SpikeHD:
Adds campfire cooking

# Remarks

* Renamed `SmeltingRecipe` (and similar) to `CookingRecipe` because it is more generic (and it is now shared)
  * I will also note the smelting/campfire codegen helper. I added that because they share the same recipe shape and it's otherwise a bit repetitious, but I can revert if desired.
* The TODO mentioned ejecting items on douse but that's not actually vanilla behavior
* Also tested on soul campfires!

https://github.com/user-attachments/assets/0a17af53-d839-4c1b-b739-1da09298da2c

### DIFF
--- a/steel-core/src/behavior/blocks/building/campfire_block.rs
+++ b/steel-core/src/behavior/blocks/building/campfire_block.rs
@@ -39,6 +39,16 @@ pub struct CampfireBlock {
     fire_damage: i32,
 }
 
+/// Outcome of attempting to place a held item onto a campfire.
+enum CampfirePlaceOutcome {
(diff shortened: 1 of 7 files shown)


coco875:
> yep but its blocked by 4lve who said he wanted to implement all recipes AND recipe property sets
I will defer the review until I can see how recipe are done to try doing the improvement (or that somone else do it)

JunkyDeveloper:
#160 to have it linked

CuzImClicks:
#565

PR #318, opened 2026-08-07
allow flint to process chunks

JunkyDeveloper:
For flint, we need to load the chunks synchronous the previous version is not anymore possible.
First I created a function which exposes via a feature the adance_scheduling (see comment).
But now the method was changed from `pub (crate)` to `pub`.

### DIFF
--- a/steel-core/src/chunk/chunk_map/mod.rs
+++ b/steel-core/src/chunk/chunk_map/mod.rs
@@ -468,16 +468,6 @@ impl ChunkMap {
         Some(holder)
     }
 
-    /// Drives one scheduling step (ticket propagation, holder creation, and
-    /// generation-epoch commit).
-    ///
-    /// Exposed for Flint's synchronous chunk-loading test harness, which has
-    /// no server run loop to drive scheduling on its behalf.
-    #[cfg(feature = "flint")]
-    pub fn drive_scheduling_for_flint(self: &Arc<Self>) {
-        let _ = self.advance_scheduling();
-    }
-
     /// Inserts a non-simulated holder into an empty gameplay view for worldgen benchmarks.
     ///
     /// Runtime lifecycle code must use ticket-driven insertion. Benchmark holders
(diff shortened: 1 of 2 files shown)


Joniii11 (approved):
lgtm xD

steel-core/src/chunk/chunk_map/mod.rs
```diff
@@ -468,6 +468,16 @@ impl ChunkMap {
...
 
+    /// Drives one scheduling step (ticket propagation, holder creation, and
+    /// generation-epoch commit).
+    ///
+    /// Exposed for Flint's synchronous chunk-loading test harness, which has
+    /// no server run loop to drive scheduling on its behalf.
+    #[cfg(feature = "flint")]
+    pub fn drive_scheduling_for_flint(self: &Arc<Self>) {
+        let _ = self.advance_scheduling();
+    }
```
coco875: if it's an Arc<Self> why not just make advance_scheduling pub instead ? to avoid the wrapper

JunkyDeveloper: I also thought about it, but I decided against that to make the "impact" on steel for flint as minimal as possible. So I did now your way, because I'm not the only person with that opinion

coco875 (dismissed):

coco875:
now fix lint x) I guess put also ChunkMapSchedulingTimings in pub, if there is lib that use steel as lib I can see the benefit of access it

coco875 (dismissed):

Purdze (changes requested):
the timings struct is still unusable from flint. chunk_scheduler is a private module and the re export at chunk_map/mod.rs:35 is pub(crate) use, so the type can't be named outside the crate and its fields are all still pub(crate) anyway. needs pub use on the re-export and pub fields, like we do for GameplayChunkLookupCacheStats / ChunkMapGameTickTimings

also advance_scheduling needs its calling constraint documented now that it's public as it must only run on the thread driving the world, between ticks; calling it mid tick from another thread commits an epoch while tick_game iterates the old chunk snapshot also its first doc line is copy pasted from tick_game and wrong

coco875 (approved):

Purdze (approved):

PR #319, opened 2026-08-07
Feat: tinted glass and stained glass

chocodev11:


Joniii11:
why is this even needed tho? the only impls i can see is the block state for placement and its just the default one.

chocodev11:
> why is this even needed tho? the only impls i can see is the block state for placement and its just the default one.

oh wait, i added light logic to tinted glass but i realize it was client-side and remove it

PR #322, opened 2026-08-08
Fix server crash from invalid use-item rotations

kdcokenny:
I was looking through the use-item packet handling and noticed that NaN or infinite rotation values could reach `set_rotation` and panic the server.

I reproduced it by sending a use-item packet with a NaN rotation while holding an item.

Fixed it by matching vanilla behavior and ignoring each invalid rotation component while still applying valid ones. Also added tests for NaN and infinities.

Part of #323

### DIFF
--- a/steel-core/src/player/game_mode/item_interaction.rs
+++ b/steel-core/src/player/game_mode/item_interaction.rs
@@ -193,18 +193,19 @@ impl Player {
 
         let current_rotation = self.rotation();
         // Vanilla entity setters discard each non-finite rotation component independently.
-        let target_yaw = if packet.y_rot.is_finite() {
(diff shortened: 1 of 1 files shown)


JunkyDeveloper:
I think this checks are not really in place for many other packages

kdcokenny:
> I think this checks are not really in place for many other packages

Want me to check the other packet handlers too?

JunkyDeveloper:
yes would be nice!

coco875 (approved):

kdcokenny:
@JunkyDeveloper mind giving this a quick review?

JunkyDeveloper (approved):

Issue #323, opened 2026-08-08
Non-finite item interaction values can crash the server


kdcokenny:
I was checking the item interaction packet handlers and found two places where client-controlled non-finite floats can reach entity assertions and shut down the server.

- [x] `SUseItem` rotation values — addressed in #322
- [x] `SUseItemOn` click coordinates — addressed in #324

Splitting these into two PRs so each fix stays focused and has its own regression test.

PR #324, opened 2026-08-08
Fix server crash from invalid use-item-on locations

kdcokenny:
I was checking the other packet handlers after #322 and found that NaN click coordinates in `SUseItemOn` could bypass the distance check.

I reproduced it by using a firework with a NaN hit location, which reached the entity position assertion and panicked the server.

Fixed it by matching vanilla's positive bounds check so invalid locations are rejected. Also added tests for NaN and infinities.

Part of #323

### DIFF
--- a/steel-core/src/player/game_mode/block_interaction.rs
+++ b/steel-core/src/player/game_mode/block_interaction.rs
@@ -48,11 +48,11 @@ impl Player {
         let center_z = f64::from(pos.z()) + 0.5;
         let location = &packet.block_hit.location;
         let limit = 1.000_000_1;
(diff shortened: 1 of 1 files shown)


kdcokenny:
Not sure why `world::tests::set_block_matches_vanilla_update_limit_and_client_publication_gates` failed here. It seems unrelated to this change and passes locally, so it may be flaky. Could someone rerun the failed job?

JunkyDeveloper:
Please create quickly a issue for that :)

kdcokenny:
> Please create quickly a issue for that :)

done #325

JunkyDeveloper (approved):

coco875 (approved):

Issue #328, opened 2026-08-08
Chat session updates bypass validation and can exhaust memory


kdcokenny:
I was checking chat session updates and found two issues:

- [Client-controlled key and signature lengths are cast to `usize` and allocated immediately](https://github.com/Steel-Foundation/SteelMC/blob/f53b66ca8f7c4c17c4123270c5729e2a993ec3a2/steel-protocol/src/packets/game/chat/session.rs#L94-L100). Vanilla limits [public keys to 512 bytes](https://github.com/Renekovski/26.2-mcp/blob/51f3128ba7265299dce20b9abd2b37d7dd27d096/src/net/minecraft/network/FriendlyByteBuf.java#L612-L615) and [signatures to 4096 bytes](https://github.com/Renekovski/26.2-mcp/blob/51f3128ba7265299dce20b9abd2b37d7dd27d096/src/net/minecraft/world/entity/player/ProfilePublicKey.java#L48-L50).
- [Profile keys are accepted using `NoValidation`](https://github.com/Steel-Foundation/SteelMC/blob/f53b66ca8f7c4c17c4123270c5729e2a993ec3a2/steel-core/src/player/chat/mod.rs#L391-L402), which [always succeeds](https://github.com/Steel-Foundation/SteelMC/blob/f53b66ca8f7c4c17c4123270c5729e2a993ec3a2/steel-crypto/src/signature.rs#L158-L170).

- [x] Bound both fields before allocation and add malformed-packet tests.
- [ ] Cache Mojang service keys and validate profile-key signatures without an insecure fallback.

[...]

PR #329, opened 2026-08-08
Bound chat session fields before allocation

kdcokenny:
I was checking chat session update decoding and noticed packet-controlled key and signature lengths were cast from `i32` to `usize` and used for allocation before validation.

A negative or oversized VarInt could therefore request an enormous allocation from a tiny packet.

Fixed it by matching vanilla's 512-byte public-key and 4096-byte signature limits and using bounded prefixed reads before allocation. Also added tests for boundary, malformed, and truncated packets.

Part of #328

### DIFF
--- a/steel-protocol/src/packets/game/chat/session.rs
+++ b/steel-protocol/src/packets/game/chat/session.rs
@@ -1,11 +1,15 @@
 use std::{
-    io::{Cursor, Read},
+    io::Cursor,
     time::{SystemTime, UNIX_EPOCH},
 };
 
 use steel_macros::ServerPacket;
 use steel_utils::codec::VarInt;
+use steel_utils::serial::PrefixedRead;
 use uuid::Uuid;
+
+const MAX_PUBLIC_KEY_SIZE: usize = 512;
+const MAX_KEY_SIGNATURE_SIZE: usize = 4096;
 
 /// Network-serializable chat session data.
 ///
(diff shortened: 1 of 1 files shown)


JunkyDeveloper (approved):

steel-protocol/src/packets/game/chat/session.rs
```diff
@@ -91,13 +95,9 @@ impl steel_utils::serial::ReadFrom for SChatSessionUpdate {
...
-        let key_len = VarInt::read(reader)?.0 as usize;
-        let mut public_key = vec![0u8; key_len];
-        reader.read_exact(&mut public_key)?;
-
-        let sig_len = VarInt::read(reader)?.0 as usize;
-        let mut key_signature = vec![0u8; sig_len];
-        reader.read_exact(&mut key_signature)?;
+        let public_key = Vec::<u8>::read_prefixed_bound::<VarInt>(reader, MAX_PUBLIC_KEY_SIZE)?;
+        let key_signature =
+            Vec::<u8>::read_prefixed_bound::<VarInt>(reader, MAX_KEY_SIGNATURE_SIZE)?;
```
coco875, line 100: a random though now, we can just make fixed size array instead of vec ? to allocate on the stack instead of the heap ?

kdcokenny: These are variable-length fields; 512 and 4096 are only their maximums. Fixed arrays would need separate lengths and would inflate every ScheduledPlayPacketKind value by roughly 4ish KiB. The bounded Vec match the Vanilla impl and only allocates the actual wire length.

coco875: I though it can like C https://godbolt.org/z/evbbP51Er, didn't know that it wasn't possible in rust

kdcokenny: Ahh, you mean a C variable-length array. It could work during decoding, but it dies when the function returns, and this packet is queued for later, so we’d still need to copy it into owned storage. Due to this, i'd think a bounded Vec fits better here.

coco875 (approved):

PR #330, opened 2026-08-09
Implement DropExperienceBlock behavior

kdcokenny:
I was checking the extracted block behavior coverage and noticed Steel did not implement vanilla's `DropExperienceBlock`, so ores such as lapis and diamond never dropped mining experience.

Implemented it using SteelExtractor's constant and uniform `xpRange` data and matched vanilla's post-break sampling and enchantment processing. Redstone ore now shares the same helper, and unsupported extracted provider shapes fail code generation instead of silently receiving the wrong distribution.

Added tests for constant and ranged providers, Silk Touch, and disabled experience drops. I also validated ordinary diamond ore, Silk Touch suppression, and zero-XP iron ore through an automated Minecraft Java 26.2 client and a separate manual client run.


steel-core/src/behavior/block/mod.rs
```diff
@@ -338,6 +341,29 @@ pub trait BlockBehavior: Send + Sync {
     ) {
     }
 
+    /// Samples and applies enchantment effects to a block experience drop.
+    ///
+    /// Mirrors vanilla `Block.tryDropExperience`. Mining experience is
+    /// incidental live-gameplay randomness, so Steel samples it from an
+    /// unseeded runtime source.
+    fn try_drop_experience(
```
coco875: the function that you add only look like to be only use in spawn_after_break and not by external directly woudln't be better to just put it as an helper function ?

kdcokenny: yes, i agree. i was just matching the vanilla impl. Will move it out!

steel-core/src/behavior/blocks/redstone/redstone_ore_block.rs
```diff
@@ -236,21 +234,4 @@ mod tests {
                 .any(|entity| entity.entity_type() == &vanilla_entities::EXPERIENCE_ORB)
         );
     }
-
-    #[test]
-    fn silk_touch_suppresses_redstone_ore_experience() {
```
coco875, line 241: why remove the test ?

steel-core/src/behavior/blocks/building/drop_experience_block.rs
coco875, line 1: hmm @JunkyDeveloper do you have an idea of what category it would be ? it's use for SculkBlock but also for all block ore

coco875 (changes requested):

steel-core/src/behavior/blocks/redstone/redstone_ore_block.rs
kdcokenny: i didnt actually remove it, just moved it over to a shared cov. The new Diamond Ore test verifies Silk Touch suppression through the try_drop_experience i added, while the remaining redstone test verifies redstone's post-break XP dispatch.

steel-core/build/common.rs
```diff
@@ -204,6 +211,40 @@ pub(crate) fn json_value_to_tokens(
...
+    extra: &serde_json::Map<String, serde_json::Value>,
+    entry_name: &str,
+    key: &str,
+) -> TokenStream {
+    let value_key = format!("{key}_value");
+    let min_key = format!("{key}_min_inclusive");
+    let max_key = format!("{key}_max_inclusive");
+    let value = extra.get(&value_key);
+    let min = extra.get(&min_key);
+    let max = extra.get(&max_key);
```
JunkyDeveloper: Sorry but my problem here is, you extend one key to three json data fields, normally one key is one data field, which the docs make clear, so that is a changed behaviour without any documentation.

So two options from my side:
1. each key is an parameter in the new function at the block impl
2. Add this behaviour to the docs: https://steelmc.dev/development/block_item_registration/

maybe other maintainer see it different and give some good arguments, why this is ok this way, but I'm ok with it with a full added docs PRs which explains this behaviour

JunkyDeveloper (changes requested):

steel-core/src/behavior/blocks/building/drop_experience_block.rs
JunkyDeveloper: I think we should reorganize most blocks in general, because 2 folders have 90% of all blocks currently

So I would put it in building currently and move it with mud, sulfur, etc to somthing like natural

steel-core/build/classes.json
```diff
@@ -200,40 +201,49 @@
...
     {
       "name": "coal_ore",
       "class": "DropExperienceBlock",
-      "xp_range_min_inclusive": 0,
-      "xp_range_max_inclusive": 2
+      "xp_range": {
+        "min_inclusive": 0,
+        "max_inclusive": 2,
+        "type": "minecraft:uniform"
+      }
```
JunkyDeveloper, line 228: also here please make a PR at the extractor

JunkyDeveloper (changes requested):

steel-core/build/classes.json
kdcokenny: on it

steel-core/build/common.rs
```diff
@@ -23,6 +23,11 @@ pub(crate) fn path_ends_with(path: &syn::Path, name: &str) -> bool {
 pub(crate) enum JsonArgKind {
     /// Raw JSON value → token literal (handles numbers, strings, bools)
     Value,
+    /// Vanilla codec JSON → a constant or uniform `IntProvider`.
+    ///
+    /// Unsupported provider shapes fail code generation so new vanilla data
+    /// cannot silently receive the wrong distribution.
+    IntProvider,
```
JunkyDeveloper, line 30: Still sorry to say that, but as the main person who is behind docs, please add it XD

JunkyDeveloper (changes requested):

JunkyDeveloper (approved):
Tested the generated JSON is identical
Tested the blocks in game
Good addition to the macro, so approval from my side

Extractor PR: https://github.com/Steel-Foundation/SteelDocs/pull/46

coco875 (approved):

PR #332, opened 2026-08-09
Validate chat profile keys with Mojang service keys

### LINKED ISSUE #328: Chat session updates bypass validation and can exhaust memory

kdcokenny:
I was checking the remaining chat session validation path and noticed profile keys were accepted with `NoValidation`, so any signature succeeded.

Fixed it by loading and refreshing Mojang's service keys and validating profile-key certificates with Authlib's `SHA1withRSA` behavior. Missing keys, expiry downgrades, and invalid signatures now match vanilla, and secure chat is only advertised when validation is available.

Closes #328


coco875:
add also a pr to the doc

steel-protocol/src/packets/status/c_status_response.rs
```diff
@@ -29,6 +29,7 @@ pub struct Status {
     pub players: Option<Players>,
     pub version: Option<Version>,
     pub favicon: Option<String>,
+    #[serde(rename = "enforcesSecureChat")]
```
coco875: hmm maybe you can do rename camelCase ?

steel-core/src/server/service_keys.rs
```diff
@@ -0,0 +1,248 @@
...
+                Ok(()) => true,
+                Err(error) => {
+                    log::warn!("Failed to load Minecraft services public keys: {error}");
+                    false
+                }
+            };
+            let _ = ready_tx.send(());
+
+            let mut failure_count = 0;
+            loop {
```
coco875, line 101: no timeout ?

coco875 (changes requested):
not really update with how minecraft secure chat so will do a second deeper look when I have time

steel-core/src/server/service_keys.rs
kdcokenny: The client already has Authlib’s separate 5-second connect and read timeouts at lines 66–68. Did you mean a total request deadline or cancellation of an in-flight refresh?

kdcokenny:
> add also a pr to the doc

done

steel-core/src/server/service_keys.rs
coco875: more that if microsoft key are offline (that happen more frequently recently) not have the server in an infinite loop

kdcokenny: It’s capped at one retry every 320 minutes (5h 20m), matching [Authlib 9.0.75](https://libraries.minecraft.net/com/mojang/authlib/9.0.75/authlib-9.0.75.jar). It retries indefinitely so it can recover without a restart; it never spins.

coco875:
just to continue the conversation more clearly, how does auth work ? when a server and a client connect they check the key from the same source to make a secure chat ? and the key expire after x time ? so the server need to refresh it periodically ? or it don't expire so it's just try to get the key once ? don't fully understand how auth work

kdcokenny:
> just to continue the conversation more clearly, how does auth work ? when a server and a client connect they check the key from the same source to make a secure chat ? and the key expire after x time ? so the server need to refresh it periodically ? or it don't expire so it's just try to get the key once ? don't fully understand how auth work

Mojang gives the client a temporary signed chat key, and the server downloads Mojang’s public keys to verify it before trusting signed messages. Client keys expire and are replaced, while servers refresh Mojang’s verification keys every 24 hours in case they change.

coco875:
ok so if the server more then 24h to still maintain secure chat it need to refresh so it's why there is this infinite loop ?

coco875:
ok no it's just try again and again to get it but once it get it, it end the loop ?

kdcokenny:
It retries with backoff when it fails, but once it succeeds, the keys are refreshed every 24 hours in case Mojang changes them.

steel-core/src/server/service_keys.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    /// Starts the initial fetch and returns a signal completed after its first attempt.
+    pub(super) fn start(
+        self: &Arc<Self>,
+        cancel_token: CancellationToken,
+    ) -> oneshot::Receiver<()> {
+        let (ready_tx, ready_rx) = oneshot::channel();
+        let store = Arc::clone(self);
+        drop(tokio::spawn(async move {
+            let mut has_snapshot = match store.refresh().await {
```
coco875: I think that this local variable aren't named correctly, has_service_key ?

steel-core/src/server/service_keys.rs
```diff
@@ -0,0 +1,248 @@
...
+                    log::warn!("Failed to load Minecraft services public keys: {error}");
+                    false
+                }
+            };
+            let _ = ready_tx.send(());
+
+            let mut failure_count = 0;
+            loop {
+                let delay = if has_snapshot {
+                    REFRESH_INTERVAL
```
coco875: maybe rename the const to DAILY_REFRESH_INTERVAL ?

coco875 (changes requested):

coco875:
thanks for your explanation

steel-core/src/server/service_keys.rs
kdcokenny: well a valid response can contain no usable key but still count as a successfully loaded snapshot. How does `has_successful_snapshot` sound?

kdcokenny:
> thanks for your explanation

ofc

steel-core/src/server/service_keys.rs
coco875: how ? because there is profile_key_validator in refresh

kdcokenny: The request can succeed even if Mojang returns an empty key list. We received a valid response but still have no key, so `has_service_key` would be a bit misleading.

coco875: ok then has_successful_snapshot

coco875 (dismissed):

Cargo.toml
```diff
@@ -106,7 +106,7 @@ reqwest = { version = "0.13.4", features = [
 ], default-features = false }
 
 # Cryptography
-rsa = { version = "0.10.0-rc.18", features = ["sha2"] }
+rsa = { version = "0.10.0-rc.18", features = ["sha1", "sha2"] }
```
JunkyDeveloper, line 109: let me guess Mojang is using sha1?

kdcokenny: yep, Authlib uses SHA1withRSA for player-key certificates. Chat signatures use SHA256withRSA.

steel-core/src/player/chat/mod.rs
```diff
@@ -437,7 +482,40 @@ impl Player {
...
+            Ok(self.0)
+        }
+    }
+
+    fn session(expires_at_seconds: u64) -> profile_key::RemoteChatSessionData {
+        let (_, public_key) = generate_key_pair().expect("test player key should generate");
+        profile_key::RemoteChatSessionData {
+            session_id: Uuid::new_v4(),
+            profile_public_key: profile_key::ProfilePublicKeyData::new(
+                UNIX_EPOCH + Duration::from_secs(expires_at_seconds),
```
JunkyDeveloper: why not `system_time_from_millis`

JunkyDeveloper (changes requested):

steel-core/src/player/chat/mod.rs
kdcokenny: good call, i'll use it

JunkyDeveloper (approved):

kdcokenny:
@JunkyDeveloper sorry didnt mean to request a review from you lol mislick when clicking coco

JunkyDeveloper:
fix lint ;)

JunkyDeveloper:
ok isn't your failure

coco875 (approved):

PR #333, opened 2026-08-09
Implement falling block behavior

kdcokenny:
I was checking the remaining block behavior coverage and noticed Steel did not implement the vanilla falling-block foundation, so sand, gravel, and concrete powder never became falling entities.

Implemented the shared falling-block behaviors and `FallingBlockEntity`, including scheduling, movement, landing, drops, NBT, collision damage, and concrete powder hardening. This covers sand, red sand, gravel, and all 16 concrete powders.

Added focused tests for scheduling, stacked falls, water entry, placement, and serialization. I also validated sand, gravel, stacked falling blocks, dry powder, powder beside water, and falling powder entering water through an automated Minecraft Java 26.2 client and a separate manual client run.

Stacked falling blocks | Solidified concrete
:--: | :--:
Stacked falling blocks | Concrete powder solidified after entering water


steel-core/src/behavior/blocks/falling/colored_falling_block.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub const fn new(block: BlockRef, dust_color_rgba: i32) -> Self {
+        Self {
+            falling: FallingBlock::new(block),
+            dust_color_rgba,
+        }
+    }
+
+    /// Returns the extracted falling-dust color used by vanilla clients.
+    #[must_use]
+    pub const fn dust_color_rgba(&self) -> i32 {
```
coco875: it's use somewhere ?

coco875: because if no then we can remove it and only failling block

coco875 (changes requested):

steel-core/src/behavior/blocks/falling/colored_falling_block.rs
kdcokenny: done! turns out the colored was used only by the client for particles, not the server

coco875 (approved):

JunkyDeveloper (approved):

PR #336, opened 2026-08-09
Enforce pre-play packet sequencing

### LINKED ISSUE #335: Out-of-order pre-play packets bypass authentication and can shut down the server
I was reviewing pre-play packet handling and found Steel accepts login and configuration packets based only on the broad protocol state, not the expected step.

This allows two unauthenticated paths:

- `Hello → LoginAcknowledged → FinishConfiguration` skips `Key` and Mojang authentication. In online mode, Steel trusts the client-provided UUID and then loads that player’s saved data and permissions.
- `LoginAcknowledged → FinishConfiguration` reaches `expect("Game profile is empty")`; Steel’s panic hook then shuts down the server.

An authenticated client can also send `FinishConfiguration` before the known-pack response, skipping registry and tag synchronization.

Vanilla 26.2 uses explicit login states and configuration tasks. Paper preserves those guards through its authentication, proxy, and configuration extensions.

I think this should be one PR:

- [ ] Add typed login/configuration phases and accept each packet only from its expected phase.
- [ ] Keep the requested username separate from the authenticated profile. Only Mojang authentication or Steel’s offline UUID derivation may establish the trusted profile.
- [ ] Require known-pack negotiation and registry/tag publication before accepting `FinishConfiguration`.
- [ ] Disconnect on skipped, repeated, or out-of-order packets without advancing state or panicking.
- [ ] Remove the network-reachable profile `expect`.
- [ ] Add regression tests for valid online/offline flows and the skipped/repeated attack sequences.

Compression limits, Mojang HTTP retries, generic malformed-packet handling, and outbound backpressure should remain separate.

Research:

- [Steel login packet dispatch](https://github.com/Steel-Foundation/SteelMC/blob/e0cf0409f444df1181172e517898d62ef21dab14/steel-login/src/tcp_client.rs#L545-L559)
- [Steel trusts the pre-authentication profile](https://github.com/Steel-Foundation/SteelMC/blob/e0cf0409f444df1181172e517898d62ef21dab14/steel-login/src/handlers/login.rs#L24-L68)
- [Steel configuration finish path](https://github.com/Steel-Foundation/SteelMC/blob/e0cf0409f444df1181172e517898d62ef21dab14/steel-login/src/handlers/config.rs#L94-L121)
- [Vanilla login states](https://github.com/Renekovski/26.2-mcp/blob/51f3128ba7265299dce20b9abd2b37d7dd27d096/src/net/minecraft/server/network/ServerLoginPacketListenerImpl.java#L119-L267)
- [Vanilla configuration tasks](https://github.com/Renekovski/26.2-mcp/blob/51f3128ba7265299dce20b9abd2b37d7dd27d096/src/net/minecraft/server/network/ServerConfigurationPacketListenerImpl.java#L160-L229)
- [Paper login-state guard](https://github.com/PaperMC/Paper/blob/6c8d413223ba7fa2dc1c508f96741f49950a90d0/paper-server/patches/sources/net/minecraft/server/network/ServerLoginPacketListenerImpl.java.patch#L359-L363)

kdcokenny:
I was reviewing pre-play packet handling and noticed Steel accepted login and configuration packets based only on the broad protocol state. `finish_configuration` described a missing profile as “should be impossible,” but sending `LoginAcknowledged` before `Hello`, followed by `FinishConfiguration`, reached that panic and shut down the server.

Other out-of-order packets could skip Mojang authentication and enter play with a client-provided UUID. Exploitation requires a custom client or packet proxy, but no valid Mojang session: we reproduced an unauthenticated connection reaching play with an attacker-chosen UUID, which Steel then uses to resolve saved data and UUID-bound permissions.

Fixed it with explicit login and configuration phases that own the trusted profile and expected next packet. Invalid sequences now disconnect without advancing state or panicking.

Red/green validation:

- `LoginAcknowledged → FinishConfiguration`: server shutdown → client disconnected
- Skipped `Key` and authentication: reached play → disconnected during login
- Skipped known-pack negotiation: reached play → disconnected during configuration
- A correctly sequenced offline login still reaches play

Closes #335


steel-login/src/handlers/config.rs
```diff
@@ -71,7 +71,14 @@ impl JavaTcpClient {
     }
 
     /// Handles the select known packs packet during the configuration state.
-    pub async fn handle_select_known_packs(&self, packet: SSelectKnownPacks) {
+    pub(crate) async fn handle_select_known_packs(
```
coco875: why reduce the scope ?

coco875 (changes requested):
From what I understand you have add a state machine for the handshake protocol of mincraft and avoid interpret some packet until it finish, does it include all packet like player movement ? so to avoid have a false valid client ?

kdcokenny:
> From what I understand you have add a state machine for the handshake protocol of mincraft and avoid interpret some packet until it finish, does it include all packet like player movement ? so to avoid have a false valid client ?

This intentionally covers login and configuration transitions, not play packets like movement. Movement packets are only decoded after the authenticated profile is accepted, configuration finishes, and the connection upgrades to the play listener.

steel-login/src/handlers/config.rs
kdcokenny: ya you're right, this visibility change is unrelated to the fix. I’ll revert it.

coco875 (approved):

JunkyDeveloper (approved):

PR #337, opened 2026-08-10
Fix flaky set block publication test

### LINKED ISSUE #325: Flaky `set_block` client publication test
`world::tests::set_block_matches_vanilla_update_limit_and_client_publication_gates` occasionally fails with `6 != 3`, but passes on rerun.

The test uses the shared `test_world()` and checks a chunk-wide packet revision that also changes during light publication. Parallel tests can advance the shared world and affect the revision.

Likely fix: use a `fresh_test_world()` and settle lighting before checking exact revisions.

Failed in: https://github.com/Steel-Foundation/SteelMC/actions/runs/31272312819/attempts/1

kdcokenny:
I was looking into the flaky `set_block` test from #325 and reproduced its `left: 6`, `right: 3` failure on the third parallel suite run.

The test was using the shared world and checking a chunk-wide packet revision while background generation lighting could still update it.

Fixed it by using a fresh world, waiting for its setup work to finish, and settling the expected lighting before measuring block publication.

Red/green validation:

- Before: `cargo test -p steel-core --lib -- --test-threads=4` failed on run 3
- After: the same parallel suite passed 10 consecutive runs, followed by 3 more on current `master`

Closes #325


coco875 (approved):

kdcokenny:
@JunkyDeveloper super simple pr that fixes the flaky test we found earlier, mind giving it a quick look over?

JunkyDeveloper (approved):

PR #338, opened 2026-08-11
fix abilities on respawn

CuzImClicks:
## Type of change

- [x] Bug fix

## Description

When joining a server and being in survival, your abilities would be reset to the normal ones, despite what the server had saved locally. This is caused because sending a CGameEvent with ChangeGameMode after sending abilities resets them to the default gamemode abilities. Since game mode is already included in CRespawn and CLogin we dont have to send it again.

## How this was tested

Before: /gamemode survival; /fly; leave; join; try to fly (it wont work); /fly; /fly => you can now fly

Now: /gamemode survival; /fly; leave; join; => you can fly

## Checklist

- [x] Self-reviewed the diff


coco875 (approved):

JunkyDeveloper (approved):
I didn't tested it why this behaviour comes but the what I think it make sense
@coco875 I trust you now that you tested that

PR #340, opened 2026-08-11
chore(command): make methods that return an argument's parsed value only return `Result`

SomeYellowGuy:
## Type of change

- [x] Refactor / code cleanup
- [x] Chore / tooling

## Description

This PR makes methods returning `Option<T>` for getting a parsed argument's value return a `Result<T, CommandSyntaxError>` (or `Result<T, CommandError>` for the API) instead.

This reduces the code redundancy of having many `missing_argument` functions and `let Some(...) else` clauses scattered everywhere, and, in general, is more faithful to what Mojang's *Brigadier* library does (it throws `IllegalArgumentException`s for errors like these).

Before:
```rust
let Some(anchor) = context.entity_anchor("facingAnchor") else {
     return Err(missing_argument("facingAnchor"));
};
```

After:
```rust
let anchor = context.entity_anchor("facingAnchor")?;
```

## How this was tested

Ran some commands like:
- `/give @s diamond 12`
- `/gamerule advance_weather true`
- `/execute as @e run tp ~ ~2 ~`

They run well, with no issues during execution.

## Checklist

- [x] Self-reviewed the diff

## Additional notes

Technically, the `CommandSyntaxError`s that could be thrown should not appear under normal gameplay conditions. They are only thrown in the case of improperly written commands; for example, having a command's execution try getting the value of an argument `"target"` when no argument with the name `"target"` exists for the specific part of the command tree:

```rust
fn command() -> CommandNodeBuilder<CommandSource, SteelCommandRuntime> {
    literal("example").then(
        argument("target2", SteelArgumentType::entity()).executes(execution)
    )
}

fn execution() -> Result<i32, CommandSyntaxError> {
    let target = context.entity("target")?;
    Ok(1)
}
```


CuzImClicks:
please make sure ci passes => run clippy and all tests

coco875 (approved):
very good refactor pr, it allow to dedup missing_argument function and also simplify structure of code for command

JunkyDeveloper (approved):

PR #341, opened 2026-08-11
feat(item): implement the item behavior of name tags

SomeYellowGuy:
## Type of change

- [ ] Block implementation
- [x] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

Implements the right-click behavior of name tags: right-clicking an entity saves the custom name of the item to that of the entity. Also, if the entity is a mob, it sets the entity to be persistent.

## How this was tested

1. Run the server.
2. Use `/summon pig` to summon a pig.
3. Get a name tag and rename it with an anvil. Then, right-click the pig with the name tag.

You will see that the pig's custom name is set to that of the name tag.

I couldn't test whether the persistence actually worked (I could not get a non-tagged pig to despawn), but I verified that the code responsible for setting persistence runs when required.

## Screenshots / logs

name_tag_1
name_tag_2

## Checklist

- [x] Self-reviewed the diff


coco875:
did you have try _jeb or dinnerbone ?

SomeYellowGuy:
> did you have try _jeb or dinnerbone ?

No, but I just tried `Dinnerbone`, and it seems to work (without me doing anything). I'm guessing these names are checked *client-side*. Although I can't test `jeb_` (sheep don't spawn in Steel), it should still work because it seems to be controlled by `SheepRenderer`.

coco875:
I think jeb_ are also client-side

coco875 (approved):

JunkyDeveloper (approved):

PR #342, opened 2026-08-11
Bound compressed packet decompression

kdcokenny:
## Type of change

- [x] Bug fix

## Description

Vanilla only decompresses up to the packet's claimed size. Paper does the same when using its native compression code. Steel previously had no such limit, so a packet under 32 KiB could expand to 16 MiB.

This now limits decompression to the claimed size and rejects packets when the size does not match.

See [Paper's compression code](https://github.com/PaperMC/Paper/blob/c9e894d3cc03f21f80de4f4061a795e11941e89a/paper-server/patches/features/0007-Use-Velocity-compression-and-cipher-natives.patch#L88-L152).

## How this was tested

MC 26.2, Linux.

- Red: the 16 MiB packet was accepted before the fix.
- Green: oversized and short packets are now rejected.
- Valid compressed packets still work.
- `cargo test -p steel-protocol`
- `cargo check --workspace`
- `cargo clippy -p steel-protocol --all-targets --all-features -- -D warnings`

## Screenshots / logs

N/A

## Checklist

- [x] Self-reviewed the diff

## Additional notes

No behavior change for valid clients.


JunkyDeveloper (approved):
Good catch

As I understand if a client send a too long package the client will be disconnected, which I think is a good approach, but can maybe interfere with mods

kdcokenny:
> Good catch
> 
> As I understand if a client send a too long package the client will be disconnected, which I think is a good approach, but can maybe interfere with mods

Just for clarification: the size limits were already there. This just makes sure the decompressed size matches what the client claimed, so valid modded packets shouldn’t be affected.

coco875 (approved):
look good, it's your pr that remind of Vec vs Box<[]> idea (talk on discord).

PR #344, opened 2026-08-12
Implement chiseled bookshelf behavior

kdcokenny:
## Type of change

- [x] Block implementation

## Description

Implements vanilla chiseled bookshelf behavior.

Includes all six slots, item storage, sounds, block-state updates, comparator output, persistence, drops, and container automation.

## How this was tested

Minecraft 26.2 on Linux.

- `cargo test -p steel-core chiseled_bookshelf --no-fail-fast`
- `cargo check -p steel-core`
- `cargo clippy -p steel-core --all-targets --all-features`
- `cargo fmt --all --check`
- Real-client test covering placement, all six slots, invalid items, enchanted books, comparator strengths 1–6, drops, and reconnect persistence

## Screenshots / logs

All slots occupied:

Chiseled bookshelf with all six slots occupied

Comparator fixture after a slot 6 interaction:

Powered comparator fixture after slot 6 interaction

Stored books dropped after mining:

Stored books dropped after mining

## Checklist

- [x] Self-reviewed the diff

## Additional notes

No menu was added. The block is used directly, like vanilla.


steel-core/src/behavior/blocks/container/chiseled_bookshelf_block.rs
```diff
@@ -0,0 +1,835 @@
...
+    }
+
+    fn section_index(normalized_coordinate: f64, section_count: i32) -> usize {
+        let targeted_pixel = normalized_coordinate as f32 * Self::PIXELS_PER_BLOCK_EDGE;
+        let section_size = Self::PIXELS_PER_BLOCK_EDGE / section_count as f32;
+        ((targeted_pixel / section_size).floor() as i32).clamp(0, section_count - 1) as usize
+    }
+
+    fn hit_slot(state: BlockStateId, hit_result: &BlockHitResult) -> Option<usize> {
+        let facing = state.get_value(&BlockStateProperties::HORIZONTAL_FACING);
```
JunkyDeveloper: do you remember a PR with 130 changed files, which addresses exact that problem? xD

steel-core/src/behavior/blocks/container/chiseled_bookshelf_block.rs
```diff
@@ -0,0 +1,835 @@
...
+            return &sound_events::BLOCK_CHISELED_BOOKSHELF_PICKUP_ENCHANTED;
+        }
+        &sound_events::BLOCK_CHISELED_BOOKSHELF_PICKUP
+    }
+}
+
+impl BlockBehavior for ChiseledBookShelfBlock {
+    fn get_state_for_placement(&self, context: &BlockPlaceContext<'_>) -> Option<BlockStateId> {
+        Some(self.block.default_state().set_value(
+            &BlockStateProperties::HORIZONTAL_FACING,
```
JunkyDeveloper: also here

JunkyDeveloper (changes requested):
Claude finding:
 > Vanilla's default (ListBackedContainer.canPlaceItem, inherited since the block entity doesn't override it) is:
acceptsItemType(itemStack) && (getItem(slot).isEmpty() || getItem(slot).getCount() < getMaxStackSize(itemStack))
i.e. it also requires the target slot be empty (max stack size is 1, so "count < max" is only true when empty). Steel's version returns true for an already-occupied slot as long as the item type matches. Currently latent — nothing in the codebase calls can_place_item on this container in a way that would overwrite an occupied slot (no hopper implementation yet, and Container::add's empty-slot pass checks occupancy separately before calling it) — but it's a real contract deviation from vanilla that will misbehave once hopper/automation insertion into containers is implemented directly against can_place_item.

Also verified that behaviour, so this can be now a trash can, to take for ever items

steel-core/src/block_entity/entities/chiseled_bookshelf.rs
```diff
@@ -0,0 +1,561 @@
...
+
+    fn get_max_stack_size(&self) -> i32 {
+        MAX_BOOKS_PER_SLOT
+    }
+
+    fn set_changed(&mut self) {}
+
+    fn can_place_item(&self, _slot: usize, stack: &ItemStack) -> bool {
+        stack.item().has_tag(&ItemTag::BOOKSHELF_BOOKS)
+    }
```
JunkyDeveloper, line 345: This is the function

steel-core/src/behavior/blocks/container/chiseled_bookshelf_block.rs
kdcokenny: Updated to use the local property alias.

steel-core/src/behavior/blocks/container/chiseled_bookshelf_block.rs
kdcokenny: Updated to use the local property alias.

steel-core/src/block_entity/entities/chiseled_bookshelf.rs
kdcokenny: Fixed. Books are now only accepted in empty slots, matching vanilla. I added a test for occupied slots.

JunkyDeveloper (approved):

coco875 (approved):

PR #348, opened 2026-08-12
opti init_grid in native

coco875:
## Type of change

- [x] Performance improvement

## Description

init_grid use a loop of x and z (25 iteration), so I have the idea of why not making it simd, the full details of everything I have try are in Additional notes. But endup with enable init_grid_simd only for architecture with avx or more.

## How this was tested

I have add new benchmark to test it and also I have use pregen as the final comparaison

## Checklist

- [x] Self-reviewed the diff

## Additional notes

@ -0,0 +1,58 @@
# Worldgen SIMD experiments

Benchmarks used the real generated Overworld density functions. Full pregen
results used a fresh world, seed `-1`, `PREGEN_SIZE=301`, and default settings.

## Results

| Strategy | Result | Conclusion |
|---|---:|---|
| Fresh pregen, non-native scalar | 1,802.3 chunks/s | Baseline |
| Fresh pregen, native SIMD8 | 1,818.1 chunks/s | 0.9% faster; small end-to-end gain |
| X/Z SIMD8, SSE4.1 | 11.9% faster | Useful with an explicit ISA |
| X/Z SIMD8, AVX | 19.4% faster | Useful with an explicit ISA |
| X/Z SIMD8, AVX2 | 22.7% faster | Best explicit-ISA result |
| X/Z SIMD8, native | 20.9% faster | Best tested logical width |
| Non-native scalar X/Z | 7.970 ms | Isolated baseline |
| Non-native SIMD1 X/Z | 9.517 ms; 19.4% slower | Portable SIMD overhead exceeds any benefit |
| Non-native SIMD2 X/Z | 10.083 ms; 26.5% slower | Portable baseline generates poor code |
| Non-native SIMD4 X/Z | 7.660 ms; 3.9% faster | Wins in isolation, not proven in full pregen |
| Non-native SIMD8 X/Z | about 6.972 ms; 12.5% faster | Wins in isolation, not proven in full pregen |
| Non-native SIMD25 X/Z | 11.157 ms; 40.0% slower | One logical vector for the whole grid is too wide |
| X-only SIMD4 (`4+1` per row) | 0.6% slower than flattened SIMD4 | Five scalar tails remove the benefit |
| Y-density scalar, non-native | 1.0698 ms | Isolated baseline |
| Y-density SIMD4, non-native | 870.7 us; 18.6% faster | Clear improvement |
| Y-density scalar, native | 976.1 us | Native isolated baseline |
| Y-density SIMD4, native | 794.1 us; 18.6% faster | Clear improvement with native too |

An earlier result near 12,000 chunks/s was invalid because it reused generated
chunks and measured loading rather than generation.

## Strategies tested

- **Flattened X/Z SIMD:** Processes the 25 grid positions as complete batches
  followed by one scalar remainder (`4+4+4+4+4+4+1` for SIMD4). This was faster
  than processing each Z row independently and is retained.
- **X-only SIMD:** Processes every row as SIMD X values plus a scalar tail
  (`4+1`, repeated five times). It was slower because it creates five tails.
- **Portable/default SIMD:** Small widths were slower than the original scalar
  loop due to vector setup, lane movement, and weaker baseline code generation.
  SIMD4 and SIMD8 win the isolated benchmark, but SIMD25 becomes slower because
  it must be lowered across several hardware vectors and increases register
  pressure. The isolated wins lack a confirmed default-build full-pipeline gain.
- **Explicit-ISA SIMD:** SSE4.1, AVX, AVX2, and native builds generate much
  better SIMD8 code. Compile-time SIMD8 is used for these targets.
- **Y-density SIMD:** Four adjacent Y corners are evaluated together. It is
  consistently 18.6% faster and produces identical results, so it is retained.
- **Runtime ISA dispatch:** Separate ISA functions and cached function pointers
  added complexity without improving full pregen. The runtime dispatch was
  removed.

## Retained implementation

- Default portable builds use the original scalar X/Z grid initialization.
- AVX, AVX2, or AVX-512F builds select X/Z SIMD8 at compile time.
- Y-density evaluation keeps its fixed four-lane SIMD path.
- X/Z SIMD uses complete flattened batches and the original scalar code for the
  exact remainder.
- No runtime detection, indirect calls, `OnceLock`, or per-ISA dispatch files.


steel-core/benches/worldgen.rs
```diff
@@ -1864,6 +1866,318 @@ fn bench_noise_kernel(c: &mut Criterion) {
...
+                    black_box(&cache);
+                });
+            });
+        };
+    }
+    group.bench_function("scalar", |b| {
+        let mut cache = OverworldColumnCache::new();
+        b.iter(|| {
+            for &(x, z) in black_box(&grid_origins) {
+                sampler.init_column_grid(&mut cache, x, z);
```
kdcokenny, line 1916: why does the `scalar` benchmark call `init_column_grid`? on AVX builds that dispatches to SIMD-8, and the oracle above is SIMD-25, so what is this comparing against the scalar path?

steel-worldgen/build/density/transpiler/codegen_expr.rs
```diff
@@ -21,9 +21,266 @@ use super::TranspilerInput;
...
+            DensityFunction::RangeChoice(r) => {
+                let value = self.gen_expr_xz(&r.input, input);
+                let in_range = self.gen_expr_xz(&r.when_in_range, input);
+                let out_of_range = self.gen_expr_xz(&r.when_out_of_range, input);
+                let min = Literal::f64_unsuffixed(r.min_inclusive);
+                let max = Literal::f64_unsuffixed(r.max_exclusive);
+                quote! {{
+                    let v = #value;
+                    (v.simd_ge(Simd::splat(#min)) & v.simd_lt(Simd::splat(#max)))
+                        .select(#in_range, #out_of_range)
```
kdcokenny, line 147: does `select` evaluate both operands before masking here? would that make inactive branches and every emitted spline interval run, unlike vanilla?

steel-utils/src/lib.rs
```diff
@@ -68,6 +68,21 @@ pub mod version;
...
 pub const MC_VERSION: &str = version::MINECRAFT_VERSION;
 
+/// Number of `f64` lanes in the widest SIMD register enabled for this build.
+///
+/// This describes the compilation target, not features detected on the CPU at
+/// runtime. Native builds can therefore expose a wider register than portable
+/// builds running on the same machine.
+pub const SIMD_REGISTER_F64_SIZE: usize = if cfg!(target_feature = "avx512f") {
+    8
+} else if cfg!(target_feature = "avx2") {
```
kdcokenny: why does this check AVX2 but not AVX when the X/Z path enables SIMD for AVX? would AVX-only builds use different lane-width policies?

coco875: yeah, I have done it because I have avx2, ideally I should test on each cpu and see what are the best for the highest level of isa

steel-core/benches/worldgen.rs
coco875: because it's the original version of it and behave differently then init_column_grid_simd<1>

steel-worldgen/build/density/transpiler/codegen_expr.rs
coco875: yes, I have made try and it's more effiecient to put like that then doing the un-SIMD way

steel-core/benches/worldgen.rs
kdcokenny: could this also compare against the original scalar output? right now expected and all the compared values use SIMD paths.

coco875: normally if you don't bench with native it should use the non-simd version

JunkyDeveloper:

coco875:
I will do after 26.3 because it will likely change rust

PR #349, opened 2026-08-12
perf(protocol): Eliminate inbound packet payload copies

kdcokenny:
## Type of change

- [x] Performance improvement

## Description

I was looking into the packet buffer allocations and noticed Steel copied every inbound packet payload after reading its packet ID. When compression was enabled but a packet was sent uncompressed, the payload was copied twice.

`RawPacket` now keeps the packet's existing allocation as a boxed buffer with a payload offset and exposes the payload as a slice. This removes both copies without introducing shared ownership or buffer pooling.

## How this was tested

Minecraft 26.2 on Linux, using an Intel i7-1260P.

- `cargo test -p steel-protocol`
- `cargo test -p steel-core connection::java::tests`
- `cargo clippy -p steel-protocol -p steel-login -p steel-core --all-targets --all-features`
- `cargo fmt --all --check`
- `typos` on the changed files
- DHAT allocation benchmarks over 10,000 packets per case
- CPU-pinned, interleaved release-mode before/after timing benchmarks

### Allocations

| Packet case | Allocations per packet | Allocation delta | Bytes allocated per packet | Byte delta |
|---|---:|---:|---:|---:|
| Plain, 32 B | 2 → 1 | −1 (50.0%) | 65 → 33 | −32 B (49.2%) |
| Plain, 255 B | 2 → 1 | −1 (50.0%) | 511 → 256 | −255 B (49.9%) |
| Compression enabled, uncompressed, 32 B | 3 → 1 | −2 (66.7%) | 99 → 34 | −65 B (65.7%) |
| Compression enabled, uncompressed, 254 B | 3 → 1 | −2 (66.7%) | 765 → 256 | −509 B (66.5%) |
| Compressed, 1 KiB | 5 → 4 | −1 (20.0%) | 78,737 → 77,713 | −1,024 B (1.3%) |
| Compressed, 32 KiB | 5 → 4 | −1 (20.0%) | 142,746 → 109,978 | −32,768 B (23.0%) |

The removed allocation sizes match the copied payloads. Converting the retained buffers to `Box<[u8]>` did not add an allocation. The compressed totals are dominated by the 76,064 bytes allocated for zlib state per packet, which this PR does not change.

### Decode time

Both binaries were pinned to the same i7-1260P performance core. Each result is the median of five interleaved before/after runs with alternating execution order. Each run used seven samples of 100,000 packets.

| Packet case | Before | After | Difference |
|---|---:|---:|---:|
| Plain, 32 B | 23.56 ns | 28.12 ns | 19.4% slower |
| Plain, 255 B | 34.84 ns | 34.04 ns | 2.3% faster |
| Compression enabled, uncompressed, 32 B | 33.32 ns | 28.66 ns | 14.0% faster |
| Compression enabled, uncompressed, 254 B | 44.47 ns | 35.09 ns | 21.1% faster |
| Compressed, 1 KiB | 6.163 µs | 6.171 µs | 0.1% slower |
| Compressed, 32 KiB | 11.112 µs | 10.411 µs | 6.3% faster |

The smallest plain-packet case regressed. At that size, the removed allocation and copy are extremely cheap relative to the additional offset handling. The 1 KiB compressed result was effectively unchanged, while the larger and compression-negotiated copy-heavy paths improved.

## Checklist

- [x] Self-reviewed the diff

## Additional notes

I initially retained the packet in a `Vec<u8>`, but adding the payload offset increased `RawPacket` from 32 to 40 bytes. The retained implementation uses `Box<[u8]>` and a `u32` offset, reducing `RawPacket` to 24 bytes while still avoiding the payload copies. The offset is sufficient because packet buffers are limited to 8 MiB.

Compressed packets still allocate for zlib state. Reusing the decompressor is a separate optimization and is intentionally not included here.


JunkyDeveloper (approved):

coco875:
can you add the delta because seems weird at some place for number and didn't dhat-rs not update since 1 years ?

kdcokenny:
I only used DHAT’s `GlobalAlloc` wrapper to count allocations and bytes; timing was measured separately with mimalloc. Both were purely for benchmarking and aren’t added to Steel. I’ve added explicit deltas and clarified that zlib’s 76,064-byte state allocation dominates the compressed totals.

coco875 (approved):

PR #353, opened 2026-08-13
fix some magic number

coco875:
## Type of change

- [x] Refactor / code cleanup

## Description

less magic number

## How this was tested

it compile

## Checklist

- [x] Self-reviewed the diff


JunkyDeveloper:
Fix lint and tests!

coco875:
done x)

teknostom (dismissed):
LGTM

CuzImClicks:
also fix typo in title

coco875:
I just see that I have start to write magick instead of magic

steel-core/src/behavior/blocks/redstone/pressure_plate/standard.rs
```diff
@@ -60,7 +61,11 @@ impl PressurePlateBlock {
     }
 
     fn signal_for_state(state: BlockStateId) -> i32 {
-        if state.get_value(POWERED) { 15 } else { 0 }
+        if state.get_value(POWERED) {
+            MAX_REDSTONE_SIGNAL
+        } else {
+            0
```
JunkyDeveloper: Why nothing here???? You complained about that here xD

steel-core/src/behavior/blocks/redstone/pressure_plate/standard.rs
```diff
@@ -75,7 +80,7 @@ impl PressurePlateBlock {
                 // entities become eligible when they gain `LivingEntity` behavior.
                 PressurePlateSensitivity::Mobs => entity.is_living_entity(),
             });
-        if count > 0 { 15 } else { 0 }
+        if count > 0 { MAX_REDSTONE_SIGNAL } else { 0 }
```
JunkyDeveloper: Also here

steel-core/src/behavior/blocks/vegetation/beetroots.rs
```diff
@@ -46,7 +46,7 @@ impl CropLike for BeetrootBlock {
     }
 
     fn max_age(&self) -> u8 {
-        3
+        AGE_3.max
```
JunkyDeveloper: Don't like the that, I'm on my phone in some hours I can give a deeper look into that

steel-core/src/behavior/blocks/vegetation/stem_block.rs
```diff
@@ -83,7 +83,7 @@ impl StemBlock {
         }
 
         let growth_speed = crop_growth_speed(self.block, world, pos);
-        let growth_chance = (25.0 / growth_speed) as u32 + 1;
+        let growth_chance = (super::crop_block::CROP_GROWTH_CHANCE_BASE / growth_speed) as u32 + 1;
```
JunkyDeveloper: Why no use?

JunkyDeveloper (changes requested):
Will have a deeper look again later please say why redstone doesn't have a NO_Power constant

steel-core/src/behavior/blocks/redstone/pressure_plate/standard.rs
coco875: after thinking of it, not sure if it was necessary because 0 is already clear enough that it's nothing

steel-core/src/behavior/blocks/vegetation/stem_block.rs
coco875: because clippy didn't complain x) but yeah will see

steel-core/src/behavior/blocks/redstone/target_block.rs
```diff
@@ -112,7 +113,7 @@ impl BlockBehavior for TargetBlock {
         _moved_by_piston: bool,
     ) {
         if old_state.get_block() != state.get_block()
-            && state.get_value(POWER) > 0
+            && i32::from(state.get_value(POWER)) > MIN_REDSTONE_SIGNAL
             && !world.has_scheduled_block_tick(pos, self.block)
         {
```
JunkyDeveloper, line 118: why 2 different versions???

steel-core/src/behavior/blocks/vegetation/beetroots.rs
JunkyDeveloper: I have now the deep look and I still don't like it!
Rename it from AGE_3 to AGE or should I make a PR, but I think it would for us better you do it, because non of ours are merged soon

steel-core/src/behavior/blocks/vegetation/cactus_block.rs
```diff
@@ -138,7 +138,7 @@ impl BlockBehavior for CactusBlock {
             == &vanilla_blocks::CACTUS
         {
             height += 1;
-            if height == MAX_CACTUS_HEIGHT && age == 15 {
+            if height == MAX_CACTUS_HEIGHT && age == AGE_15.max {
                 return;
```
JunkyDeveloper, line 142: same here I still don't like that, use AGE not AGE_15

steel-core/src/behavior/blocks/vegetation/cactus_block.rs
```diff
@@ -159,7 +159,7 @@ impl BlockBehavior for CactusBlock {
                     UpdateFlags::UPDATE_ALL,
                 );
             }
-        } else if age == 15 && height < MAX_CACTUS_HEIGHT {
+        } else if age == AGE_15.max && height < MAX_CACTUS_HEIGHT {
```
JunkyDeveloper: here

steel-core/src/behavior/blocks/vegetation/cactus_block.rs
```diff
@@ -170,7 +170,7 @@ impl BlockBehavior for CactusBlock {
             world.neighbor_changed(above_pos, &vanilla_blocks::CACTUS);
         }
 
-        if age < 15 {
+        if age < AGE_15.max {
```
JunkyDeveloper: here

JunkyDeveloper (changes requested):
final judgment

steel-core/src/behavior/blocks/redstone/target_block.rs
coco875: what do you mean by two deifferent version ?

JunkyDeveloper: really it didn't added the other thing....
```rust
if state.get_value(POWER) != MIN_REDSTONE_SIGNAL as u8 {
```
is that not the same thing but different style?

coco875: so convert the const instead of the number that we get ?

JunkyDeveloper: I like to have everything a bite more structured, so we only allow as and not ::from for example, to make the code base more similar

coco875: the difference between from and as is from are cast that don't lose data while from, we can lose, change an i32 to u8 mean that you lost the sign, and 24 bit

coco875: clippy force from when there is no loss of information

JunkyDeveloper: than let's do only from and no as?

coco875: it's the purpose

JunkyDeveloper: or are you ok with both versions to have in our code, than I'm also fine, only want to raise that as a discussion

coco875: personally i am fine, you can see the exact rule here https://rust-lang.github.io/rust-clippy/master/index.html#cast_lossless

JunkyDeveloper: ok than I would resolve this discussion also from my side, maybe change all data types of these consts to the smallest datatype to save at least 20 bytes xD

JunkyDeveloper (approved):

steel-core/src/behavior/blocks/redstone/target_block.rs
coco875: if it's small number often const get inline because in asm you can write small number directly in register without reading it from somewhere in memory

teknostom (approved):
LGTM. Discussed seperatley.

coco875:
like the discussion was on discord not in a public channel I will make a resume. Clippy and other tool also have error/warning for magic number but we don't want to blindly make it for all number, only for key number of frequent one

PR #354, opened 2026-08-13
feat: Add in-memory gameplay test harness

kdcokenny:
## Summary

Adds a feature-gated, Flint-independent harness for running real Steel gameplay code without starting a network server.

- creates isolated RAM-only worlds with complete Steel bootstrap
- advances one production-equivalent tick at a time
- prepares production Full safety halos for block-ticking and entity-ticking chunks
- attaches real players through the normal world lifecycle
- exposes a recording connection for protocol-facing assertions
- applies daytime and game rules through a small test API
- cancels and drains background world work during shutdown

This is the Steel half of the working Flint-Steel MVP. The adapter remains in `flint-steel`; this PR adds no Flint dependency.

## Correctness details

`ensure_chunk` installs a 3x3 Full halo with only the center block-ticking. `ensure_entity_chunk` installs a 5x5 Full halo with only the center entity-ticking. Both go through holder publication and readiness reconciliation rather than directly forcing readiness states.

The regression suite includes scheduled water flow across x=15 to x=16, repeated create/use/drop cycles, exact tick advancement, player attachment, world configuration, and clean shutdown.

## Verification

- `cargo test -p steel-core --features test-harness test_harness::tests --no-fail-fast` (7 passed)
- `cargo clippy -p steel-core --features test-harness --all-targets -- -D warnings`
- `cargo check -p steel-core`
- `cargo fmt --all --check`
- `git diff --check`

The complete Flint-Steel runner was also tested five consecutive times against this exact commit, with 35 fixture executions and no skipped tests, panics, or leaked world-generation work.

### DIFF
--- a/steel-core/Cargo.toml
+++ b/steel-core/Cargo.toml
@@ -12,6 +12,7 @@ stand-alone = []
 slow_chunk_gen = []
 flint = []
 benchmark-support = []
+test-harness = []
 
 [dependencies]
 # Internal crates
--- a/steel-core/src/bootstrap.rs
+++ b/steel-core/src/bootstrap.rs
@@ -30,9 +30,8 @@ pub(crate) fn init_globals() -> Result<(), String> {
     Ok(())
 }
 
-/// Idempotent [`init_globals`] for tests and benchmarks, which bootstrap
-/// repeatedly in one process.
-#[cfg(any(test, feature = "benchmark-support"))]
+/// Idempotent [`init_globals`] for in-process test and benchmark runtimes.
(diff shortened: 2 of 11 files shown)


kdcokenny:
Working consumer for this API: FlintTestMC/flint-steel#25. The companion Flint execution contract is FlintTestMC/flint-core#82.

kdcokenny:
Closing after clarifying the intended Flint/Steel boundary. Steel should expose small generic APIs useful beyond Flint, while Flint-specific test orchestration stays outside Steel. Keeping the branch for reference.

PR #357, opened 2026-08-14
fix: respect block drops for face-directed items

### LINKED ISSUE #347: Face-directed item drops ignore the block-drops gamerule
I noticed using a hoe on rooted dirt can drop hanging roots while the `block_drops` gamerule is disabled.

I reproduced this in my client on the Steel server:

- With `block_drops = true`, rooted dirt became dirt and spawned one hanging-roots item.
- With `block_drops = false`, rooted dirt still became dirt, but incorrectly spawned the same item.

Steel’s normal item-drop helper checks the gamerule, but `pop_resource_from_face` does not. The rooted-dirt hoe interaction uses that unchecked helper.

Vanilla routes face-directed drops through the same gamerule-aware item-spawning logic as other block drops.

- [ ] Make `pop_resource_from_face` respect the `block_drops` gamerule.
- [ ] Preserve the rooted-dirt conversion to dirt when drops are disabled.
- [ ] Add a regression test proving no hanging-roots item is spawned when block drops are disabled.

Research:

- [Steel’s rooted-dirt hoe interaction](https://github.com/Steel-Foundation/SteelMC/blob/0708f9ccfc7d2b82a181a1dba23e2faa3fab14fd/steel-core/src/behavior/items/hoe.rs#L62-L70)
- [Steel’s unchecked face-drop helper](https://github.com/Steel-Foundation/SteelMC/blob/0708f9ccfc7d2b82a181a1dba23e2faa3fab14fd/steel-core/src/world/entity_management.rs#L395-L439)
- [Vanilla routes face drops through its gamerule-aware helper](https://github.com/Renekovski/26.2-mcp/blob/51f3128ba7265299dce20b9abd2b37d7dd27d096/src/net/minecraft/world/level/block/Block.java#L418-L434)

floze-the-genius:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

Add the missing `BLOCK_DROPS` gamerule guard to face-directed item drops. This matches Minecraft 26.2's shared drop helper, which applies the same guard to both regular and face-directed resource drops.

The regression test disables block drops and verifies that `pop_resource_from_face` does not create an item entity.

Closes #347.

## How this was tested

Environment: macOS arm64, Minecraft 26.2, `nightly-2026-07-23`.

- `cargo test -p steel-core face_directed_item_drops_respect_block_drops_game_rule`
- `cargo fmt --all --check`
- `RUSTFLAGS=-Dwarnings cargo clippy -r --all-targets --all-features`
- `RUSTFLAGS=-Dwarnings cargo test --verbose`

## Screenshots / logs

Not applicable.

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated (not applicable; behavior-only fix)

## Additional notes

None.


JunkyDeveloper (approved):

coco875 (approved):
ok, I have compare to vanilla and find it weird that we don't have todo so I guess a mistake, the pr are good

PR #363, opened 2026-08-14
Normalize all the line endings

coco875:
## Type of change

- [x] Bug fix
- [x] Refactor / code cleanup

## Description

change crlf to lf, I have add the file to force that but not have re-normalize file

## How this was tested

it compile

## Checklist

- [x] Code builds w/o errors or warnings
- [x] Self-reviewed the diff
- [x] Docs updated (if applicable)
- [x] No leftover debug code / comments

## Additional notes

I don't like windows

PR #366, opened 2026-08-15
refactor: Simplify test player creation

### LINKED ISSUE #364: Simplify test player creation
Most tests have to provide a UUID to `TestPlayerBuilder` even when player identity does not matter.

This leads to repeated `Uuid::from_u128(...)` values and small local helpers that only hide the setup.

- [ ] Add a simple test-player helper or builder path that handles the UUID internally.
- [ ] Keep explicit UUID support for tests that care about identity.
- [ ] Update straightforward call sites to use the simpler path.

Current builder:

- [`TestPlayerBuilder` requires an explicit UUID](https://github.com/Steel-Foundation/SteelMC/blob/0fd1769893988720fb76ab13ed1749ee092f595f/steel-core/tests/support/player.rs#L53-L72)

kdcokenny:
## Type of change

- [x] Refactor / code cleanup

## Description

Simplifies test player construction by generating UUIDs inside `TestPlayerBuilder::new`.

Explicit UUIDs are now optional builder configuration through `.uuid(uuid)` and are retained only where a test genuinely requires a predetermined identity, such as replacement sessions or UUID-keyed state prepared before player construction.

Straightforward test helpers and call sites now use generated player identities directly, removing arbitrary `Uuid::from_u128(...)` values and UUID parameters that were unrelated to the behavior under test.

Closes #364.

## How this was tested

Minecraft 26.2 on Linux.

- `cargo check -p steel-core --tests`
- `cargo test -p steel-core --lib command::builtins::invsee::tests` (17 passed)
- `cargo test -p steel-core --lib server::tests` (37 passed)
- `cargo fmt --all --check`
- `git diff --check`

## Screenshots / logs

N/A

## Checklist

- [x] Self-reviewed the diff

## Additional notes

This only changes test infrastructure and test setup; production behavior is unchanged.


steel-core/src/command/builtins/invsee/tests.rs
```diff
@@ -87,7 +86,7 @@ struct RecordingPlayer {
     callbacks_saw_unlocked_inventories: Arc<AtomicBool>,
 }
 
-fn recording_player(uuid: u128, name: &str, entity_id: i32) -> RecordingPlayer {
+fn recording_player(name: &str, entity_id: i32) -> RecordingPlayer {
     init_vanilla_registry();
```
coco875: by memory we have already an init registry somewhere no ?

steel-core/src/command/builtins/invsee/tests.rs
```diff
@@ -142,16 +136,11 @@ fn player_inventory_updates(packets: &SyncMutex<Vec<EncodedPacket>>) -> Vec<(i32
         .collect()
 }
 
-fn test_player(uuid: u128, name: &str, entity_id: i32) -> Arc<Player> {
+fn test_player(name: &str, entity_id: i32) -> Arc<Player> {
     init_vanilla_registry();
```
coco875: same

coco875: I talk specifically for this function that only call two other function

steel-core/src/entity/entities/mobs/passive/cow/tests/mod.rs
coco875, line 1: crlf... it would be fix by 363

coco875 (changes requested):
Very good, I wonder why we don't have done this simplification early to avoid have a bunch of uuid with 1 2 3, etc

steel-core/src/command/builtins/invsee/tests.rs
JunkyDeveloper: that is the new init_registry function

kdcokenny: Yep, `test_world()` already initializes the vanilla registry through the shared test-world setup, so this call was redundant. Removed in `963458aa6`.

steel-core/src/command/builtins/invsee/tests.rs
kdcokenny: Removed the redundant registry initialization here as well. I kept the `test_player` helper because it has 18 callers and centralizes the shared world and two-player test configuration.

steel-core/src/entity/entities/mobs/passive/cow/tests/mod.rs
kdcokenny: Restored the touched cow test files to CRLF in `86144a037`. PR #363 can handle their normalization separately.

coco875 (dismissed):

JunkyDeveloper (dismissed):

coco875:
test fail x)

JunkyDeveloper (approved):

coco875 (approved):

PR #370, opened 2026-08-15
fix: Prevent ghost players after stalled disconnects

### LINKED ISSUE #368: Timed-out connections can leave ghost players and block reconnects
## Description

A timed-out client can remain registered as an online player indefinitely when its TCP connection is backpressured. Subsequent connections using the same UUID are rejected with:

> You are already connected to this server

This was reproduced against Steel 0.15.2+mc26.2.

## Reproduction

1. Join a Steel server normally.
2. Allow the server to queue enough chunk data to fill the connection's TCP send buffer.
3. Stop reading packets and responding to keepalives, simulating a frozen client or broken connection.
4. Wait for Steel's keepalive timeout.
5. Attempt to reconnect using the same UUID.

A protocol client requesting 500 chunks per tick was used to reproduce the backpressure reliably.

## Actual behavior

- Steel triggers the keepalive timeout after approximately 15 seconds.
- The player does not leave the world and remains in the online-player index.
- The connection's send queue remains full. During reproduction it contained approximately 985 KB of unsent data and was receive-window-limited 99.8% of the time.
- Repeated connections with the same UUID complete login/configuration, then receive a play-state duplicate-login disconnect.
- The player is removed only after the original socket is externally closed. Once closed, Steel immediately finishes cleanup.

## Expected behavior

- Player removal and persistence should proceed even if the final disconnect packet cannot be delivered.
- Following vanilla behavior, a duplicate login should disconnect the old session and wait for its removal before admitting the new session.
- A stalled transport should not retain a player indefinitely.

## Cause

`JavaConnection::disconnect` queues a disconnect packet and cancels the connection.

The sender observes cancellation and awaits `write_queued_disconnect()`. Player removal is queued only after that write and `release_network_writer()` finish:

```rust
self.write_queued_disconnect(&mut sender_recv).await;
// ...
player.server().queue_player_disconnect(player);
```

The underlying `write_all` and `flush` have no timeout. If the peer is still connected but no longer reading, the final write can remain pending indefinitely, so `queue_player_disconnect` is never reached.

Relevant code:

- `steel-core/src/player/connection/java.rs`
- `steel-protocol/src/packet_writer.rs`
- `steel-core/src/server/player_admission.rs`
- `steel-login/src/handlers/login.rs`

## Vanilla and Paper comparison

Vanilla 26.2:

- Installs a 30-second transport-level `ReadTimeoutHandler`.
- Detects duplicates during login.
- Disconnects all existing players with the profile.
- Holds the new connection in `WAITING_FOR_DUPE_DISCONNECT` until the old player is removed.
- Applies a 600-tick slow-login deadline.

Paper retains this flow, adds duplicate-name matching for proxy online mode, and associates a dedicated duplicate-login disconnect cause.

Steel currently checks for duplicates only during world admission and rejects the new connection. The login handler still contains:

```rust
//TODO: Check for duplicate player UUID or name
```

## Suggested resolution

1. Decouple authoritative player cleanup from successful disconnect-packet delivery.
2. Make the final disconnect flush best-effort and bounded.
3. Add a transport inactivity timeout covering login, configuration, and play.
4. Implement vanilla's duplicate-session eviction-and-wait behavior.
5. Add integration coverage using a backpressured socket.

kdcokenny:
## Type of change

- [x] Bug fix

## Description

Steel queued player cleanup only after writing the final disconnect packet. If that write stalled, the player remained registered online and reconnect attempts were rejected.

Cleanup now happens before bounded best-effort socket teardown. Duplicate logins also disconnect the existing session, wait for its persistence lifecycle to finish, and atomically reserve the replacement join.

Closes #368.

## How this was tested

Minecraft 26.2 on Linux.

- Held the Java network writer and verified the player leaves the online and world indexes while the sender remains blocked.
- Verified a duplicate login evicts a relocating player with the vanilla reason and waits for disconnect admission to release.
- Verified concurrent replacements cannot reserve the same UUID.
- `cargo test -p steel-core --lib` (2,357 passed, 1 ignored)
- `cargo test -p steel-login` (16 passed)
- `cargo clippy -p steel-core -p steel-login --all-targets --all-features -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`

## Screenshots / logs

N/A

## Checklist

- [x] Self-reviewed the diff

## Additional notes

Broad transport inactivity timeouts and platform-dependent TCP backpressure tests are intentionally left for separate follow-up work.


steel-login/src/handlers/login.rs
```diff
@@ -17,6 +17,40 @@ use crate::{
 };
 
 impl JavaTcpClient {
+    async fn disconnect_duplicate_player(&self, profile: &GameProfile) -> bool {
+        match self
+            .server
+            .disconnect_duplicate_player_and_wait(profile.id, &self.cancel_token)
```
Joniii11: maybe add todo here vanilla has 600 tick login timeout while waiting for dupe disconnect see ServerLoginPacketListenerImpl::tick()

steel-core/src/player/connection/java.rs
```diff
@@ -546,6 +591,7 @@ impl JavaConnection {
     /// Closes the connection.
     pub fn close(&self) {
         self.cancel_token.cancel();
+        self.queue_player_disconnect_once();
```
Joniii11: vanilla connection disconnect always goes through ServerGamePacketListenerImpl::onDisconnect() ---> PlayerList::remove() maybe cleanup should be guaranteed for every NetworkConnection too not only JavaConnection::close()

steel-login/src/handlers/login.rs
kdcokenny: decided against a todo, bc that would ultimately be me introducing a bug through this pr. it's good now! (at least should be lol)

Joniii11 (approved):
lgtm

coco875 (approved):

JunkyDeveloper (approved):

PR #372, opened 2026-08-16
feat(command): implement `/setidletimeout` command and the idle-kick system

SomeYellowGuy:
## Type of change

- [x] Command implementation
- [x] New feature

## Description

This PR adds the `/setidletimeout` command, and, along with it, the system to kick players for being idle for too long (depending on the server's player idle timeout value).

## How this was tested

I temporarily set the idle timeout duration to be ~3 seconds, and I tested every hook that could be reproduced (like a player moving, leaving a bed, updating a sign, etc.), which all delay the idle kick. Of course, I also tested the command itself, and it works properly.

## Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/steel-core/src/command/builtins/setidletimeout.rs
+++ b/steel-core/src/command/builtins/setidletimeout.rs
@@ -5,6 +5,7 @@ use crate::command::execution::{
     CommandSource, SteelCommandContext, SteelCommandRuntime, argument, literal,
 };
 use crate::command::registration::CommandRegistration;
+use std::sync::atomic::Ordering;
 use steel_utils::Identifier;
 use steel_utils::translations::{
     COMMANDS_SETIDLETIMEOUT_SUCCESS, COMMANDS_SETIDLETIMEOUT_SUCCESS_DISABLED,
@@ -25,10 +26,11 @@ fn set_idle_timeout(
 ) -> Result<i32, CommandSyntaxError> {
     let time = context.integer("minutes")?;
 
-    {
-        let mut guard = context.source().server().player_idle_timeout.lock();
-        *guard = time;
-    }
+    context
+        .source()
+        .server()
+        .player_idle_timeout
+        .store(time, Ordering::Relaxed);
 
     if time > 0 {
         context.source().send_success(
--- a/steel-core/src/command/builtins/setidletimeout.rs
+++ b/steel-core/src/command/builtins/setidletimeout.rs
@@ -1,0 +1,38 @@
+//! The vanilla `/setidletimeout` command.
+
+use text_components::TextComponent;
+use steel_utils::Identifier;
+use steel_utils::translations::{COMMANDS_SETIDLETIMEOUT_SUCCESS, COMMANDS_SETIDLETIMEOUT_SUCCESS_DISABLED};
+use crate::command::brigadier::{ArgumentType, CommandNodeBuilder, CommandSyntaxError};
+use crate::command::execution::{argument, literal, CommandSource, SteelCommandContext, SteelCommandRuntime};
+use crate::command::registration::CommandRegistration;
+
+pub(super) fn registration() -> CommandRegistration<CommandSource> {
+    CommandRegistration::new(Identifier::vanilla_static("setidletimeout"), |_| command())
+}
+
+fn command() -> CommandNodeBuilder<CommandSource, SteelCommandRuntime> {
+    literal("setidletimeout").then(
+        argument("minutes", ArgumentType::integer(0, i32::MAX)).executes(set_idle_timeout),
+    )
+}
+
+fn set_idle_timeout(context: &SteelCommandContext<CommandSource>) -> Result<i32, CommandSyntaxError> {
+    let Some(time) = context.integer("minutes") else {
+        return Err(missing_argument("pos"));
+    };
+    context.source().server().player_idle_timeout = time;
+
+    if time > 0 {
+        context.source().send_success(&COMMANDS_SETIDLETIMEOUT_SUCCESS.message([TextComponent::plain(format!("{}", time))]).component(), true);
+    } else {
+        context.source().send_success(&TextComponent::from(&COMMANDS_SETIDLETIMEOUT_SUCCESS_DISABLED), true);
+    }
+
+    Ok(1)
+}
+fn missing_argument(name: &str) -> CommandSyntaxError {
+    CommandSyntaxError::dynamic(format!(
+        "Parsed value for {name} is missing from the command context"
+    ))
+}
--- a/steel-core/src/command/builtins/setidletimeout.rs
+++ b/steel-core/src/command/builtins/setidletimeout.rs
@@ -46,5 +46,5 @@ fn set_idle_timeout(
         );
     }
 
-    Ok(1)
+    Ok(time)
 }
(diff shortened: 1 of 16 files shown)


steel-core/src/command/builtins/setidletimeout.rs
```diff
@@ -0,0 +1,50 @@
...
+    context
+        .source()
+        .server()
+        .player_idle_timeout
+        .store(time, Ordering::Relaxed);
+
+    if time > 0 {
+        context.source().send_success(
+            &COMMANDS_SETIDLETIMEOUT_SUCCESS
+                .message([TextComponent::plain(format!("{time}"))])
```
Laptop59: Why use `format!` when `time.to_string()` should do the same?

steel-core/src/player/mod.rs
```diff
@@ -253,6 +254,9 @@ pub struct Player {
     /// In-flight ender pearls thrown by this player, kept weakly so they persist
     /// with the player and re-spawn on login (vanilla `ServerPlayer.enderPearls`).
     ender_pearls: SyncMutex<Vec<Weak<dyn Entity>>>,
+
+    /// The last action time of this player.
+    last_action_time: SyncMutex<Instant>,
```
coco875, line 265: Instant aren't just an int or a float ? if it's the case then an atomic wouldn't be better ?

steel-core/src/server/mod.rs
```diff
@@ -415,6 +415,9 @@ pub struct Server {
     player_admissions: SyncMutex<FxHashMap<Uuid, PlayerAdmissionState>>,
     /// The tick rate manager for the server.
     pub tick_rate_manager: SyncRwLock<TickRateManager>,
+    /// The number of minutes required for a player to be idle for them to be kicked (timed out) from the server.
+    /// If this is equal to 0, no kicking will happen.
+    pub player_idle_timeout: SyncMutex<i32>,
```
coco875: why not use atomic ?

coco875 (changes requested):

steel-core/src/server/mod.rs
SomeYellowGuy: I initially used atomics, but I later removed them because I barely saw any code with atomics. Is that a problem?

coco875: normally it depend of kind of operation, like we only read or write, I don't think we need a mutex because we don't want to avoid lock any other modification for x time

steel-core/src/player/mod.rs
SomeYellowGuy: > Instant aren't just an int or a float ? if it's the case then an atomic wouldn't be better ?

An `Instant` has an `i64` for the seconds value and a `u32` for the nanosecond value. I don't think that fits in a single atomic (at least one of a size of 64 bits).

coco875: ok thanks, then like that it's fine

coco875 (dismissed):

steel-core/src/command/builtins/setidletimeout.rs
```diff
@@ -0,0 +1,50 @@
...
+            true,
+        );
+    } else {
+        context.source().send_success(
+            &TextComponent::from(&COMMANDS_SETIDLETIMEOUT_SUCCESS_DISABLED),
+            true,
+        );
+    }
+
+    Ok(1)
```
JunkyDeveloper: vanilla returns here time and not 1, so please explain please

JunkyDeveloper (changes requested):
Please add also the config settings so that this feature would be fully finished

JunkyDeveloper (changes requested):
These are claude findings, which I checked my self and confirmed:
> CRITICAL steel-core/src/player/game_mode/block_interaction.rs:24 (swing(), dispatched from steel-core/src/player/connection/java.rs:313) vs ServerGamePacketListenerImpl.java:1708-1709 (handleAnimate) — arm-swing packet never calls reset_last_action_time(). Vanilla resets idle timer on every swing (movement.lengthSqr()-independent, unconditional). A player who only swings their arm gets idle-kicked in Steel but not vanilla. Fix: call self.reset_last_action_time() in swing().

> CRITICAL steel-core/src/player/connection/java.rs:266-274 (ScheduledPlayPacketKind::ChatCommand branch) vs ServerGamePacketListenerImpl.java:1635-1636 (handleSignedChatCommand) — running a slash command via chat never resets the idle timer. A player repeatedly issuing commands (no movement/interaction) gets idle-kicked in Steel but not vanilla. Fix: add player.reset_last_action_time() before/after submit_command.

It depends how easy you can add things maybe create for each an own handle function

SomeYellowGuy:
> These are claude findings, which I checked my self and confirmed:
> 
> > CRITICAL steel-core/src/player/game_mode/block_interaction.rs:24 (swing(), dispatched from steel-core/src/player/connection/java.rs:313) vs ServerGamePacketListenerImpl.java:1708-1709 (handleAnimate) — arm-swing packet never calls reset_last_action_time(). Vanilla resets idle timer on every swing (movement.lengthSqr()-independent, unconditional). A player who only swings their arm gets idle-kicked in Steel but not vanilla. Fix: call self.reset_last_action_time() in swing().
> 
> > CRITICAL steel-core/src/player/connection/java.rs:266-274 (ScheduledPlayPacketKind::ChatCommand branch) vs ServerGamePacketListenerImpl.java:1635-1636 (handleSignedChatCommand) — running a slash command via chat never resets the idle timer. A player repeatedly issuing commands (no movement/interaction) gets idle-kicked in Steel but not vanilla. Fix: add player.reset_last_action_time() before/after submit_command.
> 
> It depends how easy you can add things maybe create for each an own handle function

1. `swing()` (that Claude says) does not even call `resetLastActionTime()` in Minecraft. There is a different handler, `handleAnimate()`, that actually does call this function, but I don't think it exists in Steel yet. Should I implement this handler?

2. Currently, Steel only has the handler for `handleChat` (not the other two), and I don't understand where to place the function. The handler's code is so unreadable for me (I can't cross-check Steel's and Minecraft's method). Also, Claude talks about `handleSignedChatCommand`, which isn't correct if we are talking about `ScheduledPlayPacketKind::ChatCommand`.

JunkyDeveloper:
> 
>     1. `swing()` (that Claude says) does not even call `resetLastActionTime()` in Minecraft. There is a different handler, `handleAnimate()`, that actually does call this function, but I don't think it exists in Steel yet. Should I implement this handler?
> 
>     2. Currently, Steel only has the handler for `handleChat` (not the other two), and I don't understand where to place the function. The handler's code is so unreadable for me (I can't cross-check Steel's and Minecraft's method). Also, Claude talks about `handleSignedChatCommand`, which isn't correct if we are talking about `ScheduledPlayPacketKind::ChatCommand`.

Need to check again, because I found these things at steel

JunkyDeveloper:
So some more information to swing
image
that is what we have and this is the same package, so we already call it and yes the problem is it doesn't call the reset and I tested it, if I attack, I get kicked. So add before player.swing() the reset, this should fix it I think

SomeYellowGuy:
I think I accidentally looked at the clientbound packet for swinging, so I think you're right. However, I still don't know what to do for the chat handler.

coco875 (dismissed):

JunkyDeveloper (approved):

coco875 (approved):

JunkyDeveloper:
> I think I accidentally looked at the clientbound packet for swinging, so I think you're right. However, I still don't know what to do for the chat handler.

added the reset two the missing places at the newest commit from my side, it was "that" simple

PR #377, opened 2026-08-19
allow simulation_distance of 0 in config schema

### LINKED ISSUE #345: Allow zero simulation distance in the packaged schema
I noticed Steel supports `simulation_distance = 0`, but the packaged configuration schema requires a minimum of `1`.

At zero, chunks still load normally, but only the player’s current chunk receives entity updates. This can reduce server work, with the tradeoff that nearby mobs and other activity may freeze across chunk borders and natural mob spawning may be reduced.

Zero should remain an intentional option for server owners. The default should remain `10`, matching Vanilla.

- [ ] Change the packaged schema minimum for `simulation_distance` from `1` to `0`.
- [ ] Keep the default at `10`.
- [ ] Update the schema description to explain the effect of low values.
- [ ] Add a focused configuration test proving `simulation_distance = 0` is accepted.

Research:

- [Steel runtime validation already accepts zero](https://github.com/Steel-Foundation/SteelMC/blob/0708f9ccfc7d2b82a181a1dba23e2faa3fab14fd/steel/src/config/server.rs#L148-L150)
- [Steel’s packaged schema currently requires a minimum of 1 and defaults to 10](https://github.com/Steel-Foundation/SteelMC/blob/0708f9ccfc7d2b82a181a1dba23e2faa3fab14fd/package-content/config.schema.json#L37-L42)
- [Steel uses the configured value as the player ticket’s simulation radius](https://github.com/Steel-Foundation/SteelMC/blob/0708f9ccfc7d2b82a181a1dba23e2faa3fab14fd/steel-core/src/chunk/chunk_ticket_manager.rs#L191-L209)
- [Vanilla 26.2 accepts an unrestricted integer and defaults to 10](https://github.com/Renekovski/26.2-mcp/blob/51f3128ba7265299dce20b9abd2b37d7dd27d096/src/net/minecraft/server/dedicated/DedicatedServerProperties.java#L102-L104)
- [Paper’s public world API separately restricts programmatic changes to `[2, 32]`](https://github.com/PaperMC/Paper/blob/c9e894d3cc03f21f80de4f4061a795e11941e89a/paper-server/patches/sources/io/papermc/paper/FeatureHooks.java.patch#L212-L217)

3ch0less:
Steel already accepts simulation_distance = 0 at runtime, but the packaged schema required a minimum of 1. Drop the minimum to 0, update descriptions, and add a validation test.

Fixes #345

### DIFF
--- a/package-content/config.schema.json
+++ b/package-content/config.schema.json
@@ -36,8 +36,8 @@
         },
         "simulation_distance": {
           "type": "integer",
-          "description": "Maximum simulation distance in chunks",
-          "minimum": 1,
+          "description": "Simulation distance in chunks. Lower values reduce server work by restricting entity updates to fewer chunks around the player. At 0, only the player's current chunk receives entity updates.",
+          "minimum": 0,
           "maximum": 128,
           "default": 10
         },
--- a/package-content/config.toml
+++ b/package-content/config.toml
@@ -10,7 +10,9 @@ max_players = 20
 allow_extended_view_distance = false
 # Maximum view distance in chunks. Normally 1-32; 1-128 when allow_extended_view_distance is true.
 view_distance = 10
-# Maximum simulation distance in chunks
+# Simulation distance in chunks. Lower values reduce server work by restricting entity
+# updates to fewer chunks around the player. At 0, only the player's current chunk
+# receives entity updates.
 simulation_distance = 10
 # Maximum chained neighbor-update tasks. Negative values disable the limit.
 max_chained_neighbor_updates = 1000000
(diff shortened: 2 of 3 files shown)


coco875:
by memory @JunkyDeveloper have suggest to emit a warning when it's below 4 because it can be not correct manip or alterate behavior (like hostile mob that spawn at 15 blocks from the player)

coco875:
I ping him to make sure he still want it or not

3ch0less:
Sounds good, waiting for their feedback

JunkyDeveloper:
yes it should be a warning, because it can come to not vanilla behaviour!

coco875:
any update ?

JunkyDeveloper:
no activity and not even validation added

PR #379, opened 2026-08-19
fix: clear snowy state when snow above is removed (#378)

### LINKED ISSUE #378: Grass remains in snowy state after snow is removed
Breaking the last snow layer above a grass block leaves the block stuck at `snowy=true`. With the snow gone, the client exposes the untinted `grass_block_snow` top. That is the gray surface shown below.

Grass block remains snowy after the snow layer is broken

## Reproduce

1. Find or place a grass block with a snow layer above it.
2. Break the snow layer.
3. Inspect the exposed grass block.

After step 2, `snowy` should become `false` and the grass should use its biome-tinted texture. It stays `true` instead.

## Cause

The break path already requests full neighbor updates, and the shape update reaches the block below. The stale state comes from missing block behaviors:

- World generation sets `SNOWY=true` below newly placed snow.
- The extracted mappings name `GrassBlock`, `SnowyBlock`, and `MyceliumBlock`, but Steel does not implement those behaviors.
- Without them, `DefaultBlockBehavior::update_shape` returns the old state unchanged.

Minecraft 26.2 recalculates `SNOWY` in `SnowyBlock::updateShape` whenever the block above changes. It checks whether the new block belongs to `#minecraft:snow`.

Grass, podzol, and mycelium all need this behavior. It must work when snow is placed or removed, including snow layers, snow blocks, and powder snow.

## Fix

Add the shared `SnowyBlock` placement and upward-neighbor logic, then use it for grass, podzol, and mycelium. Add tests that cover setting and clearing `SNOWY`.

kartheesan05:
## Type of change

- [x] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

Fixes #378.

When the last snow layer above grass/podzol/mycelium is broken, those blocks keep `snowy=true`. The client then draws the untinted `grass_block_snow` top instead of the biome-tinted grass top.

The break path already sends a full shape update to neighbors. The block below receives `update_shape(..., Direction::Up, ..., new_above_state)`. That is the same hook vanilla uses.

Vanilla `SnowyBlock` does this:

- `updateShape`: if the changed neighbor is **up**, set `SNOWY` to `aboveState.is(BlockTags.SNOW)`; otherwise leave the state alone
- `getStateForPlacement`: set `SNOWY` from the block at `clickedPos.above()`

Steel already maps:

- `grass_block` → `GrassBlock`
- `podzol` → `SnowyBlock`
- `mycelium` → `MyceliumBlock`

Those classes were not implemented, so they fell through to `DefaultBlockBehavior::update_shape`, which returns the old state. Worldgen (`freeze_top_layer`) can still write `SNOWY=true`, so after snow is removed the property never flips back.

This PR adds the missing behaviors:

- `SnowyBlock` implements the vanilla placement + upward-neighbor `SNOWY` logic with `BlockTag::SNOW` (`snow`, `snow_block`, `powder_snow`)
- `GrassBlock` and `MyceliumBlock` are thin wrappers that delegate those two methods. The behavior generator matches extracted class names, so all three structs need `#[block_behavior]`

Not in this PR: `SpreadingSnowyBlock` random ticks, grass spreading, or bonemeal. Vanilla `GrassBlock` / `MyceliumBlock` do not override `updateShape` / `getStateForPlacement`; they inherit them from `SnowyBlock`, so this is enough to fix the stale `snowy` state.

## How this was tested

Environment: MC 26.2 (`0.15.2+mc26.2`), macOS, nightly Rust 2026-07-23

```bash
RUSTFLAGS="-Dwarnings" cargo test -p steel-core snowy --verbose
RUSTFLAGS="-Dwarnings" cargo test -p steel-core --verbose
RUSTFLAGS="-Dwarnings" cargo clippy -p steel-core -r --all-targets --all-features
cargo fmt --all --check
```

Unit tests in `snowy_block.rs`:

- `BlockTag::SNOW` members (snow layer, snow block, powder snow)
- upward updates set and clear `SNOWY`
- non-upward updates keep `SNOWY`
- placement with and without snow above, for grass, podzol, and mycelium

Manual check:

1. Place grass with a snow layer on top.
2. Break the snow layer.
3. Grass top should tint with the biome (`snowy=false`).

## Screenshots / logs

N/A

## Checklist

- [x] Self-reviewed the diff

## Additional notes

No edits to generated Rust, `classes.json`, or extracted JSON.


coco875:
your pr clearly add Block implementation

coco875:
I would like that you read more about what you do, to avoid doing pr where you don't know how it fix a bug

AshuraV1:
> I would like that you read more about what you do, to avoid doing pr where you don't know how it fix a bug

Do i make my pr?

kartheesan05:
You're right on the template — this is a block implementation as well as a bug fix. I updated the type checkboxes.

On the bug itself: the neighbor-update path was already firing. Breaking snow above grass calls `update_shape` on the grass with `direction == Up` and the new above state (air). Vanilla `SnowyBlock` then sets `SNOWY` from `#minecraft:snow`. Steel had no `SnowyBlock` / `GrassBlock` / `MyceliumBlock` behaviors, so `DefaultBlockBehavior` returned the old state and `snowy` stayed `true`.

This PR only ports that placement + upward-neighbor logic. Spreading and bonemeal are still out of scope.

coco875:
close this pr, in favor of https://github.com/Steel-Foundation/SteelMC/pull/380

PR #380, opened 2026-08-19
fix: update snowy state on grass, podzol, and mycelium when snow is removed (#378)

### LINKED ISSUE #378: Grass remains in snowy state after snow is removed

AshuraV1:
## Type of change

- [x] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

Resolves #378 by adding behaviors for `SnowyBlock`, `GrassBlock`, and `MyceliumBlock` in `steel-core`

Earlier removing snow from grass, podzol or mycelium blocks would leave those blocks stuck with `snowy=true` due to fallback shape update behavior.
### Changes
- Added `SnowyBlock` behavior to `steel-core/src/behavior/blocks/vegetation/snowy_block.rs` that recalculates `snowy` state while placing blocks and on upward shape update with `#minecraft:snow` (`BlockTag::SNOW`).
- Added `GrassBlock` and `MyceliumBlock` behaviors that rely on `SnowyBlock` for placement and shape updates.
- Added unit tests for placement and setting/clearing of `snowy` state in upward direction for all three block types.

## Checklist

- [x] Self-reviewed the diff


steel-core/src/behavior/blocks/vegetation/snowy_block.rs
coco875, line 1: split each implementation in their own file instead have snowy block that also contain mycelium

coco875 (commented):
look like there is todo that are missing no ? for bonemeal, that it become dirt when you place a block above etc no ?

AshuraV1:
> look like there is todo that are missing no ? for bonemeal, that it become dirt when you place a block above etc no ?

yea i forgot about em
I've added concise TODO comments in GrassBlock and MyceliumBlock for these features
that should be it!

steel-core/src/behavior/blocks/vegetation/mycelium_block.rs
```diff
@@ -8,6 +8,7 @@ use crate::behavior::context::BlockPlaceContext;
 use crate::world::ScheduledTickAccess;
 
 /// Behavior for mycelium blocks.
+// TODO: Implement SpreadingSnowyBlock random ticks (spreading, turning to dirt when covered) and client particle effects.
```
coco875: client particle effects, aren't by definition only on client ?

steel-core/src/behavior/blocks/vegetation/snowy_block.rs
```diff
@@ -0,0 +1,237 @@
...
+            is_snowy_setting(neighbor_state),
+        )
+    } else {
+        state
+    }
+}
+
+/// Behavior for podzol and other basic snowy dirt blocks.
+#[block_behavior]
+pub struct SnowyBlock {
```
coco875, line 46: the snowyblock are complete ?

AshuraV1: Thats it for the SnowyBlock in vanilla SnowyBlock.java SnowyBlock only defines getStateForPlacement, updateShape ( for Direction.UP ) and isSnowySetting (#minecraft:snow)
Podzol uses this directly and has no other mechanics

coco875 (approved):

JunkyDeveloper (approved):

PR #381, opened 2026-08-19
Fix damage command attacker attribution

### LINKED ISSUE #346: Damage command loses attacker attribution
I noticed `/damage <target> <amount> <damageType> by <entity>` does not record the entity that caused the damage.

I reproduced this in the client on a Steel server:

- Killing a player with a diamond sword recorded the attacker for 100 ticks.
- Killing a player with `/damage CommandVictim 100 minecraft:player_attack by Attacker` succeeded, but recorded no attacker.

Vanilla treats `by <entity>` as both the direct and causing entity. Paper inherits the same behavior.

- [ ] Record `by <entity>` as both the direct and causing entity.
- [ ] Keep `from <cause>` as an override for the causing entity.
- [ ] Add a regression test proving command damage preserves player attribution.

Research:

- [Steel only records the direct entity](https://github.com/Steel-Foundation/SteelMC/blob/0708f9ccfc7d2b82a181a1dba23e2faa3fab14fd/steel-core/src/command/builtins/damage.rs#L67-L78)
- [Vanilla damage command behavior](https://github.com/Renekovski/26.2-mcp/blob/51f3128ba7265299dce20b9abd2b37d7dd27d096/src/net/minecraft/server/commands/DamageCommand.java#L61-L93)
- [Vanilla damage-source constructors](https://github.com/Renekovski/26.2-mcp/blob/51f3128ba7265299dce20b9abd2b37d7dd27d096/src/net/minecraft/world/damagesource/DamageSource.java#L43-L56)
- [Paper does not patch the damage command](https://github.com/PaperMC/Paper/tree/c9e894d3cc03f21f80de4f4061a795e11941e89a/paper-server/patches/sources/net/minecraft/server/commands)

genkii:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

Fixes `/damage <target> <amount> <damageType> by <entity>` not recording the by entity as the causing entity.
Now the `by` entity is recorded as both the direct and causing entity, matching vanilla behavior. causing entity will be overridden if `from <cause>` is specified.

Closes #346 

## How this was tested

Verified with:

`/damage Victim 1 minecraft:player_attack by Attacker`

Verified that `Attacker` is recorded as the player responsible for the damage.

Added one regression test:
  `damage_command_records_by_entity_as_the_responsible_player`

## Screenshots / logs

Not applicable.

## Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/steel-core/src/command/builtins/damage.rs
+++ b/steel-core/src/command/builtins/damage.rs
@@ -65,7 +65,10 @@ fn damage(context: &SteelCommandContext<CommandSource>) -> Result<i32, CommandSy
 
     // Else, it's from the "by", or maybe it's nothing
     if let Ok(entity) = context.entity("entity") {
-        damage_source.direct_entity_id = Some(entity.id());
+        let entity_id = entity.id();
+
+        damage_source.direct_entity_id = Some(entity_id);
+        damage_source.causing_entity_id = Some(entity_id);
 
         // Maybe even the causing entity is known
         if let Ok(cause) = context.entity("cause") {
--- a/steel-core/src/server/tests.rs
+++ b/steel-core/src/server/tests.rs
@@ -29,8 +29,8 @@ use uuid::Uuid;
 
 use crate::behavior::init_behaviors;
 use crate::command::execution::{
-    CommandArgumentSource, CommandPermissionSource, CommandSource, ExecutionCommandSource,
-    parse_entity_selector_text,
+    CommandArgumentSource, CommandExecutionContext, CommandPermissionSource, CommandResultCallback,
+    CommandSource, ExecutionCommandSource, ExecutionStop, parse_entity_selector_text,
 };
 use crate::command::sender::{CommandExecutionOwner, CommandSender};
 use crate::config::{ResolvedDomainConfig, RuntimeConfig, StorageSelection};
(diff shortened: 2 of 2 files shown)


steel-core/src/command/builtins/damage.rs
```diff
@@ -65,7 +65,10 @@ fn damage(context: &SteelCommandContext<CommandSource>) -> Result<i32, CommandSy
 
     // Else, it's from the "by", or maybe it's nothing
     if let Ok(entity) = context.entity("entity") {
-        damage_source.direct_entity_id = Some(entity.id());
+        let entity_id = entity.id();
+
+        damage_source.direct_entity_id = Some(entity_id);
+        damage_source.causing_entity_id = Some(entity_id);
```
coco875, line 71: why you put the same on both ?

genkii: Because in the vanilla code using `by` means that the entity is both the causing and direct entity previously only the direct entity was set which refering to issue #346 was incorrect: https://github.com/Steel-Foundation/SteelMC/blob/0708f9ccfc7d2b82a181a1dba23e2faa3fab14fd/steel-core/src/command/builtins/damage.rs#L67-L78

coco875 (approved):

JunkyDeveloper (approved):

PR #383, opened 2026-08-19
feat: Support capes and outer skin layers

kdcokenny:
## Type of change

- [x] Bug fix
- [x] New feature

## Description

Syncs player main hand and model-part visibility through Avatar metadata on join and later client-information updates. Hat changes also update the tab-list overlay. Initial player-info packets no longer hardcode the hat as visible.

The packet fields changed from VarInt-backed `i32` values to `i8` and `u8` because Vanilla 26.2 reads view distance with `readByte` and model customization with `readUnsignedByte`. Values below `0x80` made the old decoder appear correct. A value such as `0xff` is a VarInt continuation and causes the decoder to consume the following setting. This is primarily a wire-format fix, though `u8` is also smaller than `i32`.

Entity metadata stores the customization byte as `i8`. `cast_signed()` preserves the same eight bits when converting from the packet's `u8`; for example, `0xff` becomes `-1`.

The hat is bit 6, or `0b0100_0000`. World-model layers use entity metadata, while `UPDATE_HAT` separately controls the tab-list face overlay.

`player_mode_customisation` comes from extracted Vanilla Java data and is not sent over the network. The generator maps that exact identifier to the US-English `player_mode_customization`. The `_typos.toml` entry only permits the extracted input spelling.

The `0xff` test value exercises the high bit that exposed the VarInt bug. `0b0001_0101` enables cape, left sleeve, and left pants leg to verify that separated bits survive unchanged.

These changes are kept in one PR because packet decoding, metadata projection, and player-info updates form one functional path.

## How this was tested

Minecraft 26.2 on Linux.

```sh
cargo test
cargo clippy -r --all-targets --all-features
cargo fmt --all --check
typos
```

A Fabric client GameTest was run under Xvfb against a live Steel server. It verified:

- all seven model parts enabled
- all parts disabled with left main hand
- cape, jacket, and hat enabled with right main hand
- matching tab-list hat state

The same behavior was also verified manually in-game against a live Steel server. Cape and outer-layer visibility, main-hand changes, and the tab-list hat overlay all updated correctly.

## Screenshots / logs

```text
[steel-cosmetics] passed; local cape texture present=false
BUILD SUCCESSFUL
```

## Checklist

- [x] Self-reviewed the diff

## Additional notes

Cape texture fetching, rendering, and animation remain client-side. Steel already forwards Mojang-signed `textures` properties, with regression coverage in this PR.

The Fabric test used an offline profile, so it could verify the cape visibility bit but could not receive an official Mojang-signed cape texture.


chocodev11:
LGTM

Joniii11 (approved):
LGTM

steel-core/src/player/mod.rs
```diff
@@ -133,6 +133,7 @@ use steel_utils::{
 use crate::inventory::container::Container;
 
 const RESPAWN_SEARCH_READY_CANDIDATE_BUDGET: usize = 8;
+const HAT_MODEL_PART_MASK: i8 = 0b0100_0000;
```
coco875, line 136: like there is 8 bit, does hall have a meaning ?

coco875:
does the pr consist of repeat to everyone what a skin have ?

steel-core/src/player/mod.rs
chocodev11: cape, jacket, left/right sleeve, left/right pants leg, and hat

kdcokenny: Vanilla uses bits 0 through 6 for cape, jacket, sleeves, pants legs, and hat. Bit 7 is unused.

kdcokenny:
> does the pr consist of repeat to everyone what a skin have ?

No skin texture is rebroadcast. The model-part mask goes only to clients tracking that player. Hat visibility is broadcast globally because it affects the tab list. Signed skin/cape profile properties were already sent during player-info admission.

steel-core/src/player/mod.rs
coco875: and did we already read other ? in this case an enum woudln't be better ?

kdcokenny: Only the hat is read server-side; clients handle the other bits. Multiple parts can be enabled at once, so the value remains a bitmask.

coco875: oh ok, a bit weird that they not make it also client side the read for hat

coco875 (dismissed):
I have try and confirmed that it work

coco875 (dismissed):

coco875 (approved):

teknostom (approved):
LGTM

Issue #387, opened 2026-08-20
Pick block ignores item components when matching inventory stacks


kdcokenny:
## Summary

Steel's pick-block handler searches the player's inventory for the cloned block item using item identity only. If the first matching item has different data components, Steel selects it instead of adding and selecting the clone stack.

For example, when the inventory contains a custom-named stone but no plain stone, pick-blocking plain stone selects the named stone.

## Vanilla behavior

Vanilla routes block picking through [`tryPickItem`](https://github.com/ohnodev/decompiled-minecraft-26-2-SNAPSHOT-2/blob/535cd83164af1680abe81dadd7fbe93ded4968a1/src/net/minecraft/server/network/ServerGamePacketListenerImpl.java#L738-L753), which calls `Inventory.findSlotMatchingItem`. That method uses [`ItemStack.isSameItemSameComponents`](https://github.com/ohnodev/decompiled-minecraft-26-2-SNAPSHOT-2/blob/535cd83164af1680abe81dadd7fbe93ded4968a1/src/net/minecraft/world/entity/player/Inventory.java#L134-L142), so a named stone does not match a plain stone.

Steel's [`handle_pick_item_from_block`](https://github.com/Steel-Foundation/SteelMC/blob/29452b529e5e96c52bf1593db91d3ae26c92c98a/steel-core/src/player/game_mode/block_interaction.rs#L193-L214) calls [`find_slot_matching_item`](https://github.com/Steel-Foundation/SteelMC/blob/29452b529e5e96c52bf1593db91d3ae26c92c98a/steel-core/src/player/player_inventory/core.rs#L226-L235), which uses `ItemStack::is_same_item`. Steel already has the component-sensitive [`ItemStack::is_same_item_same_components`](https://github.com/Steel-Foundation/SteelMC/blob/29452b529e5e96c52bf1593db91d3ae26c92c98a/steel-registry/src/item_stack.rs#L322-L332) helper.

## Verification

Confirmed in game with a Fabric Client GameTest against vanilla 26.2 and Steel. The client used `MultiPlayerGameMode.handlePickItemFromBlock(target, false)`, the same packet path used by ordinary middle-click.

The test started in Creative mode with one stone named `PickProbe`, no plain stone, and an empty hotbar slot selected. It then pick-blocked a plain stone:

- Vanilla selected slot 1 with a new plain stone. The inventory contained one named stone and one plain stone.
[...]

PR #389, opened 2026-08-20
fix(pick block)

ezraclintoc:
Picking blockes with middle mouse now does not pick renamed blocks and doesn't search for open slots, just overwrites current slot

## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

When middle clicking it will not fetch blocks from your inventory which are renamed, which is vanilla parity. Also will look for free hotbar slots to swap the item with. Also deprioritized the swapping of enchanted items.

## How this was tested

Get two stacks of items, rename one stack, middle click, and see which one you get. It should be the non-renamed one. 

## Checklist

- [x] Self-reviewed the diff


ezraclintoc:
#387

steel-core/src/player/game_mode/block_interaction.rs
```diff
@@ -202,12 +202,14 @@ impl Player {
...
         let mut inventory = self.inventory.lock();
 
-        let slot_with_item = inventory.find_slot_matching_item(&item_stack);
+        let slot_with_item = inventory.find_slot_matching_item_with_same_components(&item_stack);
 
         if slot_with_item != -1 {
             if PlayerInventory::is_hotbar_slot(slot_with_item as usize) {
                 inventory.set_selected_slot(slot_with_item as u8);
             } else {
+                // TODO: If current slot contains item it should pick another slot
```
banocean: To be clear, is this to do for this pr or added as missing to do from earlier in this function?

ezraclintoc: I just submitted another PR that accomplishes this #390

coco875: I think if you want ezrclintoc you can fix the todo in the same pr to avoid making two pr

coco875:
yeah the two pr (https://github.com/Steel-Foundation/SteelMC/issues/390 and https://github.com/Steel-Foundation/SteelMC/issues/389) are very small and related to each other I think just combine it would better, what do you thing @ezraclintoc ?

ezraclintoc:
> yeah the two pr (#390 and #389) are very small and related to each other I think just combine it would better, what do you thing @ezraclintoc ?

I am open to that. I just don't have much knowledge on how to merge two PRs. I will try by asking AI, or if you know off the top of your head, just tell me what to run, or if you can run the commands yourself

coco875:
I have merge, can you update the name and decription ?

banocean:
> > yeah the two pr (#390 and #389) are very small and related to each other I think just combine it would better, what do you thing @ezraclintoc ?
> 
> I am open to that. I just don't have much knowledge on how to merge two PRs. I will try by asking AI, or if you know off the top of your head, just tell me what to run, or if you can run the commands yourself

Either one person can apply selected by them diffs (imo easiest way is to move them to stash and apply stash and commit) to their own branch or you make first pr point to 2nd pr's branch and marge it there like every other pr, that's how I got it done

coco875:
I have done it already in https://github.com/Steel-Foundation/SteelMC/pull/389/commits/c88867823f9e236d835073dfc567ed0f0ae0dcb1 x)

steel-core/src/player/player_inventory/core.rs
```diff
@@ -223,6 +223,18 @@ impl PlayerInventory {
         -1
     }
 
+    /// Finds the next empty slot after the selected in the inventory, or -1 if full.
+    #[must_use]
+    pub fn get_suitable_hotbar_slot(&self) -> u8 {
+        let selected = self.selected as usize;
+        let order = || (selected..Self::SELECTION_SIZE).chain(0..selected);
```
coco875: why make it a lamda function ? I think it should be fine because you borrow but not mut

coco875 (changes requested):

steel-core/src/player/player_inventory/core.rs
ezraclintoc: I don't understand

steel-core/src/player/player_inventory/core.rs
```diff
@@ -223,6 +223,18 @@ impl PlayerInventory {
...
+
+        order()
+            .find(|&i| self.items[i].is_empty())
+            .or_else(|| order().find(|&i| !self.items[i].is_enchanted()))
+            .map_or(selected as u8, |i| i as u8)
```
coco875: why not have done that ?
```suggestion
        let order = (selected..Self::SELECTION_SIZE).chain(0..selected);

        order
            .find(|&i| self.items[i].is_empty())
            .or_else(|| order.find(|&i| !self.items[i].is_enchanted()))
            .map_or(selected as u8, |i| i as u8)
```

ezraclintoc: yeah *I think that should work*

ezraclintoc: we need lamba because line 233 consumes order, and 234 also wants to consume order

coco875: hmm don't really like create a lambda function for an iterator, there is no way ?

coco875: another solution would be to make two explicit loop, I think it can be also more clear

banocean: > we need lamba because line 233 consumes order, and 234 also wants to consume order

won't that result in duplicating memory same as cloning but with additional overhead if you add lambda?

steel-core/src/player/player_inventory/core.rs
```diff
@@ -223,6 +223,26 @@ impl PlayerInventory {
...
+        {
+            return empty_slot as u8;
+        }
+        if let Some(non_enchanted_slot) = (selected..Self::SELECTION_SIZE)
+            .chain(0..selected)
+            .find(|&i| !self.items[i].is_enchanted())
+        {
+            return non_enchanted_slot as u8;
+        }
+        selected as u8
```
coco875: this wouldn't be simpler ? instead of try to do functional absolutely ?
```suggestion
        for slot in 0..Self::SELECTION_SIZE {
            let index = (selected + slot) % Self::SELECTION_SIZE;
            if self.items[index].is_empty() {
                return index as u8;
            }
        }

        for slot in 0..Self::SELECTION_SIZE {
            let index = (selected + slot) % Self::SELECTION_SIZE;
            if !self.items[index].is_enchanted() {
                return index as u8;
            }
        }

        self.selected
```

ezraclintoc:
ok, so I replaced the lamba with two iter loops.

I also found a race condition and fixed it.
If you were in creative, didn't have the block, and hotbar was full {
it would overwrite whever you are holding (which is vanilla), but it would also overwrite the item your holding even if the item was enchanted (which is not vanilla) }

coco875 (dismissed):

steel-core/src/player/player_inventory/core.rs
ezraclintoc: I guess its readability vs preformance, but I don't think the preformance is ANYTHING that you would ever notice.

ezraclintoc: Can we resolve this, or should we discuss which is better?

coco875: I don't think it matter at this scale so it's for code to be clean

ezraclintoc:
Should we also implement Ctrl+middle in this PR? I can work on that if we want

coco875:
hmm maybe in anothe one ? I would say it depend of the size of  it

ezraclintoc:
> hmm maybe in anothe one ? I would say it depend of the size of it

it might be helpful to wait until chests and barrels are implemented cuz that is the primary use case

coco875:
barrel are already implemented normally I think

steel-core/src/player/player_inventory/core.rs
```diff
@@ -223,6 +223,28 @@ impl PlayerInventory {
         -1
     }
 
+    /// Finds the next empty slot after the selected in the inventory, or -1 if full.
```
banocean: > or -1 if full

this comment seems to be incorrect (especially since that function returns u8 hah)

banocean (changes requested):

steel-core/src/player/player_inventory/core.rs
ezraclintoc: yep, I changed that function quite a bit and that comment is outdated, will resolve

ezraclintoc: my latest commit should resolve this

coco875 (dismissed):

steel-core/src/player/game_mode/block_interaction.rs
```diff
@@ -202,15 +202,21 @@ impl Player {
...
+                let slot = inventory.get_suitable_hotbar_slot();
+
+                inventory.set_selected_slot(slot);
                 inventory.pick_slot(slot_with_item);
             }
         } else if self.has_infinite_materials() {
+            let slot = inventory.get_suitable_hotbar_slot();
+
+            inventory.set_selected_slot(slot);
             inventory.add_and_pick_item(item_stack);
```
JunkyDeveloper: I checked the vanilla code and why do you do here set_selected_slot? because add_and_pick_item does it's own search still. So how did you tested it?
and it sets the selected slot also

JunkyDeveloper (changes requested):

steel-core/src/player/game_mode/block_interaction.rs
ezraclintoc: if you have a full hotbar, and the item you are currently holding is enchanted, it should move to another slot
which it currently doesn't

steel-core/src/player/player_inventory/core.rs
```diff
@@ -235,6 +261,20 @@ impl PlayerInventory {
         -1
     }
 
+    /// Finds a slot containing an item matching the exact given stack (same item type).
+    /// Returns -1 if not found.
+    #[must_use]
+    pub fn find_slot_matching_item_with_same_components(&self, stack: &ItemStack) -> i32 {
```
banocean: Not sure is there any benefit of using -1 instead of using `Option<u8>`

steel-registry/src/item_stack.rs
```diff
@@ -586,6 +586,20 @@ impl ItemStack {
...
+    pub fn is_enchanted(&self) -> bool {
+        let Some(enchantments) = self.get_enchantments() else {
+            return false;
+        };
+
+        if enchantments.is_empty() {
+            return false;
+        }
+
+        true
```
banocean:
```suggestion
        self.get_enchantments().map_or(false, |e| !e.is_empty())
```

seems like there was a lot more going on earlier because `!enchantments.is_empty()` could be written here instead of the whole if statement

steel-core/src/player/game_mode/block_interaction.rs
```diff
@@ -174,6 +174,7 @@ impl Player {
     /// # Panics
     ///
     /// Panics if the behavior registry has not been initialized.
+    /// TODO Implement ctrl + middle click
```
banocean: I have no idea is that how are partial implementations handled here, but it's something to double check are todo's like that having place in random comments

We could always stack pr's or something 😅

banocean (changes requested):
that if statement is for sure something to change, other things look alright

steel-core/src/player/player_inventory/core.rs
ezraclintoc: I just used that because get_free_slot was using that, but I think we should also change get_free_slot to use option

ezraclintoc: also find_slot_matching_item returns -1

steel-core/src/player/game_mode/block_interaction.rs
ezraclintoc: actually wait isn't `// TODO: If include_data, copy the block entity data into the picked item stack.` talking about ctrl middle click?

steel-core/src/player/player_inventory/core.rs
banocean: > I just used that because get_free_slot was using that, but I think we should also change get_free_slot to use option

I mean here it still feels obsolete since its casted to u8 and usize from what I see

ezraclintoc:
yo, clippy wants to replace 

`266 | /         for i in 0..Self::INVENTORY_SIZE {`
`267 | |             if !self.items[i].is_empty()`
`268 | |                 && ItemStack::is_same_item_same_components(&self.items[i], stack)`
`273 | |         None`

to 
`266 ~         (0..Self::INVENTORY_SIZE).find(|&i| !self.items[i].is_empty()`
`267 +                 && ItemStack::is_same_item_same_components(&self.items[i], stack))`

personally I find it less readable, but I thin clippy will fail if I don't make the change. and idk, maybe I am just not used to the notation

banocean:
I'll take a look at it after I wake up (so somewhere between 6 and 10h from now)

steel-core/src/player/player_inventory/core.rs
```diff
@@ -212,33 +212,57 @@ impl PlayerInventory {
...
-        }
-        -1
+    pub fn find_slot_matching_item(&self, stack: &ItemStack) -> Option<usize> {
+        (0..Self::INVENTORY_SIZE)
+            .find(|&i| !self.items[i].is_empty() && ItemStack::is_same_item(&self.items[i], stack))
+    }
+
+    /// Finds a slot containing an item matching the exact given stack (same item type).
+    #[must_use]
+    pub fn find_slot_matching_item_with_same_components(&self, stack: &ItemStack) -> Option<usize> {
```
banocean: Why `usize` when other functions like `set_selected_slot` take `u8`?

banocean: it seems to be stored as u8 in `<PlayerInventory>.selected`

steel-core/src/player/player_inventory/core.rs
```diff
@@ -212,33 +212,57 @@ impl PlayerInventory {
...
+            !self.items[i].is_empty()
+                && ItemStack::is_same_item_same_components(&self.items[i], stack)
+        })
     }
 
     /// Swaps items between selected hotbar slot and the given slot.
     /// Used for pick block when item is in main inventory but not hotbar.
-    pub fn pick_slot(&mut self, slot: i32) {
-        let slot = slot as usize;
+    pub fn pick_slot(&mut self, slot: usize) {
```
banocean: All this is so inconstant :sob:

steel-core/src/player/player_inventory/core.rs
banocean: I really don't know what would bve a good approach since indexes are stored as usize but slot values are max u8 anyway

banocean (commented):
I don't know honestly besides that seems good

steel-core/src/player/player_inventory/core.rs
ezraclintoc: should we make permanent values (like values that live with the player) like a u8 or u32, and then all temporary variables (like variables that are created and deleted when user middle clicks) usize, or whatever is most convenient?

ezraclintoc: or should I just switch to just using u8. I think the only reason I could see that we'd want u32  (over u8) for inventory ID is for mod support

coco875: what are the size in packet encoded ? to know the limit, if it's u8 then let use u8 for everything, if it's u16 then u16, etc

coco875 (commented):
hmm it seems that the max for slot currently in vanilla are 36 for inventory, 43 for container, so limit to u8 are enough, so just put u8 everywhere

coco875 (approved):

JunkyDeveloper (approved):

Issue #390, opened 2026-08-20
Pick block swaps the held item even when an empty hotbar slot is available


ezraclintoc:
## Summary

Steel's [`pick_slot` always swaps into the currently selected hotbar slot](https://github.com/Steel-Foundation/SteelMC/blob/29452b529e5e96c52bf1593db91d3ae26c92c98a/steel-core/src/player/player_inventory/core.rs#L238-L248), without first choosing a suitable hotbar slot.

When pick-block finds the clone item in the main inventory rather than the hotbar, Steel swaps that stack with whatever the player is currently holding, even when another hotbar slot is empty. The held item is displaced into the main-inventory slot the picked item came from.

For example, holding a diamond sword with an empty hotbar slot beside it and middle-clicking a stone block that is stored in the main inventory moves the sword out of the hotbar and the stone into its place.

## Vanilla behavior

Vanilla's `Inventory.pickSlot` re-selects the hotbar slot before swapping:

```java
public void pickSlot(int slot) {
    this.setSelectedSlot(this.getSuitableHotbarSlot());
    ItemStack tmp = this.items.get(this.selected);
    this.items.set(this.selected, this.items.get(slot));
    this.items.set(slot, tmp);
}
```

See [`Inventory.pickSlot`](https://github.com/ohnodev/decompiled-minecraft-26-2-SNAPSHOT-2/blob/535cd83164af1680abe81dadd7fbe93ded4968a1/src/net/minecraft/world/entity/player/Inventory.java#L108-L113).

[`Inventory.getSuitableHotbarSlot`](https://github.com/ohnodev/decompiled-minecraft-26-2-SNAPSHOT-2/blob/535cd83164af1680abe81dadd7fbe93ded4968a1/src/net/minecraft/world/entity/player/Inventory.java#L140-L155) scans the nine hotbar slots starting at the selected slot and wrapping, preferring, in order:

1. the first empty slot,
2. the first slot holding an unenchanted item,
3. the selected slot as a fallback.

```java
public int getSuitableHotbarSlot() {
    int index;
    int slot;
    for (slot = 0; slot < 9; ++slot) {
        index = (this.selected + slot) % 9;
        if (!this.items.get(index).isEmpty()) continue;
        return index;
    }
    for (slot = 0; slot < 9; ++slot) {
        index = (this.selected + slot) % 9;
        if (this.items.get(index).isEnchanted()) continue;
        return index;
    }
    return this.selected;
}
```

[...]

PR #392, opened 2026-08-20
Fix/pick slot

ezraclintoc:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

This PR fixes a bug in which middle click a block would just overwrite what you are currently holding.
This PR fixes that by scanning your hotbar twice (both starting from the current selected position):
The first pass checks for empty slots.
The second pass checks for items that are not enchanted. 
If both return false (meaning that you hotbar is filled with enchanted items, then it returns the current slot)
#390 

## How this was tested

FIll your hotbar completely with some item, then middle click the another block, it should overwrite your current slot.
Empty two slots and test in different hotbar slots what middle clicking does. It should also fill the slot to the right (and wrap around the hotbar)
Fill up your hotbar, get an enchanted item, select it, and middle click, it should not overwrite the enchanted item. 

## Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/steel-core/src/player/game_mode/block_interaction.rs
+++ b/steel-core/src/player/game_mode/block_interaction.rs
@@ -1,4 +1,5 @@
 use super::{block_breaking::BlockBreakAction, *};
+use log::info;
 
 impl Player {
     /// Sends block update packets for a position and its neighbor.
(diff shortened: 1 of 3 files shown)


coco875:
have been merge in https://github.com/Steel-Foundation/SteelMC/pull/389

PR #393, opened 2026-08-20
fix: Prevent chunk ACK races from corrupting pacing

kdcokenny:
## Type of change

- [x] Bug fix
- [x] Breaking change

## Description

an ACK arriving while a chunk batch is being encoded can reset its quota before the batch is sent. subtracting the sent chunks then leaves a negative quota and delays more chunks from loading.

this queues accepted ACK feedback until the next batch is prepared, so it can't change the quota for a batch already being encoded. the queue is bounded by the number of outstanding batches.

pacing fields are now private. pacing still belongs to the connection and survives world changes and respawns.

## How this was tested

regression tests cover ACKs arriving between prepare and commit, feedback limits and ordering, fractional rates, discarded batches, and pacing across respawns.

ran the Fabric 26.2 client GameTest harness against Steel. checked the full chunk view on join, after movement, and after world transfers in both directions. all passed using normal client ACK handling.

also tested in game.

## Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/steel-core/src/player/chunk_sender.rs
+++ b/steel-core/src/player/chunk_sender.rs
@@ -6,7 +6,11 @@
 //! `mark_chunk_pending_to_send` and `drop_chunk` are never blocked for long.
 use rayon::{ThreadPool, prelude::*};
 use rustc_hash::{FxHashMap, FxHashSet};
-use std::sync::{Arc, Weak};
+use smallvec::SmallVec;
+use std::{
+    mem,
+    sync::{Arc, Weak},
+};
 
 use steel_protocol::packet_traits::{ClientPacket, CompressionInfo, EncodedPacket};
 use steel_protocol::packets::game::{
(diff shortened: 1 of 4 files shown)


kdcokenny:
ig i need to reopen it

steel-core/src/player/chunk_sender.rs
```diff
@@ -35,6 +39,76 @@ const START_CHUNKS_PER_TICK: f32 = 9.0;
 /// Maximum unacknowledged batches after first ack (vanilla: 10)
 const MAX_UNACKNOWLEDGED_BATCHES: u16 = 10;
 
+#[derive(Debug)]
+struct ChunkBatchPacing {
+    unacknowledged_batches: u16,
+    desired_chunks_per_tick: f32,
```
banocean, line 45: f32? What does it mean? Is if it's 1.22 it means 1 and 0.22 of a packet of another chunk? I'm really confused here

(question unrelated to code review since it was the case before it was edited by this PR)

steel-core/src/player/chunk_sender.rs
```diff
@@ -35,6 +39,76 @@ const START_CHUNKS_PER_TICK: f32 = 9.0;
...
+    fn drain_feedback(&mut self) {
+        for desired_chunks_per_tick in mem::take(&mut self.accepted_feedback) {
+            self.unacknowledged_batches = self.unacknowledged_batches.saturating_sub(1);
+            self.desired_chunks_per_tick = if desired_chunks_per_tick.is_nan() {
+                MIN_CHUNKS_PER_TICK
+            } else {
+                desired_chunks_per_tick.clamp(MIN_CHUNKS_PER_TICK, MAX_CHUNKS_PER_TICK)
+            };
+
+            if self.unacknowledged_batches == 0 {
```
banocean, line 91: Should it keep looping even when there is no more batches that weren't properly recived?

steel-core/src/player/chunk_sender.rs
```diff
@@ -35,6 +39,76 @@ const START_CHUNKS_PER_TICK: f32 = 9.0;
 /// Maximum unacknowledged batches after first ack (vanilla: 10)
 const MAX_UNACKNOWLEDGED_BATCHES: u16 = 10;
 
+#[derive(Debug)]
+struct ChunkBatchPacing {
+    unacknowledged_batches: u16,
+    desired_chunks_per_tick: f32,
+    batch_quota: f32,
+    max_unacknowledged_batches: u16,
```
banocean, line 47: What is the purpose of this this value since I'm struggling to understand what does it change in practice

banocean (commented):
Overall looks good to me, but I get confused like in half of the places I look at 😓

JunkyDeveloper:
Will you continue working on it and what is the blocker?

kdcokenny:
> Will you continue working on it and what is the blocker?

I changed to draft bc i want to focus on getting the explosions pr and all of it's prereq blockers merged b4 focusing on other features/bug fixes

steel-core/src/player/chunk_sender.rs
kdcokenny: its a sending rate, chunks are still sent whole. fractional credit lets rates below 1 spread a chunk across multiple sending ticks, so 0.25 allows one chunk every four ticks. 1.22 doesn't mean sending part of another packet, the available quota is rounded down when selecting chunks.

steel-core/src/player/chunk_sender.rs
kdcokenny: it can only reach zero on the last queued ACK. we reject feedback once there's already one queued ACK per outstanding batch, so there can't be extra entries left to process after zero. added a comment and assertion for that.

steel-core/src/player/chunk_sender.rs
kdcokenny: it limits how many batches we can send before waiting for ACKs. starts at 1, then opens up to 10 after the first ACK is applied. that lets us keep sending while earlier batches are still being acknowledged, without letting the queue grow indefinitely.

CuzImClicks (approved):

Purdze (approved):

PR #394, opened 2026-08-20
fix: Limit entries in `SSelectKnownPacks` packet to 64

### LINKED ISSUE #386: Limit selected known packs to 64 entries
## Summary

[`SSelectKnownPacks`](https://github.com/Steel-Foundation/SteelMC/blob/29452b529e5e96c52bf1593db91d3ae26c92c98a/steel-protocol/src/packets/config/s_select_known.rs#L5-L9) reads its pack list with Steel's [default 32,767-entry bound](https://github.com/Steel-Foundation/SteelMC/blob/29452b529e5e96c52bf1593db91d3ae26c92c98a/steel-utils/src/serial/mod.rs#L15-L40).

Vanilla limits this list to 64 entries.

## Vanilla behavior

Vanilla applies [`ByteBufCodecs.list(64)`](https://github.com/ohnodev/decompiled-minecraft-26-2-SNAPSHOT-2/blob/535cd83164af1680abe81dadd7fbe93ded4968a1/src/net/minecraft/network/protocol/configuration/ServerboundSelectKnownPacks.java#L19-L21).

A configuration client that sends 65 or more valid `KnownPack` entries should fail during decoding. Steel currently accepts the packet.

## Impact

Steel currently [offers one known pack and does not use the returned list beyond debug logging](https://github.com/Steel-Foundation/SteelMC/blob/29452b529e5e96c52bf1593db91d3ae26c92c98a/steel-login/src/handlers/config.rs#L65-L84), so normal vanilla clients are unaffected.

A custom client can make Steel allocate and parse a much larger list than vanilla permits. Existing packet and string limits keep this bounded, so this is a protocol-validation mismatch rather than an unbounded allocation issue.

## Reproduction

Decode an `SSelectKnownPacks` packet containing:

- 64 valid entries: decoding succeeds.
- 65 valid entries: Steel currently succeeds, while vanilla rejects the packet.

banocean:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

Closes #386 

## How this was tested

I run thru fork of azalea that based on environment variable send either 0, 64 or 65 packs ([changes I made](https://github.com/azalea-rs/azalea/compare/main...banocean:azalea-steel-tests:limit-packs))

To run it (while your server runs on localhost with default port)
1. `git clone -b limit-packs git@github.com:banocean/azalea-steel-tests.git`
2. `cd azalea-steel-tests/test`
3. `export MODE=spam` or `export MODE=wrong` depending on what outcome you test
4. `cargo run`

Desired outcome: Steel emits warn about incorrectly formatted packet when you run the fork with `MODE=wrong` environment variable and doesn't emit warn when you add `MODE=spam` or don't set environment variable at all 

## Checklist

- [X] Self-reviewed the diff


steel-protocol/src/packets/config/s_select_known.rs
```diff
@@ -4,6 +4,6 @@ use crate::packets::shared_implementation::KnownPack;
 
 #[derive(ReadFrom, ServerPacket, Clone, Debug)]
 pub struct SSelectKnownPacks {
-    #[read(as = Prefixed(VarInt))]
+    #[read(as = Prefixed(VarInt), bound = 64)]
```
kdcokenny, line 7: I think you could route this decode failure through the invalid-packet disconnect path. Right now Steel logs the error and continues, leaving the client stuck in configuration, while vanilla disconnects it.

kdcokenny (changes requested):
Also, a regression test would be nice here. It could confirm that 64 valid KnownPack entries decode successfully and 65 are rejected.

banocean:
> Also, a regression test would be nice here. It could confirm that 64 valid KnownPack entries decode successfully and 65 are rejected.

Isn't this behavior shared with other packets that would be sent wrong in configuration?

kdcokenny:
> > Also, a regression test would be nice here. It could confirm that 64 valid KnownPack entries decode successfully and 65 are rejected.
> 
> Isn't this behavior shared with other packets that would be sent wrong in configuration?

fair enough, create the test that accepts 64 packs and rejects 65 and then create an issue for the connection bug and i think we will be good 2 go.

steel-protocol/src/packets/config/s_select_known.rs
banocean: Marking as resolved because of https://github.com/Steel-Foundation/SteelMC/pull/394#issuecomment-5417078308

kdcokenny (approved):
The tests look good. I don’t see the follow-up issue for the connection bug yet. Please create it and link it here

banocean:
> The tests look good. I don’t see the follow-up issue for the connection bug yet. Please create it and link it here

#527, created by @carabistouflette

kdcokenny:
> > The tests look good. I don’t see the follow-up issue for the connection bug yet. Please create it and link it here
> 
> #527, created by @carabistouflette

killa, ready for the maintainers now! @coco875 @JunkyDeveloper

coco875 (approved):

JunkyDeveloper (approved):

PR #395, opened 2026-08-20
feat(dragon egg block gravity)

ezraclintoc:
Added gravity to dragon egg block. The current implementation is just a copy of sand_block.rs, but renamed to DragonEggBlock.

## Type of change

- [x] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

Implements dragon egg. Implements gravity, and teleportation. Includes check that make sure the teleport location is valid.

## How this was tested

Dropping the egg unto solid blocks.
Dropping the egg onto torches. (broke)
Right clicking the egg in creative. (teleported)
Left clicking egg in creative. (broke)
Right and Left clicking egg in survival. (teleported)

All of these are expect and are vanilla parity.

## Checklist

- [x] Self-reviewed the diff


steel-core/src/behavior/blocks/falling/dragon_egg_block.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// Creates the server-side behavior for the dragon egg
+    #[must_use]
+    pub const fn new(block: BlockRef) -> Self {
+        Self {
+            falling: FallingBlock::new(block),
+        }
+    }
+
+    fn teleport(state: BlockStateId, world: &Arc<World>, pos: BlockPos) {
+        for _ in 0..1000 {
```
coco875: maybe const MAX_RANDOM_TELEPORT_TRIAL ? (I don't think it's a good name but it's still a magic number)

steel-core/src/behavior/blocks/falling/dragon_egg_block.rs
```diff
@@ -0,0 +1,120 @@
...
+            let x = random_range(0..15) - random_range(0..15);
+            let y = random_range(0..7) - random_range(0..7);
+            let z = random_range(0..15) - random_range(0..15);
```
coco875: also magic number maybe MAX_X, MAX_Y, MAX_Z ?

coco875 (changes requested):

steel-core/src/behavior/blocks/falling/dragon_egg_block.rs
ezraclintoc: Ok, I fixed this in my latest commit

steel-core/src/behavior/blocks/falling/dragon_egg_block.rs
ezraclintoc: I also fixed this in my latest commit. They are i32, because
1. It doesn't really matter if its negative
2. We don't have to convert to i32 later

coco875 (dismissed):

coco875 (approved):

JunkyDeveloper (approved):
did this bug an

PR #399, opened 2026-08-20
feat: implement chorus growth

dogbonewish:
## Type of change

- [X] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

This PR implements vanilla chorus flower growth and chorus plant update behavior 

CF (Chorus Flowers) now grow upward, create horizontal branches, they can age, they can/will become dead flowers, emit growth and death events, and break when obstructed. The CP (Chorus Plant, skinny purple bit) now update and attach when the neighboring blocks change, and obstructed plants break on scheduled ticks, chorus plants are not included in the "pathfinding".

## How this was tested

Tested on NixOS (Linux) with Vanilla Minecraft 26.2.

```text
cargo +nightly-2026-07-23 test -p steel-core chorus_
cargo +nightly-2026-07-23 check -p steel-core
cargo +nightly-2026-07-23 fmt --all --check
```

## Screenshots / logs

image
Lots of chorus plants planted and grown.

## Checklist

- [X] Self-reviewed the diff

## Additional notes

I believe this completes the tracker TODOs for both ChorusFlowerBlock and ChorusPlantBlock.


steel-core/src/behavior/blocks/vegetation/chorus_flower_block.rs
```diff
@@ -38,6 +43,126 @@ impl ChorusFlowerBlock {
...
+        }
+
+        let mut grow_upwards = false;
+        let mut pillar_on_support_block = false;
+        let below = world.get_block_state(pos.below());
+        if below.get_block().has_tag(&BlockTag::SUPPORTS_CHORUS_FLOWER) {
+            grow_upwards = true;
+        } else if below.get_block() == self.plant {
+            let mut height = 1;
+            for _ in 0..4 {
```
coco875: the 4 is for what ? (magic number)

coco875: (when i write magic number I mean that you need to add a const to easily know what it do)

steel-core/src/behavior/blocks/vegetation/chorus_flower_block.rs
```diff
@@ -38,6 +43,126 @@ impl ChorusFlowerBlock {
...
+                    pillar_on_support_block = test_state
+                        .get_block()
+                        .has_tag(&BlockTag::SUPPORTS_CHORUS_FLOWER);
+                    break;
+                }
+                height += 1;
+            }
+
+            if height < 2
+                || height <= rng.random_range(0..if pillar_on_support_block { 5 } else { 4 })
```
coco875: 2, 5, 4 are for what ? (magic number)

coco875 (changes requested):

steel-core/src/behavior/blocks/vegetation/chorus_flower_block.rs
dogbonewish: Yeah, fair. I pulled the vanilla limits into named constants in fd39daca7 so the scan depth and growth ranges are clear.

steel-core/src/behavior/blocks/vegetation/chorus_flower_block.rs
dogbonewish: Fixed in fd39daca7. The minimum pillar height, support growth range, and branch-attempt range are named now.

coco875 (approved):

steel-core/src/behavior/blocks/vegetation/chorus_plant_block.rs
```diff
@@ -80,6 +82,17 @@ impl ChorusPlantBlock {
             west.get_block() == block || west.get_block() == &vanilla_blocks::CHORUS_FLOWER,
         )
     }
+
+    const fn property_for_direction(direction: Direction) -> &'static BoolProperty {
```
Joniii11, line 86: maybe move this to shared utility since other directional blocks can reuse it too

Joniii11 (approved):
LGTM

steel-core/src/behavior/blocks/vegetation/chorus_plant_block.rs
JunkyDeveloper: In general I would agree here, but idk how many will use it (because many blocks have different ways to express that), so I think it is ok to be here. But good thought and please more like this!

JunkyDeveloper (approved):

PR #405, opened 2026-08-20
feat: implement glow lichen spreading

dogbonewish:
## Type of change

- [X] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

This PR implements vanilla glow lichen spreading and bonemeal behavior.

Glow lichen can now spread to open faces, spread around supporting blocks, become waterlogged when placed in source water, and update correctly when neighboring blocks change. Unsupported faces are removed using the shared multiface block behavior.

Bonemeal can now spread glow lichen to valid nearby faces using vanilla direction selection and placement rules.

## How this was tested

Tested on NixOS (Linux) with Vanilla Minecraft 26.2.

```text
cargo +nightly-2026-07-23 test -p steel-core glow_lichen_
cargo +nightly-2026-07-23 check -p steel-core
cargo +nightly-2026-07-23 fmt --all --check
```

Tested in-game with glow lichen spreading across supporting blocks, spreading around corners, waterlogging, unsupported-face removal, and bonemeal growth.

## Screenshots / logs

Glow lichen was tested in-game. I don't have any screenshots/gif recordings because i forgor :(

## Checklist

- [X] Self-reviewed the diff

## Additional notes

I believe this completes the tracker TODOs for GlowLichenBlock and its spreading and bonemeal behavior.


steel-core/src/behavior/blocks/vegetation/multiface_block.rs
```diff
@@ -106,7 +122,7 @@ impl MultifaceBlock {
     fn get_state_for_placement_with_dir(
         block: BlockRef,
         old_state: BlockStateId,
-        world: &Arc<World>,
+        world: &dyn LevelReader,
```
coco875, line 152: can we use impl instead of dyn ?

coco875: also I don't get why this change ? because the only other LevelReader are WorldGenRegion but it's not use ?

coco875: hmm I just see it's use in spread_from_face_toward_direction, can we make it generic like the function ?

coco875 (changes requested):
can you try reduce the usage of dyn because of performance penalty that dyn make (see https://doc.rust-lang.org/std/keyword.dyn.html#trade-offs) ?

steel-core/src/behavior/blocks/vegetation/multiface_block.rs
dogbonewish: The mutating spread function was already generic. The read-only helpers stay on `&dyn LevelReader` because bonemeal supplies that trait object and the dynamic block-shape API uses the same interface. I also moved the spread-position calculation into one helper shared by runtime and worldgen.

dogbonewish:
I pushed a small cleanup here. The mutating spread path is generic; the remaining `dyn LevelReader` use is at the existing read-only/block-shape boundary, where the API already receives a trait object. The shared spread geometry now lives in one helper used by both runtime and worldgen.

steel-core/src/behavior/blocks/vegetation/multiface_block.rs
```diff
@@ -26,18 +26,40 @@ pub(super) struct MultifaceSpreader {
...
 
 #[derive(Clone, Copy)]
-enum MultifaceSpreadType {
+pub(crate) enum MultifaceSpreadType {
     SamePosition,
     SamePlane,
     WrapAround,
 }
 
+pub(crate) fn multiface_spread_pos(
```
coco875: why not just pub ?

coco875 (changes requested):

steel-core/src/behavior/blocks/vegetation/multiface_block.rs
JunkyDeveloper: I think pub (crate) is enough because I think it will not really used on a thing which sits on top

dogbonewish: it’s only used inside steel-core, so pub(crate) is enough here. making it pub would expose an API we don’t need yet.

JunkyDeveloper (approved):

coco875 (approved):

PR #413, opened 2026-08-21
blocks/vegetation: Fixed multiface attach checks offsetting the neighbour position twice.

didntpot:


steel-core/src/behavior/blocks/vegetation/mossy_carpet_block.rs
```diff
@@ -78,8 +78,7 @@ impl MossyCarpetBlock {
         pos: BlockPos,
         direction: Direction,
     ) -> bool {
-        direction != Direction::Up
-            && MultifaceBlock::can_attach_to(world, pos.relative(direction), direction)
+        direction != Direction::Up && MultifaceBlock::can_attach_to(world, pos, direction)
```
GreenedDev, line 81: we should instead change args of can_attach_to() function.
instead of taking pos, it should accept neighbor_pos. (getting neighbor pos should happen outside of the function, like vanilla)

after changing this, you should check all uses of can_attach_to() function to see if it is used correctly

steel-core/src/behavior/blocks/vegetation/multiface_block.rs
```diff
@@ -37,11 +37,12 @@ impl MultifaceBlock {
         Self { block }
     }
 
-    /// Vanilla `MultifaceBlock.canAttachTo(level, directionTowardsNeighbor, neighborPos, neighborState)`.
+    /// Vanilla `MultifaceBlock.canAttachTo(level, pos, direction)`.
```
GreenedDev, line 40: look at what i commented above

coco875 (dismissed):
Look good, thanks to help fix the regression

steel-core/src/behavior/blocks/vegetation/vine_block.rs
```diff
@@ -71,7 +71,12 @@ impl VineBlock {
         neighbour_pos: BlockPos,
         direction_to_neighbour: Direction,
     ) -> bool {
-        MultifaceBlock::can_attach_to(level, neighbour_pos, direction_to_neighbour)
```
GreenedDev, line 74: this shouldn't be changed. fix the can_attach_to() fn as i said above.

GreenedDev (changes requested):
can_attach_to() functions needs to be fixed.

the misalignment was introduced mistakenly by me, in https://github.com/Steel-Foundation/SteelMC/commit/a3ebd125b6a6dcea6ec790c5805aa0121d8fea97#diff-b5ce5c0a0758ab3d2db58b1594efe31a7c41217a29d64799a8ac4783c3cbebe3

steel-core/src/behavior/blocks/vegetation/mossy_carpet_block.rs
GreenedDev: thus, previous code on line 82 was correct.

didntpot: that makes sense, will change

coco875:
I let handle @GreenedDev for the review like it's him that have made the regression and not expert in block implementation

steel-core/src/behavior/blocks/vegetation/multiface_block.rs
```diff
@@ -145,12 +146,7 @@ impl MultifaceBlock {
             && (old_state.get_block() != block || !Self::has_face(old_state, placement_direction))
         {
             let neighbor_pos = placement_pos.relative(placement_direction);
-            return Self::can_attach_to_state(
-                world,
-                placement_direction,
-                neighbor_pos,
-                world.get_block_state(neighbor_pos),
-            );
+            return Self::can_attach_to(world, neighbor_pos, placement_direction);
```
GreenedDev: why did you change this?
vanilla isValidStateForPlacement uses the can_attach_to function (but the one with neighborState arg. this function is equivalent to our can_attach_to_state())

Image

didntpot:
@GreenedDev I reverted back to the previous fix, added another test in there while at it

GreenedDev (approved):
LGTM. your current changes seem to be aligned with vanilla.

not sure about the tests you wrote tho.

didntpot:
fair enough on the tests

GreenedDev (approved):
LGTM

coco875 (approved):

JunkyDeveloper (approved):

PR #414, opened 2026-08-21
perf(chunk): Index prepared chunks by position when committing a batch

BryanFRD:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [x] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

`ChunkSender::commit_batch` re-scanned the whole prepared batch for every encoded chunk to
find its matching `PreparedChunk`:

```rust
let Some(prepared) = batch.chunks.iter().find(|chunk| chunk.pos == encoded.pos) else {
```

That is quadratic in the batch size. `encode_batch` maps over `batch.chunks` with an
order-preserving parallel iterator and only ever drops entries, so `encoded_chunks` is a
subsequence of `batch.chunks` in the same relative order. A single forward cursor over
`batch.chunks` therefore resolves the whole batch in one pass, which is what this PR does.
Behavior is unchanged.

Following the review, the resolution loop now lives in a pure `resolve_valid_chunks`
instead of being inlined in `commit_batch`. `commit_batch` needs a `PlayerConnection` and
cannot be driven from a test or a benchmark; the extracted function can, and it is the part
this PR actually changes.

## Measured

`steel-core/benches/chunk_sender.rs`, run with:

```
cargo bench -p steel-core --bench chunk_sender --features benchmark-support
```

The fixtures are real `PreparedChunk` and `EncodedChunk` values produced by `encode_batch`,
so the real `is_current_for` cost (weak upgrade, pointer equality, readiness and revision
loads) is inside every measurement. Each strategy is asserted to resolve the same chunks as
the rescan baseline before it is timed. `index` is the `FxHashMap` this PR originally
proposed, `cursor` is the forward pass it ships.

Ryzen 7 9850X3D, 16 threads, `[profile.bench]`. Medians.

Batch fully encoded, the nominal case:

| chunks | rescan | index | cursor | rescan/cursor |
| -----: | -----: | ----: | -----: | ------------: |
| 16 | 0.19 us | 0.26 us | 0.18 us | 1.1x |
| 64 | 0.80 us | 0.85 us | 0.63 us | 1.3x |
| 128 | 2.81 us | 1.66 us | 1.22 us | 2.3x |
| 256 | 9.74 us | 3.19 us | 2.42 us | 4.0x |
| **500** | **31.31 us** | **6.34 us** | **4.79 us** | **6.5x** |
| 1000 | 110.14 us | 12.98 us | 9.75 us | 11.3x |
| 2000 | 411.10 us | 25.37 us | 19.13 us | 21.5x |

`MAX_CHUNKS_PER_TICK` is 500 today, the bolded row. That is the number this PR changes:
31.31 us down to 4.79 us per player per chunk-send tick, on a full batch.

Where that sits in a whole send tick at 500 chunks:

| step | time |
| ---- | ---: |
| `encode_batch`, cold cache | 167.29 us |
| `encode_batch`, warm cache | 70.42 us |
| resolution, rescan | 31.19 us |
| resolution, cursor | 4.86 us |

So resolution was about 16% of a cold send tick and 31% of a warm one, and is now about 3%
and 6%.

The bench also covers two batch shapes where the win is smaller or absent, because they are
the honest cases:

- `half-encoded`, every other chunk dropped during encoding: same shape as above, 6.7x at 500.
- `tail-only`, a single chunk surviving encoding: 0.14 us rescan against 0.20 us cursor at
  500. Both walk the batch once and the cursor pays a little iterator overhead. `index` is
  the worst of the three here at 1.21 us, since it builds the entire map to serve one lookup.

Two corrections to what I claimed earlier in this PR. The first version of this bench used a
synthetic struct rather than real chunks and reported 17.3x at the 500-chunk cap; with the
real validity check in the loop it is 6.5x. And chunk sending runs on its own loop, separate
from the game tick, so none of this is TPS.

The cursor beats the hash index at every size that matters, which is why this PR ships the
cursor rather than the `FxHashMap` it originally proposed.

## How this was tested

```
cargo fmt --check                                                      clean
cargo clippy -p steel-core --all-targets --features benchmark-support  no warnings
cargo test -p steel-core                                               2535 passed, 11 ignored
```

Four new tests cover `resolve_valid_chunks`: batch order is preserved, prepared chunks that
failed to encode are walked past, chunks no longer pending are dropped, and chunks whose
holder was demoted after encoding are dropped. Replacing the cursor's `find` with a `next`
fails two of them, so they do catch the regression they are there for.

The `prepared_full_chunk` fixture that already existed in the test module moved into the
`benchmark-support` module so the tests and the bench share one fixture instead of two
copies. That feature and its pattern already exist in the crate (`generation_benchmark_support`).

## Screenshots / logs

N/A

## Checklist

- [x] Self-reviewed the diff

## Additional notes

I left `MAX_CHUNKS_PER_TICK` alone. It is the 500 vs vanilla's 64 that makes this worth
fixing at all, but that is a deliberate choice in this file and not something this PR
should touch.


steel-core/src/player/chunk_sender.rs
```diff
@@ -252,12 +252,18 @@ impl ChunkSender {
         }
         drop(epoch);
 
+        let prepared_chunks: FxHashMap<ChunkPos, &PreparedChunk> = batch
```
Joniii11: Could we avoid building an FxHashMap here Cause encoded_chunks preserve the batch order a forward only iterator over batch.chunks should keep this O(n) while avoiding the extra allocation and hashing overhead

i mean like this:
```rs
let mut prepared_chunks = batch.chunks.iter();
let mut valid_chunks = Vec::with_capacity(encoded_chunks.len());

for encoded in encoded_chunks {
    if !self.pending_chunks.contains(&encoded.pos) {
        continue;
    }

    let Some(prepared) = prepared_chunks.find(|c| c.pos == encoded.pos) else {
        continue;
    };

    if encoded.is_current_for(prepared) {
        valid_chunks.push(encoded);
    }
}
```
could you try and see what the results are from this version?

Joniii11 (changes requested):

coco875:
How have you measured this performance improvement ?

GreenedDev:
i think only someone more experienced with chunkgen (like 4lve) should review this. preferably 4lve

coco875:
here it optimise chunk send not chunkgen

BryanFRD:
> How have you measured this performance improvement ?

Its a Criterion bench that puts the two strategies side by side with a real `PreparedChunk`, in an ither file, it does not call `commit_batch` itself

steel-core/src/player/chunk_sender.rs
BryanFRD: It is indeed faster

coco875:
already existing in the repo ? (then say the name) or that you made ?

steel-core/src/player/chunk_sender.rs
Joniii11: Wounderful, if you want you can use this than :D

BryanFRD:
Criterion is already used in the project, I've made a new file for the bench that doesn't test `commit_bathc` itself, only the changes with real data etc, but I didn't make a `PlayerConnection` to test `commit_batch`

BryanFRD:
I can take some time to try figuring out how to bench it with a real connection and propose it in the same PR if you want

coco875:
yes to being able to reproduce number, else you can split the part that gather chunk into a pure function

BryanFRD:
The numbers are reproducible now. I extracted the resolution loop out of `commit_batch` into a pure `resolve_valid_chunks`, like you suggested, and added `steel-core/benches/chunk_sender.rs` behind the existing `benchmark-support` feature:

```
cargo bench -p steel-core --bench chunk_sender --features benchmark-support
```

It builds real `PreparedChunk` and `EncodedChunk` values through `encode_batch`, so the actual `is_current_for` cost is in the measurement, and it checks every strategy against the rescan baseline before timing it. Three batch shapes and seven sizes, plus a group placing the resolution inside a full send tick.

This also corrects my earlier claim. My first bench used a synthetic struct and reported 17.3x at the 500-chunk cap. With real chunks it is 6.5x, 31.31 us down to 4.79 us. The PR description is updated with the full tables, including the shapes where the win is small or absent.

steel-core/src/player/chunk_sender.rs
```diff
@@ -296,6 +285,36 @@ impl ChunkSender {
...
+    /// `encoded_chunks` is a subsequence of `prepared` in the same relative order:
+    /// [`Self::encode_batch`] maps over `prepared` with an order-preserving parallel
+    /// iterator and only drops entries. A single forward cursor over `prepared` therefore
+    /// resolves every encoded chunk in one pass, instead of restarting the scan per chunk.
+    fn resolve_valid_chunks(
+        prepared: &[PreparedChunk],
+        encoded_chunks: Vec<EncodedChunk>,
+        pending: &FxHashSet<ChunkPos>,
+    ) -> Vec<EncodedChunk> {
+        let mut prepared_chunks = prepared.iter();
```
coco875, line 299: hmm from what I see the only optimisation are only do the convertion to iter once, is it correct ?

BryanFRD: Right about the diff, but the win isn't the .iter()`call. That one is free: slice::iter is just a pointer pair and compiles away. If that were the only effect the change would be worth nothing.
What actually changes is that the cursor now survives between lookups. Iterator::find takes &mut self and consumes as it goes:

- before, a fresh iterator per encoded chunk, so every find restarts at index 0. Lookup x scans x entries, quadratic over the batch.
- after, one iterator for the whole loop, so each find resumes where the previous one stopped. The cursor crosses prepared exactly once, linear.

So it's the restart that was expensive, not the construction.

CuzImClicks (approved):

steel-core/src/player/chunk_sender.rs
coco875: ok thanks for the explanation

coco875 (approved):

PR #416, opened 2026-08-21
fix inverted keep mode in setblock

itzzfinicalfire:
keep was failing on air targets and placing into occupied ones. vanilla only places when the target block is air and fails otherwise.

## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Checklist

- [ ] Code builds w/o errors or warnings
- [ ] Self-reviewed the diff
- [ ] No leftover debug code / comments

### DIFF
--- a/steel-core/src/command/builtins/setblock.rs
+++ b/steel-core/src/command/builtins/setblock.rs
@@ -90,8 +90,8 @@ fn set_block(
     // World the player is in
     let level = context.source().world();
 
-    // Keep mode throw an error when you try to replace an air block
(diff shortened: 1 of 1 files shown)


coco875 (approved):
a similar fix then https://github.com/Steel-Foundation/SteelMC/commit/71142b0cbd36c816a0ba05ed3f5087040d76e160 x)

JunkyDeveloper (approved):

PR #417, opened 2026-08-21
feat: implement hopper block

pxperr:
## Type of change

- [x] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

This adds the hopper: the `HopperBlock` behavior, `HopperBlockEntity` with the item transfer logic, and the 5-slot `HopperMenu`. I worked from the decompiled 26.2 source and mirrored the vanilla methods rather than reimplementing the behavior from scratch.

Block side:

- `getStateForPlacement` points the spout opposite the clicked face. A Y-axis result collapses to `Down`, since `facing_hopper` has no `up` value. `enabled` is `true` at placement.
- `onPlace` and `neighborChanged` both run `checkPoweredState`, so a hopper placed next to an already-powered neighbor locks straight away.
- `useWithoutItem` unpacks a pending loot table and then opens the menu.
- `getTicker` is unconditional, like vanilla. `enabled` is read inside the tick, not when the ticker is handed out.
- `getAnalogOutputSignal` drives comparators, and `entityInside` forwards to the block entity.

Block entity side, keeping the vanilla names: `pushItemsTick`, `tryMoveItems`, `ejectItems`, `suckInItems`, `tryTakeInItemFromSlot`, `addItem`, `tryMoveInItem` (including the cooldown hand-off when an empty hopper is filled), `canMergeItems`, `isFullContainer`, and `getItemsAtAndAbove` over the vanilla suck box. `TransferCooldown` persists in NBT and defaults to -1, and `LootTable` / `LootTableSeed` are handled on load and save.

I also added `block_entity/randomizable.rs`, which mirrors vanilla `RandomizableContainer` and `LootTable.fill`, including the shuffled empty slots and `shuffleAndSplitItems`. It is wired up for the hopper. Barrel and chest can adopt it when they gain loot support.

Divergences, each documented in code:

- Loot unpacking is deterministic for a given table and seed within Steel, but the slot layout is not bit-identical to vanilla. Loot evaluation does not run on the vanilla LCG yet. This is the same limitation the brushable block already has, tracked by the TODO on `LootTable::get_random_items`.
- `getContainerAt` only finds block-entity containers. Steel has no container entities and no `WorldlyContainerHolder` blocks yet, so there is nothing else to look for. Noted at the function.
- The `INSPECT_HOPPER` stat is left as a TODO, matching the `OPEN_BARREL` TODO in barrel.

Deferred on purpose:

- Hopper minecart.
- Sided container slots, so furnace and brewing stand insertion uses the flat slot order for now.
- Jukebox disc insertion and composter filling.
- Double chests.
- The container open/close counter, which barrel carries the same TODO for.

One visibility change outside the hopper: `World::is_collision_shape_full_block_at` is now `pub(crate)` so the suction path can check whether the block above blocks item entities.

## How this was tested

16 unit tests sit next to the implementation and cover:

- Placement facings, including the vertical collapse to `Down`.
- The redstone lock, including placing a hopper against a neighbor that is already powered.
- The ticker staying unconditional while `enabled` is false.
- Comparator output.
- Pushing one item into the facing container and pulling one from the container above, with the cooldown.
- Item entity suction, including the exact y+2 boundary of the suck box, and a solid block above blocking it.
- `entityInside` reaching the block entity.
- `TransferCooldown` and loot table NBT round-trips.
- Deterministic seeded loot unpacking.

`cargo test -p steel-core --lib hopper` gives 16 passed, 0 failed.

Full checks with `RUSTFLAGS=-Dwarnings` all pass:

- `cargo test`
- `cargo clippy -r --all-targets --all-features`
- `cargo fmt --all --check`

In game on 26.2 I checked hoppers feeding a barrel, picking up items dropped above them, a lever locking and unlocking one, comparator readings as a hopper filled, and that a solid block on top stops suction.

## Screenshots / logs
image
image
image

## Checklist

- [x] Self-reviewed the diff

## Additional notes

`randomizable.rs` is the only part here meant to be shared. I put it in its own module rather than inside the hopper so the other randomizable containers can pick it up without code moving later.


steel-core/src/block_entity/randomizable.rs
```diff
@@ -0,0 +1,167 @@
...
+    ///
+    /// Seed 0 rolls with runtime randomness; any other seed is deterministic.
+    /// Steel's loot evaluation runs on `rand` rather than the vanilla LCG (see
+    /// the RNG note on `LootTable::get_random_items`), so slot arrangement is
+    /// deterministic per seed but not bit-identical to vanilla saves.
+    pub fn fill_container(self, origin: BlockPos, container: &mut dyn Container) {
+        if self.seed == 0 {
+            self.fill_with_rng(origin, container, &mut rand::rng());
+        } else {
+            let mut rng = StdRng::seed_from_u64(self.seed as u64);
```
Joniii11: Vanilla 26.2 uses its own RandomSource for seeded loot. StdRng is deterministic but not Vanillaidentical so the same LootTableSeed can produce different loot

so use the world rng not rand::rng

steel-core/src/block_entity/entities/hopper.rs
```diff
@@ -0,0 +1,933 @@
...
+        self.try_move_items(world, pos, state, || self.add_item_from_entity(item_entity));
+    }
+
+    /// Vanilla `Hopper.SUCK_AABB` moved to the hopper position: the full column
+    /// from the bowl rim (y 11/16) to one block above the hopper.
+    fn suck_aabb(pos: BlockPos) -> WorldAabb {
+        let base = DVec3::new(f64::from(pos.x()), f64::from(pos.y()), f64::from(pos.z()));
+        WorldAabb::from_min_max(
+            base + DVec3::new(0.0, 11.0 / 16.0, 0.0),
+            base + DVec3::new(1.0, 2.0, 1.0),
```
coco875: magic number

steel-core/src/block_entity/entities/hopper.rs
```diff
@@ -0,0 +1,933 @@
...
+    }
+
+    fn is_empty_after_unpack(&self, pos: BlockPos) -> bool {
+        let mut container = self.container.lock();
+        container.unpack_loot_table(pos);
+        container.is_empty()
+    }
+
+    /// Vanilla `ejectItems`: moves one item into the container this hopper faces.
+    fn eject_items(&self, world: &Arc<World>, pos: BlockPos, state: BlockStateId) -> bool {
```
Joniii11, line 190: vanila hopper use WorldlyContainer sided slots and perface insert & extract checks, the flat slot fallback can move items through slots Vanilla would reject

So add  WorldlyContainer capability and use getSlotsForFace or per face permission checks for hopper push & pull

steel-core/src/inventory/menu/kinds/hopper_menu.rs
```diff
@@ -0,0 +1,68 @@
...
+#[must_use]
+pub fn hopper(
+    inventory: Shared<PlayerInventory>,
+    container_id: u8,
+    container: impl Into<ContainerRef>,
+) -> Menu {
+    let container = container.into();
+
+    let mut builder = MenuBuilder::new(&vanilla_menu_types::HOPPER, container_id);
+    let hopper = builder.section(&container, 5);
```
coco875: magic number

Joniii11 (changes requested):

coco875 (commented):
overall look good but will make more test and search about the behavior of hopper

coco875:
I have try:
* double hopper speed
* sorting with water
* sorting with ice
* locking hopper

I need to try comparetor and some more sorting system (like filter)

pxperr:
> I have try:
> 
> * double hopper speed
> * sorting with water
> * sorting with ice
> * locking hopper
> 
> I need to try comparetor and some more sorting system (like filter)

locking hopper is working, I can work through the others now

steel-core/src/block_entity/randomizable.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// deterministic per seed but not yet bit-identical to vanilla saves
+    /// (tracked on `LootTable::get_random_items`).
+    pub fn fill_container(self, origin: BlockPos, container: &mut dyn Container) {
+        if self.seed == 0 {
+            // Vanilla falls back to random sequences / the level random here;
+            // Steel has neither yet, so runtime randomness stands in.
+            self.fill_with_rng(origin, container, &mut rand::rng());
+        } else {
+            // Vanilla `RandomSource.create(seed)`.
+            let mut rng = RandBridge(LegacyRandom::from_seed(self.seed as u64));
```
Joniii11, line 92: RandBridge still leaves the sampling behavior to rand so fixed seeds can still differ from Vanilla! Please make the loot pipeline use Steel Vanilla Random trait directly and use LegacyRandom::from_seed(seed) for nonzero LootTableSeed values.

Like so:
```rs
fn fill_with_rng<R: steel_utils::random::Random>(
    &self,
    origin: BlockPos,
    container: &mut dyn Container,
    rng: &mut R,
) {
    let mut ctx = LootContext::new(rng);
    // ...
}

let mut rng = LegacyRandom::from_seed(self.seed as u64);
self.fill_with_rng(origin, container, &mut rng);
```

Joniii11: the lower code after the curly braces is what where the comment is should be then there and the fn is how you could solve it :)

Joniii11 (changes requested):

coco875:
" I can work through the others now" what do you mean ? you don't need to do everything in one pr, you can just add todo if there is thing who are missing

coco875:
ok I have basic sort, it's also work, will take a look at lithium to see how it optimise it (I will push commit for that), else it look good

coco875:
hi @jackpipergit, thanks for the contribution, by try seing what lithium do like optimisation I found some very good gain, so it's fine if I take the hand on the pr to add optimisation myself ? prefer do it manually to have fully understand it and check it.

pxperr:
> hi @jackpipergit, thanks for the contribution, by try seing what lithium do like optimisation I found some very good gain, so it's fine if I take the hand on the pr to add optimisation myself ? prefer do it manually to have fully understand it and check it.

Okay no problem I guess

steel-core/src/block_entity/randomizable.rs
coco875, line 1: hmm I didn't see the part where it do loot-table

coco875: will like delay the pr until we have made one the fundation for this, sorry

coco875 (commented):
hmm while reading again, I found other weird thing, will wait until we make loot table

JunkyDeveloper:
waiting for loot table so draft :)

PR #428, opened 2026-08-22
feat: Add Vanilla-compatible destructive explosions and TNT

### LINKED ISSUE #575: Finish vanilla compatible explosions and TNT
tracking issue for #428.

#428 adds the shared server explosion system, TNT block behavior, the Primed TNT entity, explosion packets, block and entity effects, and the physics and query work needed for large TNT chains.

full vanilla behavior is still the goal. the PR should stay draft until its damage ownership dependencies and remaining review work are finished.

## blockers

- [ ] #573 creates a fresh `Player` for death and End credits respawns
- [ ] #571 keeps exact `DamageSource` entities alive without `Arc` cycles
- [ ] split the temporary damage source and `EntityReference` refactor out of #428
- [ ] rebase #428 onto #573 and #571
- [ ] resolve the remaining review threads and get a fresh approval

this only tracks the shared explosion system and TNT implemented by #428. other explosion producers can use it through their own issues.

close this when #428 merges.

kdcokenny:
## Type of change

- [x] Block implementation
- [x] Entity implementation
- [x] New feature
- [x] Refactor / code cleanup
- [x] Performance improvement

## Description

Adds server explosions, TNT blocks, Primed TNT, explosion packets, damage attribution, and block/entity effects.

Optimized paths preserve Vanilla ray order, RNG use, callbacks, exposure, drops, knockback, and ownership. Dynamic and plugin behavior falls back to live queries.

## How this was tested

- `cargo test -p steel-core --lib`: 2,514 passed, 1 ignored
- `cargo clippy -r --all-targets --all-features -- -D warnings`
- `cargo fmt --all --check`
- `typos`
- Minecraft 26.2 Fabric client GameTest: all 9 explosion scenarios passed
- 10,000-TNT chain: Steel 3.38 ms/tick, Vanilla 9.44 ms/tick

## Screenshots / logs

Real-client coverage includes ignition, chain reactions, water, damage, creative/spectator behavior, and the TNT gamerule.

## Checklist

- [x] Self-reviewed the diff

## Additional notes

Steel retains Vanilla's air processing and does not use the Lithium/Paper skip-air shortcut.

HotSpot's identity-based ordering cannot be reproduced for rare, unusually large custom explosions. Affected blocks remain the same, but callback order may differ.

Closes #575


ezraclintoc:
Do you know if it works with https://www.youtube.com/watch?v=Ay9ZnCVZhVc
Although this might also rely on other systems as well

kdcokenny:
> Do you know if it works with https://www.youtube.com/watch?v=Ay9ZnCVZhVc Although this might also rely on other systems as well

the goal is full parity with vanilla, so if it doesn't work let us know

ezraclintoc:
> > Do you know if it works with https://www.youtube.com/watch?v=Ay9ZnCVZhVc Although this might also rely on other systems as well
> 
> the goal is full parity with vanilla, so if it doesn't work let us know

I'll have to wait until powered snow get implemented

steel-protocol/src/packets/game/c_explode.rs
```diff
@@ -0,0 +1,220 @@
+use glam::DVec3;
+use steel_macros::{ClientPacket, WriteTo};
+use steel_registry::{
+    packets::play::C_EXPLODE, particle_type::ParticleData, sound_event::SoundEventHolder,
+};
+use thiserror::Error;
+
+/// A block-debris particle and the client-side scaling applied to it.
+#[derive(WriteTo, Clone, Debug)]
+pub struct ExplosionParticleInfo {
```
coco875, line 10: explosion particle aren't just like other particle ?

kdcokenny:
now faster than paper while maintaining full parity with vanilla!!!

victorialondynn:
dupers work omfg
https://drive.google.com/file/d/1XM1eX1oH5y_QjKZnAImhuKkpbNvaDtOJ/view?usp=sharing

kdcokenny:
500 tps duper
https://github.com/user-attachments/assets/e767667d-1339-4d36-98c4-d9a05f9d2b7a

steel-core/src/behavior/block/mod.rs
```diff
@@ -378,6 +378,80 @@ pub trait BlockBehavior: Send + Sync {
...
+        &self,
+        state: BlockStateId,
+        world: &Arc<World>,
+        pos: BlockPos,
+        explosion: &dyn Explosion,
+        on_hit: &mut dyn FnMut(ItemStack, BlockPos),
+    ) {
+        if state.is_air() {
+            return;
+        }
```
coco875, line 404: it's necessary ? because most default block aren't air and we can just overwrite the function for air block no ?

steel-core/src/behavior/block/mod.rs
```diff
@@ -378,6 +378,80 @@ pub trait BlockBehavior: Send + Sync {
...
+        world.set_block(
+            pos,
+            vanilla_blocks::AIR.default_state(),
+            UpdateFlags::UPDATE_ALL,
+        );
+        self.was_exploded(state, world, pos, explosion);
+    }
+
+    /// Called after this block is removed by an explosion.
+    #[expect(unused_variables, reason = "default implementation has no callback")]
```
coco875: why add it and not _ like other ?

steel-core/src/behavior/blocks/redstone/tnt_block.rs
```diff
@@ -0,0 +1,244 @@
...
+        );
+        let has_infinite_materials = player.has_infinite_materials();
+        inv.with_item(|item| {
+            if is_flint_and_steel {
+                item.hurt_and_break(FLINT_AND_STEEL_DAMAGE_PER_USE, has_infinite_materials);
+            } else if !has_infinite_materials {
+                item.shrink(FIRE_CHARGE_ITEMS_PER_USE);
+            }
+        });
+        // TODO: Award the ITEM_USED statistic once Steel has shared statistics support.
```
coco875: we have stat now x)

steel-core/src/chunk/section.rs
coco875, line 1: hmm this file are above 1k, maybe in another pr we should split it

steel-core/src/chunk/section.rs
```diff
@@ -191,6 +194,39 @@ impl<'a> SectionWriteGuard<'a> {
             was_randomly_ticking,
         }
     }
+
+    /// Sets one block while keeping all section-derived metadata current.
+    pub fn set_block_state(
```
coco875, line 202: a bit weird to add the set_block_state until now

coco875 (changes requested):
didn't finish read all

steel-protocol/src/packets/game/c_explode.rs
kdcokenny: No, it's based on the shared ParticleData but additionally carries scaling and speed for entries in the weighted block-particle palette. Replacing it with a normal particle would lose fields required by ClientboundExplodePacket.

steel-core/src/behavior/block/mod.rs
kdcokenny: Not strictly necessary. I mean we could gve every air block a no-op behavior, but they currently share the default behavior. Keeping the check here matches Vanilla and covers all air states.

steel-core/src/behavior/block/mod.rs
kdcokenny: good catch

steel-core/src/chunk/section.rs
kdcokenny: yeah, but we need it so that we don't invalidate the entire cache during normal block writes.

steel-core/src/chunk/section.rs
kdcokenny: ok, not exactly sure how we should split it, but if you have any ideas lmk!

steel-core/src/behavior/blocks/redstone/tnt_block.rs
```diff
@@ -0,0 +1,247 @@
...
+        );
+        true
+    }
+
+    fn send_disabled_message(player: &Player) {
+        player.send_overlay_message(&TextComponent::translated(TranslatedMessage {
+            key: Cow::Borrowed("block.minecraft.tnt.disabled"),
+            fallback: None,
+            args: None,
+        }));
```
Laptop59: Use the `BLOCK_MINECRAFT_TNT_DISABLED` translation constant instead.

steel-core/src/behavior/blocks/redstone/tnt_block.rs
```diff
@@ -0,0 +1,247 @@
...
+        &self,
+        state: BlockStateId,
+        world: &Arc<World>,
+        pos: BlockPos,
+        player: &Player,
+    ) -> BlockStateId {
+        if !player.has_infinite_materials() && state.get_value(&BlockStateProperties::UNSTABLE) {
+            Self::prime(world, pos, None);
+        }
+        state
```
Laptop59: Is there a better way to call `super` here instead of just returning `state`? I know currently the default implementation for `player_will_destroy` just returns `state`, but what if that changes later?

steel-core/src/entity/damage.rs
```diff
@@ -15,9 +16,15 @@ pub struct DamageSource {
...
+    causing_entity_reference: Option<EntityReference>,
     /// The entity that directly dealt the damage (e.g. the projectile itself).
     pub direct_entity_id: Option<i32>,
-    /// Source position (for explosions, etc.).
+    /// Domain-stable live reference for the direct entity when one is available.
+    direct_entity_reference: Option<EntityReference>,
+    /// Last known direct-entity position for sources whose entity can be removed before damage.
+    direct_entity_position: Option<DVec3>,
+    /// Explicit raw source position (for `/damage ... at`, bad respawn points, etc.).
     pub source_position: Option<DVec3>,
```
Laptop59: Surely there's a better way than storing redundant information? Both entity type and entity IDs? Wouldn't a Weak reference be better here?

steel-core/src/behavior/blocks/redstone/tnt_block.rs
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    fn try_add_primed(world: &Arc<World>, entity: Arc<PrimedTntEntity>) -> bool {
+        let entity: SharedEntity = entity;
+        if let Err(error) = world.try_add_entity(entity) {
+            log::debug!("failed to add primed TNT: {error}");
+            return false;
+        }
+        true
+    }
```
coco875: why not make it as a generic function ? to be re-use or it's already in world ?

steel-core/src/behavior/blocks/redstone/tnt_block.rs
coco875: use steel_utils::translations

steel-core/src/command/builtins/damage.rs
```diff
@@ -60,29 +60,21 @@ fn damage(context: &SteelCommandContext<CommandSource>) -> Result<i32, CommandSy
...
-
-        damage_source.direct_entity_id = Some(entity_id);
-        damage_source.causing_entity_id = Some(entity_id);
-
-        // Maybe even the causing entity is known
-        if let Ok(cause) = context.entity("cause") {
-            damage_source.causing_entity_id = Some(cause.id());
-        }
+    if context.argument("entity").is_ok() {
+        let entity = context.entity("entity")?;
```
coco875: why doing an if let like before ?

steel-core/src/command/builtins/damage.rs
```diff
@@ -60,29 +60,21 @@ fn damage(context: &SteelCommandContext<CommandSource>) -> Result<i32, CommandSy
...
+        let cause = if context.argument("cause").is_ok() {
+            Some(context.entity("cause")?)
+        } else {
+            None
+        };
```
coco875, line 69: a map can do the same ?

steel-core/src/entity/base/mod.rs
```diff
@@ -83,29 +83,24 @@ pub(crate) struct EntityPhysicsStateInput {
 
 #[derive(Debug, Default)]
 struct EntityMovementTrace {
-    movement_this_tick: VecDeque<EntityMovement>,
-    final_movements_this_tick: Vec<EntityMovement>,
+    movement_this_tick: SmallVec<[EntityMovement; 1]>,
+    final_movements_this_tick: SmallVec<[EntityMovement; 1]>,
```
coco875, line 90: why use a size of 1 ? because if it need more it will spill into the heap, so a default size a bit bigger wouldn't be better ?

steel-core/src/entity/entities/objects/explosives/primed_tnt/tests.rs
```diff
@@ -0,0 +1,531 @@
...
+const VANILLA_MAX_EXPLOSION_POWER: f32 = 128.0;
+const SERIALIZED_FUSE: i32 = 37;
+const SERIALIZED_EXPLOSION_POWER: f32 = 6.5;
+const FLOWING_WATER_LEVEL: u8 = 4;
+const PLAYER_HURT_EXPERIENCE_TICKS: i32 = 100;
+const EXPERIENCE_ORB_SEARCH_MARGIN: f64 = 2.0;
+const TEST_EXPLOSION_POWER: f32 = 2.0;
+const OVERLAPPING_TNT_CENTER_OFFSET: f64 = 0.25;
+const COLLISION_QUERY_MARGIN: f64 = 1.0;
+const FINAL_FUSE_TICK: i32 = 1;
```
coco875: maybe a bit too much const, like pos, speed

steel-core/src/entity/entities/objects/explosives/primed_tnt/tests.rs
```diff
@@ -0,0 +1,531 @@
...
+use crate::entity::{EntityFluidContact, LivingEntity, change_entity_world, next_entity_id};
+use crate::physics::{CollisionWorld, WorldCollisionProvider};
+use crate::player::ResetReason;
+use crate::portal::{TeleportPostTransition, TeleportTransition};
+use crate::server::worlds::WorldMap;
+use crate::test_support::{
+    TestPlayerBuilder, fresh_test_world, fresh_test_world_in_domain, insert_ready_full_chunk,
+};
+
+const TEST_TNT_POS: BlockPos = BlockPos::new(8, 64, 8);
```
coco875: also a bit weird the initial pos of most test are 8 64 8

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
```diff
@@ -0,0 +1,428 @@
...
+const GRAVITY: f64 = 0.04;
+const AIR_DRAG: f32 = 0.98;
+const INITIAL_HORIZONTAL_SPEED: f64 = 0.02;
+const INITIAL_VERTICAL_SPEED: f32 = 0.2;
+const SHORT_FUSE_RANDOM_RANGE_DIVISOR: i32 = 4;
+const SHORT_FUSE_MINIMUM_DIVISOR: i32 = 8;
+const MIN_SHORT_FUSE_RANDOM_RANGE: i32 = 1;
+const EXPLOSION_HEIGHT_FRACTION: f32 = 0.0625;
+const GROUND_HORIZONTAL_DRAG: f64 = 0.7;
+const GROUND_VERTICAL_BOUNCE: f64 = -0.5;
```
coco875: look like you re-use some const in test

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
```diff
@@ -0,0 +1,428 @@
...
+        }
+        <DefaultExplosionDamageCalculator as ImmutableExplosionBlockCalculator>::should_explode(
+            &DefaultExplosionDamageCalculator,
+            reader,
+            pos,
+            state,
+            power,
+        )
+    }
+}
```
coco875: it shouldn't be in portal impl file ?

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
```diff
@@ -0,0 +1,428 @@
...
+
+// SAFETY: This key is owned by Steel and uniquely identifies `PrimedTntEntity`.
+unsafe impl DowncastType for PrimedTntEntity {
+    const TYPE_KEY: DowncastTypeKey = DowncastTypeKey::new("steel:entity/primed_tnt");
+}
+
+impl PrimedTntEntity {
+    /// Default fuse duration in ticks.
+    pub const DEFAULT_FUSE_TIME: i32 = DEFAULT_FUSE_TIME;
```
coco875: isn't neccessary to have two const who are equal ?

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
```diff
@@ -0,0 +1,428 @@
...
+        position: DVec3,
+        world: &Arc<World>,
+        owner: Option<&dyn Entity>,
+    ) -> Self {
+        let entity = Self::new(entity_type, id, position, Arc::downgrade(world));
+        let angle = world.with_random(Random::next_f64) * TAU;
+        entity.set_velocity(DVec3::new(
+            -angle.sin() * INITIAL_HORIZONTAL_SPEED,
+            f64::from(INITIAL_VERTICAL_SPEED),
+            -angle.cos() * INITIAL_HORIZONTAL_SPEED,
```
coco875, line 196: by memory the behavior of sin and cos on java differ from rust

steel-core/src/entity/entity/mod.rs
```diff
@@ -2131,16 +2212,26 @@ pub trait Entity: EntityEventSource + ErasedType + Send + Sync + 'static {
...
+    /// base-tick-only eye-water history.
+    fn refresh_fluid_contact_with_currents(&self) -> EntityFluidContact {
+        self.scan_and_store_fluid_contact(false, true)
     }
 
     /// Refreshes cached fluid contact with vanilla base-tick eye-water history.
     fn refresh_fluid_contact_for_base_tick(&self) -> EntityFluidContact {
-        self.scan_and_store_fluid_contact(true)
+        self.scan_and_store_fluid_contact(true, true)
     }
```
coco875, line 2227: wouldn't be better to just simplify it and remvoe wrapper ?

steel-core/src/entity/manager/tests/queries_and_order.rs
```diff
@@ -1,7 +1,99 @@
...
 
 use super::*;
 
+struct SourceAwareHardCollisionEntity {
+    base: EntityBase,
+    accepted_source_id: i32,
+    callback_count: AtomicUsize,
+}
+
+impl SourceAwareHardCollisionEntity {
```
coco875: I don't get why you add test for that ? by memory tnt don't have hard collision, only few block have it like shulker

steel-core/src/entity/manager/mod.rs
coco875, line 1: did you compare a bit with lithium or vmp to see if he do something more optimised ? Else it look good

steel-core/src/chunk/section.rs
coco875, line 202: hmm you have add this function but I don't why, because look like before function can already access to set_block_state like in write_tracked_block_batch

steel-core/src/entity/tests/damage.rs
coco875, line 1: hmm a lot of modification that seems unrelated to tnt

steel-core/src/entity/block_effects.rs
```diff
@@ -10,6 +11,37 @@ const CLIP_EPSILON: f64 = 1.0e-7;
...
+        }
+
+        let mut overflow = FxHashSet::default();
+        overflow.reserve(self.inline.len() + 1);
+        overflow.extend(self.inline.drain(..));
+        let inserted = overflow.insert(pos);
+        debug_assert!(inserted, "inline duplicate check must precede the fallback");
+        self.overflow = Some(overflow);
+        true
+    }
```
coco875: a bit weird this hybrid between vec and hashset, why not use an enum and do the convertion when it goes above 16 ? so like that when it's above it only do the hash and not the iteration

steel-core/src/entity/mod.rs
```diff
@@ -278,6 +282,37 @@ enum BlockEffectSegmentResult {
...
 
+/// Allocation-free visited-position set for the short movement segments used
+/// by normal entity ticks, with a hashed fallback for large entities or moves.
+#[derive(Default)]
+struct VisitedBlockPositions {
+    inline: SmallVec<[BlockPos; 8]>,
+    overflow: Option<FxHashSet<BlockPos>>,
+}
+
+impl VisitedBlockPositions {
```
coco875, line 298: isn't the same as MovementVisitedBlocks ?

coco875 (changes requested):
I didn't have finish to read all but it should busy you a bit

steel-core/src/behavior/blocks/redstone/tnt_block.rs
kdcokenny: making it generic here would duplicate the world API

steel-core/src/command/builtins/damage.rs
kdcokenny: good catch, fixed

steel-core/src/command/builtins/damage.rs
kdcokenny: we could use .map(...).transpose()?, but I think that it makes this harder to read and using .ok().map(...) would swallow the entity lookup error, so i kept the explicit branch

steel-core/src/entity/base/mod.rs
kdcokenny: 1 is the common case here since TNT normally records one move. EntityMovement is also pretty large and there are two persistent buffers on every entity, so making both bigger increases every entity's size to avoid an allocation on uncommon multi move ticks. id rather change it if measuring shows thats common

steel-core/src/entity/entities/objects/explosives/primed_tnt/tests.rs
kdcokenny: agreed, cleaned it up

steel-core/src/entity/entities/objects/explosives/primed_tnt/tests.rs
kdcokenny: I did that so that the TNT near its center and inside Full chunk data while it moves. For clarity, i renamed it to TEST_TNT_BLOCK_POS and added a comment

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
kdcokenny: fixed

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
kdcokenny: fixed

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
kdcokenny: well in vanilla this calculator is declared on PrimedTnt, and SulfurCube reuses it. its TNT explosion policy for an entity that used a portal, not portal implementation code, so i kept it here

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
kdcokenny: so what i found online is that they can differ in the last bits, but Java Math itself permits that across platforms. I'm glad you brought this up tho cuz it showed me a that I should use Vanilla's float rounded tau rather than f64 tau

steel-core/src/entity/entity/mod.rs
kdcokenny: i kept the wrappers because they represent different Vanilla update paths. one only refreshes contact, one also applies currents, and the base tick path also advances the eye water state. removing them would leave raw boolean pairs at the call sites and make the wrong combinations easy to use

steel-core/src/entity/manager/tests/queries_and_order.rs
kdcokenny: fair, cleaned up.

steel-core/src/entity/manager/mod.rs
kdcokenny: yeah, i checked those and couldn't find anything else which would meaningfully improve performance while staying true to vanilla

steel-core/src/chunk/section.rs
kdcokenny: the call worked before through DerefMut, but that path now marks the stable air certificate unknown because arbitrary mutable access could change anything. this method avoids that for a known single block write and updates the metadata. i clarified the docs and added a test for it

steel-core/src/entity/tests/damage.rs
kdcokenny: not sure how i missed this b4, good eye

steel-core/src/entity/mod.rs
kdcokenny: nah theyre separate sets with different scopes. one deduplicates positions inside one geometric sweep, while the other deduplicates effects across every movement segment.

steel-core/src/entity/block_effects.rs
kdcokenny: once overflow exists, the first branch returns overflow.insert(pos), so it already only hashes and never checks the vector again. an enum could express the same states differently, there's no extra iteration here... unless u mean something else?

steel-core/src/behavior/blocks/redstone/tnt_block.rs
coco875: then directly use the one of world api ? to avoid doing just a wrapper ?

steel-core/src/entity/base/mod.rs
coco875: have this explanation in code would be good then, to avoid future code dev that increase it or revert it

steel-core/src/entity/entities/objects/explosives/primed_tnt/tests.rs
coco875: was also thinking of it, in other test where it also use that, it make sense to do test at the center of the chunk, but a global const wouldn't be better ? (in another pr of course)

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
coco875: hmm UsedPortalDamageCalculator are a bit weird name, do you have a suggestion of a better name ?

steel-core/src/entity/block_effects.rs
coco875: didn't get attention on overflow, but it still use memory for the SmallVec, so it would allow to compact a bit more

steel-core/src/entity/mod.rs
coco875: maybe precise it in the doc ?

steel-core/src/chunk/section.rs
coco875: we can put test in another file it's around 200 lines but no need to doing it in this pr

steel-core/src/chunk/section.rs
```diff
@@ -234,7 +276,150 @@ pub(crate) struct BlockStateSectionCounts {
...
+const BLOCKS_PER_SECTION: u16 = BlockPalette::VOLUME as u16;
+const STABLE_AIR_OCCUPANCY_ROW_COUNT: usize = BlockPalette::SIZE * BlockPalette::SIZE;
+
+/// Section-local occupancy of every state except Vanilla's three immutable air blocks.
+///
+/// `Unknown` is deliberately treated as occupied. It is used after arbitrary mutable palette
+/// access so an optimization can never hide plugin or future mutation paths that bypass the
+/// tracked setters.
+#[derive(Debug)]
+enum StableAirOccupancy {
```
coco875, line 288: can it impact negatively tnt burried in the ground or in contraption ? so I guess mostly Mixed situation ?

steel-core/src/entity/entities/objects/explosives/primed_tnt/tests.rs
```diff
@@ -0,0 +1,508 @@
...
+use crate::server::worlds::WorldMap;
+use crate::test_support::{
+    TestPlayerBuilder, fresh_test_world, fresh_test_world_in_domain, insert_ready_full_chunk,
+};
+
+// Center the fixture in the only loaded chunk so movement stays in Full chunk data.
+const TEST_TNT_BLOCK_POS: BlockPos = BlockPos::new(8, 64, 8);
+const VANILLA_DEFAULT_FUSE_TIME: i32 = 80;
+const VANILLA_GRAVITY: f64 = 0.04;
+const VANILLA_AIR_DRAG: f32 = 0.98;
```
coco875, line 30: you already define AIR_DRAG in steel-core/src/entity/entities/objects/explosives/primed_tnt.rs

steel-core/src/entity/entities/objects/explosives/primed_tnt/tests.rs
```diff
@@ -0,0 +1,508 @@
...
+use crate::portal::{TeleportPostTransition, TeleportTransition};
+use crate::server::worlds::WorldMap;
+use crate::test_support::{
+    TestPlayerBuilder, fresh_test_world, fresh_test_world_in_domain, insert_ready_full_chunk,
+};
+
+// Center the fixture in the only loaded chunk so movement stays in Full chunk data.
+const TEST_TNT_BLOCK_POS: BlockPos = BlockPos::new(8, 64, 8);
+const VANILLA_DEFAULT_FUSE_TIME: i32 = 80;
+const VANILLA_GRAVITY: f64 = 0.04;
```
coco875, line 29: same

steel-core/src/entity/entities/objects/explosives/primed_tnt/tests.rs
```diff
@@ -0,0 +1,508 @@
...
+const VANILLA_AIR_DRAG: f32 = 0.98;
+const VANILLA_INITIAL_VERTICAL_SPEED: f32 = 0.2;
```
coco875, line 31: same

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
```diff
@@ -0,0 +1,430 @@
...
+use crate::world::explosion::default_explosion_damage_source_with_references;
+use crate::world::{
+    DefaultExplosionDamageCalculator, Explosion, ExplosionBlockReader, ExplosionDamageCalculator,
+    ExplosionInteraction, ExplosionOptions, ImmutableExplosionBlockCalculator, World,
+};
+
+const DEFAULT_EXPLOSION_POWER: f32 = 4.0;
+const MAX_EXPLOSION_POWER: f32 = 128.0;
+const GRAVITY: f64 = 0.04;
+const AIR_DRAG: f32 = 0.98;
```
coco875, line 38: we haven't already global const for that ?

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
```diff
@@ -0,0 +1,430 @@
...
+            &DefaultExplosionDamageCalculator,
+            reader,
+            pos,
+            state,
+            power,
+        )
+    }
+}
+
+static USED_PORTAL_DAMAGE_CALCULATOR: UsedPortalDamageCalculator = UsedPortalDamageCalculator;
```
coco875: hmm maybe HAVE_GONE_THROUGH_PORTAL_DAMAGE_CALCULATOR ? a bit long but in the same spirit

steel-core/src/entity/entities/objects/explosives/primed_tnt/tests.rs
```diff
@@ -0,0 +1,508 @@
...
+use crate::player::ResetReason;
+use crate::portal::{TeleportPostTransition, TeleportTransition};
+use crate::server::worlds::WorldMap;
+use crate::test_support::{
+    TestPlayerBuilder, fresh_test_world, fresh_test_world_in_domain, insert_ready_full_chunk,
+};
+
+// Center the fixture in the only loaded chunk so movement stays in Full chunk data.
+const TEST_TNT_BLOCK_POS: BlockPos = BlockPos::new(8, 64, 8);
+const VANILLA_DEFAULT_FUSE_TIME: i32 = 80;
```
coco875, line 28: in PrimedTntEntity there is also the const

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
```diff
@@ -0,0 +1,430 @@
...
+// SAFETY: This key is owned by Steel and uniquely identifies `PrimedTntEntity`.
+unsafe impl DowncastType for PrimedTntEntity {
+    const TYPE_KEY: DowncastTypeKey = DowncastTypeKey::new("steel:entity/primed_tnt");
+}
+
+impl PrimedTntEntity {
+    /// Default fuse duration in ticks.
+    pub const DEFAULT_FUSE_TIME: i32 = 80;
+    /// Sentinel used by Vanilla callers for an absent fuse.
+    pub const NO_FUSE: i32 = -1;
```
coco875, line 168: hmm why not make it an optional ?

steel-core/src/entity/reference.rs
```diff
@@ -0,0 +1,163 @@
...
+
+/// A persistent entity UUID with a weak cache of its currently loaded entity.
+///
+/// Resolution is restricted to loaded worlds in the requesting world's Steel
+/// domain. The weak cache avoids ownership cycles while preserving Vanilla's
+/// ability to retain a live reference across same-domain world transitions.
+#[derive(Clone)]
+pub struct EntityReference {
+    uuid: Uuid,
+    cached: Arc<SyncMutex<Option<WeakEntity>>>,
```
coco875, line 23: why option ? if you have the uuid you should also have the entity ?

coco875: else why not make the option a the start instead of inside Arc and Mutex ?

steel-core/src/entity/reference.rs
```diff
@@ -0,0 +1,163 @@
...
+            *cached = None;
+        }
+
+        let entity = world.get_entity_in_domain_by_uuid(&self.uuid)?;
+        if entity.is_removed() || !matches_type(entity.as_ref()) {
+            return None;
+        }
+        *self.cached.lock() = Some(Arc::downgrade(&entity));
+        Some(entity)
+    }
```
coco875, line 139: a bit weird the behavior of this function because that mean that we can give an invlid uuid or cache the wrong one

steel-core/src/physics/collision.rs
```diff
@@ -190,8 +195,11 @@ impl BlockCollisionSearchBounds {
         self,
         mut visit: impl FnMut(BlockPos, CollisionCursorType) -> ControlFlow<R>,
     ) -> ControlFlow<R> {
-        for y in self.min_y..=self.max_y {
-            for z in self.min_z..=self.max_z {
+        // Vanilla's Cursor3D advances X first, then Y, then Z. Collision behavior can be
+        // extensible, so retain that callback order even though the final shape set is usually
+        // insensitive to traversal order.
```
coco875, line 200: maybe shorten a bit, like "the order x y z are faster for iteration" ?

steel-core/src/entity/base/mod.rs
kdcokenny: done

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
kdcokenny: I renamed it to NetherPortalPreservingExplosionDamageCalculator

steel-core/src/entity/block_effects.rs
kdcokenny: yeah ur right, i was only looking at the lookup path. i changed both sets to enum states, so we dont keep a separate hash set field next to the SmallVec

steel-core/src/behavior/blocks/redstone/tnt_block/tests.rs
```diff
@@ -0,0 +1,497 @@
...
+use crate::entity::entities::{EnderPearlEntity, FireworkRocketEntity, PigEntity, PrimedTntEntity};
+use crate::entity::{Entity, SharedEntity};
+use crate::player::ResetReason;
+use crate::test_support::{TestPlayerBuilder, fresh_test_world, insert_ready_full_chunk};
+use crate::world::{ExplosionInteraction, ExplosionOptions};
+
+const TEST_TNT_POS: BlockPos = BlockPos::new(8, 64, 8);
+const ONE_BLOCK: f64 = 1.0;
+const TWO_BLOCKS: f64 = 2.0;
+const VANILLA_DEFAULT_FUSE_TIME: i32 = 80;
```
coco875, line 24: it's in the primed entity by memory

steel-core/src/behavior/blocks/redstone/tnt_block/tests.rs
```diff
@@ -0,0 +1,497 @@
...
+use crate::behavior::{BLOCK_BEHAVIORS, init_behaviors};
+use crate::entity::entities::{EnderPearlEntity, FireworkRocketEntity, PigEntity, PrimedTntEntity};
+use crate::entity::{Entity, SharedEntity};
+use crate::player::ResetReason;
+use crate::test_support::{TestPlayerBuilder, fresh_test_world, insert_ready_full_chunk};
+use crate::world::{ExplosionInteraction, ExplosionOptions};
+
+const TEST_TNT_POS: BlockPos = BlockPos::new(8, 64, 8);
+const ONE_BLOCK: f64 = 1.0;
+const TWO_BLOCKS: f64 = 2.0;
```
coco875, line 23: really useful ?

steel-core/src/behavior/blocks/redstone/tnt_block.rs
```diff
@@ -0,0 +1,237 @@
...
+            vanilla_blocks::AIR.default_state(),
+            UpdateFlags::UPDATE_ALL_IMMEDIATE,
+        );
+        let has_infinite_materials = player.has_infinite_materials();
+        let used_item = inv.with_item(|item| {
+            let used_item = item.item();
+            if is_flint_and_steel {
+                item.hurt_and_break(FLINT_AND_STEEL_DAMAGE_PER_USE, has_infinite_materials);
+            } else if !has_infinite_materials {
+                item.shrink(FIRE_CHARGE_ITEMS_PER_USE);
```
coco875, line 204: really necessary the const ? because it seems very frequent to be shrink(1) maybe have function would be better (in another pr)

steel-core/src/behavior/blocks/redstone/tnt_block.rs
```diff
@@ -0,0 +1,237 @@
...
+};
+use crate::entity::projectile::Projectile;
+use crate::entity::{Entity, entities::PrimedTntEntity, next_entity_id};
+use crate::player::Player;
+use crate::world::game_event::GameEventContext;
+use crate::world::{ClipHitResult, Explosion, SignalGetter as _, World};
+
+const PRIMING_SOUND_VOLUME: f32 = 1.0;
+const PRIMING_SOUND_PITCH: f32 = 1.0;
+const FLINT_AND_STEEL_DAMAGE_PER_USE: i32 = 1;
```
coco875, line 31: I would argue the same

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
```diff
@@ -0,0 +1,431 @@
...
+    RemovalReason, SharedEntity,
+};
+use crate::physics::MoverType;
+use crate::world::explosion::default_explosion_damage_source_with_references;
+use crate::world::{
+    DefaultExplosionDamageCalculator, Explosion, ExplosionBlockReader, ExplosionDamageCalculator,
+    ExplosionInteraction, ExplosionOptions, ImmutableExplosionBlockCalculator, World,
+};
+
+const DEFAULT_EXPLOSION_POWER: f32 = 4.0;
```
coco875, line 35: there is also the same const in another test file, steel-core/src/behavior/blocks/redstone/tnt_block/tests.rs I think

steel-core/src/chunk/section.rs
kdcokenny: gameplay wont change. if the checked area contains blocks it falls back to the normal vanilla raycasts.

steel-core/src/entity/entities/objects/explosives/primed_tnt/tests.rs
kdcokenny: thats intentional. these are independent expected Vanilla values.

steel-core/src/entity/entities/objects/explosives/primed_tnt/tests.rs
kdcokenny: thats intentional. these are independent expected Vanilla values.

steel-core/src/entity/entities/objects/explosives/primed_tnt/tests.rs
kdcokenny: thats intentional. these are independent expected Vanilla values.

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
kdcokenny: renamed

steel-core/src/entity/entities/objects/explosives/primed_tnt.rs
kdcokenny: Vanilla stores the fuse as an int and uses -1 as NO_FUSE. an Option would still need converted back to that value for synced data, so I kept Vanilla's representation

steel-core/src/entity/reference.rs
kdcokenny: bc the UUID can exist while the entity isnt loaded, like when it came from saved NBT. the Option is inside the shared mutex so every cloned reference sees the cache get filled or cleared later

steel-core/src/entity/reference.rs
kdcokenny: an unresolved UUID just returns None. it cant cache the wrong entity because both the cache path and world lookup verify the exact UUID

steel-core/src/physics/collision.rs
kdcokenny: this isnt for speed, its because collision callbacks can be extended and their order can be observable. I can shorten it to Keep Vanilla's X, then Y, then Z callback order for extensible collision behavior if you'd like?

steel-core/src/world/explosion/server/block_rays/cache.rs
```diff
@@ -0,0 +1,496 @@
...
+    entries: [ExplosionBlockCacheEntry; BLOCK_CACHE_SIZE],
+}
+
+#[repr(C)]
+#[derive(Clone, Copy)]
+struct DenseExplosionBlockCacheEntry {
+    resistance: f32,
+    state: BlockStateId,
+    flags: u8,
+    _padding: u8,
```
coco875, line 58: why add padding ?

coco875: with repr(C), it should already round

steel-core/src/world/explosion/server/block_rays/java_block_pos_set.rs
coco875, line 1: we didn't have something similar for redstone wire ?

steel-core/src/chunk/section.rs
coco875: I mean in term of performance

steel-core/src/entity/entities/objects/explosives/primed_tnt/tests.rs
coco875: the problem is if vanilla change the number, we have 2000 places to also change, prefer if we centralise constant

steel-core/src/physics/collision.rs
coco875: yes to be more direct

steel-core/src/entity/reference.rs
kdcokenny: while checking this i found an actual issue. `DamageSource` only keeps weak entity refs right now. once a removed TNT or projectile loses its last `Arc`, `direct_entity()` and `causing_entity()` return `None`. vanilla keeps the exact entity alive. I reproduced this with a focused test.

just making `EntityReference` strong isnt safe because self damage or 2 entities damaging each other could create `Arc` cycles that never get freed.

I think the clean fix is to keep normal `EntityReference`s weak, let `DamageSource` hold exact strong refs, and store `last_damage_source` outside the entity in an expiring world owned table. that would preserve vanilla’s 40 tick behavior without cycles.

this is a bigger foundation change, so I wanted to mark it as a known issue and ask what you think before changing the design. does that approach sound right?

coco875 (changes requested):
comment that I have, I let you split the damage refactor in another pr

PR #431, opened 2026-08-22
Fix build.tooHigh message and container drops on block destroy

mellytimes:
## Type of change

- [x] Bug fix

## Description

Two small unrelated fixes found while working through the same file:

1. Placing a block above the world's max build height silently reverted
   the placement with no feedback. Now sends the vanilla `build.tooHigh`
   translated message with the height limit as an argument.

2. Breaking a block with a container block entity (chest, barrel, etc.)
   discarded its stored items entirely instead of dropping them onto
   the ground, since the destroy path only ever dropped the block's
   own loot table, not the container's contents.

## How this was tested

Built with `cargo check -p steel-core` (clean), ran `cargo clippy -r
--all-targets --all-features`, `cargo fmt --all --check`, and `typos`
— all pass. Have not yet tested in-game against a live client;
manual verification steps: place a block above the height limit and
confirm the message appears; place a filled chest, break it, and
confirm items drop onto the ground instead of vanishing.

## Screenshots / logs

N/A

## Checklist

- [x] Self-reviewed the diff

## Additional notes

The container-drops fix is **not full vanilla parity**. In vanilla,
container-content dropping runs unconditionally on any block-entity
removal (`BlockEntity.preRemoveSideEffects`, called from
`LevelChunk.setBlockState`) — independent of whether resources should
drop, so it also fires on explosions, pistons, and creative-mode
breaking. Steel doesn't yet have an equivalent block-entity-removal
hook in its block-update pipeline, so this fix only covers the
`drop_items` path (normal breaking) for now. Flagging this in case a
maintainer wants to guide where that broader hook should live before
extending it further.


steel-core/src/player/game_mode/block_interaction.rs
```diff
@@ -66,7 +66,16 @@ impl Player {
         let world = self.get_world();
 
         if pos.y() >= world.max_build_height() {
-            // TODO: Send "build.tooHigh" message to player
+            self.send_message(
+                &TranslatedMessage {
```
coco875: use steel_utils::translations that already generate it

coco875 (changes requested):

steel-core/src/player/game_mode/block_interaction.rs
Joniii11: i was about to send that too :C

mellytimes: > use steel_utils::translations that already generate it

pushed the fix thanks for looking out for me

steel-core/src/world/level_effects.rs
```diff
@@ -6,7 +6,7 @@ use super::{
     UpdateFlags, World, WorldEntityManager, entity_loot_ref, fluid_state_to_block, level_events,
     vanilla_blocks, vanilla_game_events,
 };
-
+use crate::inventory::lock::{ContainerLockGuard, ContainerRef};
 pub(super) fn sound_is_within_range(
```
coco875, line 11:
```suggestion
use crate::inventory::lock::{ContainerLockGuard, ContainerRef};

pub(super) fn sound_is_within_range(
```
add a new line

coco875 (approved):

JunkyDeveloper (approved):

mellytimes:
thank you for having me be one of the people who work on this piece of great software :>

PR #434, opened 2026-08-22
Feat/pterodactyl egg

itzzfinicalfire:
## Type of change

- [x] Entity implementation
- [x] Bug fix
- [x] Refactor / code cleanup

## Description

Arrow + bow feature — fixed review comments:

- Removed duplicate `PROJECTILE_SHOOT` game event from `BowItem::release_using` (was firing in bow + again in `Projectile::projectile_base_tick` on first tick). Now only `projectile_base_tick` fires it, like vanilla.
- Added `TODO: award ITEM_USED stat` in `release_using` before return.
- Refactored arrow to use `AbstractArrow` trait (`abstract_arrow.rs`) instead of `ArrowEntity` directly using `AbstractArrow` constants. `Pickup` + constants (`BASE_DAMAGE`, `SHAKE_TIME`, `GRAVITY` etc) now live in the trait so `SpectralArrow`/`Trident` can reuse later without copy-paste.

## How this was tested

- `cargo check -p steel-core` clean
- `cargo fmt --check` clean
- `cargo clippy -p steel-core --all-targets` clean
- `cargo test -p steel-core --lib` 2402 passed

## Screenshots / logs

N/A

## Checklist

- [x] Self-reviewed the diff


JunkyDeveloper (changes requested):
We will not maintain here the pterodactyl egg. please remove it

PR #436, opened 2026-08-22
feat(block): add StructureVoidBlock behavior (invisible, tiny shape)

itzzfinicalfire:
## Type of change
- [x] Block implementation

## Description
Add `StructureVoidBlock` — vanilla is invisible, tiny shape, no server logic.

## How this was tested
- `cargo check -p steel-core` clean
- `cargo fmt --check` clean

## Checklist
- [x] Self-reviewed the diff


coco875:
similar to CryingObsidianBlock we have already everything implemented x) so we need more to remove this two

PR #441, opened 2026-08-22
Fix fire block behavior

stuffpicklesays:
## Type of change

- [x] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

Implements several missing behaviours for fire and soul fire blocks

### Changes

    - Adds support checks for FireBlock when block underneath is removed
    - Adds support checks for SoulFireBlock when soul sand/soul soil
    - Removes unsupported fire by replacing it with air
    - Plays an extinguishing sound when a player destroys the fire (note: this sound does NOT play when a fire is removed because it is unsupported in vanilla so it has not been implemented here)
    - Add regression tests for fire and soul fire support checks

## Scope

This PR focuses on behaviour that can be built using steel's current infrastructure

Fire spreading and adjacent-block burning have not been implemented; these all require a flammability system using etractor to gather the data from vanilla, this can be addressed separately

## How this was tested

- Manual test in game
- Added a regression test for unsupported fire
- Ran:
    - cargo check
    - cargo fmt --check
    - cargo test -p steel-core update_shape

## Screenshots / logs

Not applicable

## Checklist

- [x] Self-reviewed the diff

## Additional notes

The behaviour was verified manually in-game

Portal lighting has not been affected by these changes

Flammability and spreading will be handled separately


stuffpicklesays:
Update: Just noticed while testing that Soul Fire does the exact same thing (doesn't disappear after block underneath is broken). I'm going to extend the change to include soul fire blocks, as well as adding a second regression test for soul fire

JunkyDeveloper:
As I know the fire block still has many things missing so maybe can you also do that? so the fire block is fully finished?

stuffpicklesays:
Yeah, its still missing sounds, support checks when blocks are broken on the sides and quite a bit more.
I'll continue working on vanilla fire mechanics and expand the PR
This is my first contribution to SteelMC, so I'm still new to everything, but I'm happy to continue working on the remaining behaviour

JunkyDeveloper:
Yes also I think fire spreading is missing. I think good block to start and learn how steel works

stuffpicklesays:
I think this PR is ready for review, I've kept the scope of this PR limited to just the minor parity improvements, since continuing with fire spreading would require a bigger flammability system. I'm going to look into this separately in a new PR

coco875 (commented):
can you add missing todo for fire ? like randomtick impl, etc ?

coco875 (approved):

JunkyDeveloper (approved):

PR #442, opened 2026-08-22
extract commit hash at compile time

CuzImClicks:
## Type of change

- [x] New feature

## Description

Adds the current git commit hash to lib.rs

## Screenshots / logs

Needed for TextComponents rework:

image

## Checklist

- [x] Self-reviewed the diff


steel-core/build/build.rs
```diff
@@ -70,6 +70,24 @@ pub fn main() {
...
+    println!("cargo:rustc-env=GIT_HASH_SHORT={git_hash_short}");
+    println!("cargo:rerun-if-changed={manifest_dir}/../.git/HEAD");
+    println!("cargo:rerun-if-changed={manifest_dir}/../.git/refs/heads");
+}
+
+fn git_output<const N: usize>(args: [&str; N]) -> String {
+    let output = Command::new("git")
+        .args(args)
+        .output()
+        .expect("git command shouldnt fail");
```
banocean: So we are making git required dependency to compile steel now?

JunkyDeveloper: forgot to install the vm with gui, but yeah this needs to be tested on a pc without git.

I tested already without a repo and it worked

CuzImClicks: The poor guy downloading the zip without git installed trying to compile xD

No but ill make it default to something instead of failing to compile

JunkyDeveloper: Thanks!

banocean: > No but ill make it default to something instead of failing to compile

It should be possible to just make all of those options 😅, also probably some warning when complaining would be good

JunkyDeveloper (approved):

steel-core/build/build.rs
```diff
@@ -70,6 +70,37 @@ pub fn main() {
...
+                Err(format!("exited with {}", output.status))
+            }
+        })
+        .map_or_else(
+            |e| {
+                println!(
+                    "cargo:warning=git {} failed, using \"unknown\": {e}",
+                    args.join(" ")
+                );
+                "unknown".to_owned()
```
banocean, line 100: Why not an option?

teknostom (approved):
LGTM

steel-core/build/build.rs
teknostom: Option just moves the handling elsewhere, this way you can see both the results it can have.

coco875 (commented):
a bit late to make a review but would be good to put the tag for when we do release people can also easily see on what version they are

steel-core/build/build.rs
banocean: But this handling makes it very unclear, the name doesbt suggest at all its an option filled with unknown instead of being a reliable hash

PR #448, opened 2026-08-22
Added recursion limit to fix warning in steel-core

Sumutiu:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

Fixes warning when compiling steel-core caused in pub fn queue_player_join

## How this was tested
Compiled the project after to check if the warning has been mitigated.

## Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ rustc-ice-*
 .idea
 
 .DS_STORE
+/.vs
--- a/steel-core/src/lib.rs
+++ b/steel-core/src/lib.rs
@@ -3,6 +3,7 @@
 //! The core library for the Steel Minecraft server. Handles everything related to the PLAY state.
 
 #![feature(try_as_dyn)]
+#![recursion_limit = "256"]
 
 use crate::chunk::chunk_map::ChunkMap;
 


coco875:
normally all warning are treated like error, so can you explain how you get this warning ?

.gitignore
```diff
@@ -40,3 +40,4 @@ rustc-ice-*
 .idea
 
 .DS_STORE
+/.vs
```
banocean, line 43: I think you added this line by accident 😅

banocean (changes requested):

.gitignore
Sumutiu: > I think you added this line by accident 😅

I am using Visual Studio 2026. Had to ignore VS generated garbage... 😅

Sumutiu:
> normally all warning are treated like error, so can you explain how you get this warning ?

"warning: overflow evaluating the requirement `{async block@steel-core\src\server\player_admission.rs:108:22: 108:32}: std::marker::Send`
--> steel-core\src\server\player_admission.rs:108:9" - From Visual Studio 2026

This warning is still accepted by the compiler but will eventually be an hard error in the future. For now, increasing the recursion limit fixes it.

JunkyDeveloper:
No one of us is using VS I think that is the case here

coco875:
also I think ideally we should remove the recursion, but don't see why it happen

Sumutiu:
> also I think ideally we should remove the recursion, but don't see why it happen

For the future, I completely agree. However this is a minimal fix without touching the code, for now.

Sumutiu:
> No one of us is using VS I think that is the case here

I have been a fan of Visual Basic from 2008 onwards. I have the 2026 installed for work related, so I use it for stuff like this as well. No point to install another extra software if this works.

coco875:
hmm there is a way to also get this warning on linux to be able to reproduce ?

coco875:
because if it's really only VS and it don't appears on any ide it would hard to fix in the future without knowing how to reproduce

Sumutiu:
> hmm there is a way to also get this warning on linux to be able to reproduce ?

I am actually surprised this warning doesn't occur on Linux. Perhaps try to update Rust Nightly?

coco875:
ok after update rust nightly to "2026-08-22" I get the warning, I guess time to bump the nightly to get all new error
image

coco875:
I will do that in a separate pr then

coco875:
I have merge your pr on https://github.com/Steel-Foundation/SteelMC/pull/449 to make proper bump of nightly and not have a fix without the original warning

PR #450, opened 2026-08-22
Native mod loading integration.

Sumutiu:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

Added `steel-loader` crate and integrated native mod loading into SteelMC. Supports `steel.mod.json` metadata format inspired from Fabric Loader, dependency topological sorting, and dynamic dynamic-library lifecycle hooks.

This is still an early implementation and needs more work for hooks, etc.

To create and run a native mod for SteelMC using `steel-loader`:

### 1. Create a New Rust Library Project
Create a new Rust library project for your mod:

```bash
cargo new --lib my_steel_mod
cd my_steel_mod
```

### 2. Configure Cargo.toml
Configure the library to compile as a C-compatible dynamic library (`cdylib`):

```toml
[package]
name = "my_steel_mod"
version = "0.1.0"
edition = "2024"

[lib]
crate-type = ["cdylib"]

[dependencies]
# Optional: Add steel-loader as a dependency if referencing ModContext directly
```

### 3. Create the Mod Manifest (`steel.mod.json`)
Create a `steel.mod.json` file in the root of your mod project:

```json
{
    "schemaVersion": 1,
    "id": "my_steel_mod",
    "version": "0.1.0",
    "name": "My Steel Mod",
    "description": "An example native mod for SteelMC",
    "authors": ["Your Name"],
    "license": "MIT",
    "depends": {
        "steel": ">=0.15.0"
    },
    "entrypoints": {
        "main": ["my_mod_init"],
        "shutdown": ["my_mod_shutdown"]
    }
}
```

### 4. Write the Mod Entrypoints (`src/lib.rs`)
Export the initialization and shutdown entrypoint C functions specified in your manifest:

```rust
use std::ffi::c_void;

/// Opaque context passed by SteelMC server upon loading.
#[repr(C)]
pub struct ModContext {
    pub server_ptr: *mut c_void,
}

/// Initialization entrypoint called when SteelMC starts up.
///
/// Return 0 for success, or a non-zero integer for failure.
#[unsafe(no_mangle)]
pub unsafe extern "C" fn my_mod_init(ctx: *const ModContext) -> i32 {
    println!("[MySteelMod] Initialized successfully!");
    0
}

/// Shutdown entrypoint called when SteelMC stops.
#[unsafe(no_mangle)]
pub unsafe extern "C" fn my_mod_shutdown(ctx: *const ModContext) {
    println!("[MySteelMod] Shutting down!");
}
```

### 5. Build the Mod Shared Library
Build your mod in release mode:

```bash
cargo build --release
```

This generates a compiled shared library file in `target/release/`:
- **Linux:** `libmy_steel_mod.so`
- **macOS:** `libmy_steel_mod.dylib`
- **Windows:** `my_steel_mod.dll`

### 6. Install and Run the Mod in SteelMC
Copy the compiled shared library (`.so`, `.dylib`, or `.dll`) along with `steel.mod.json` into the `mods/` directory of your SteelMC server:

```text
steel-server/
├── config/
├── mods/
│   └── my_steel_mod/
│       ├── steel.mod.json
│       └── libmy_steel_mod.so
└── steel
```

*(Alternatively, you can package them together into a `.zip` or `.jar` archive in `mods/`, or place the `.so`/`.dylib`/`.dll` directly in `mods/`.)*

Run SteelMC. Upon startup, `steel-loader` will discover the mod, resolve dependencies, load the library, and invoke `my_mod_init`. When the server closes, it will automatically call `my_mod_shutdown`.

## Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ rustc-ice-*
 .idea
 
 .DS_STORE
+/.vs
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1384,6 +1384,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3eaf3ede3fee6db1a4c2ee091bf8a8b4dccdc6d17f656fb07896ee72867612f2"
 
 [[package]]
+name = "libloading"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "754ca22de805bb5744484a5b151a9e1a8e837d5dc232c2d7d8c2e3492edc8b60"
+dependencies = [
+ "cfg-if",
+ "windows-link",
+]
+
+[[package]]
 name = "libmimalloc-sys"
 version = "0.1.49"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2420,6 +2430,10 @@ name = "semver"
 version = "1.0.28"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8a7852d02fc848982e0c167ef163aaff9cd91dc640ba85e263cb1ce46fae51cd"
+dependencies = [
+ "serde",
+ "serde_core",
+]
 
 [[package]]
 name = "serde"
@@ -2681,6 +2695,7 @@ dependencies = [
  "serde",
  "serde_json",
  "steel-core",
+ "steel-loader",
  "steel-login",
  "steel-registry",
  "steel-utils",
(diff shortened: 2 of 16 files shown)


coco875:
before doing this ask on discord because we have arleady idea on how to do it and plan. We will not do plugin yet because there is still manything to implement so add plugin will only slowdown and make harder/annoying when we need to refactor move thing around. We also have on what lib to use to help doing it and other, so come on discord before doing something this much important

PR #457, opened 2026-08-23
feat: implement chests, trapped chests, and container viewer tracking

kokspudding:
> **Stacked on #460**, which adds the loot-table foundation this needs. Review that one first; the diff here is only the chest work on top.

## What

Implements chests and trapped chests, and adds the `ContainerOpenersCounter`
foundation they share with barrels.

### `ContainerOpenersCounter`

Mirrors vanilla: the first viewer opens the container, the last one closes it,
`CONTAINER_OPEN`/`CONTAINER_CLOSE` game events fire on those transitions, and a
scheduled recheck every 5 ticks drops viewers that stopped looking without
closing their menu. `BlockEntity::start_open`/`stop_open` (vanilla
`Container.startOpen`/`stopOpen`) are driven from `MenuKind::on_open`/`removed`.

### Chests

- `ChestBlockEntity`: 27 slots, the lid block event, and vanilla's double-chest
  sound placement — the left half stays silent, the right half offsets toward
  its partner so one sound plays between them.
- `ChestBlock`: double-chest formation on placement, connect and disconnect in
  `updateShape`, the blocked-chest check, waterlogging, `isPathfindable`, the
  scheduled-tick recheck, and a comparator signal spanning both halves.
- `TrappedChestBlock`: redstone signal from the viewer count, republished to its
  own and the below neighbours when the count changes.

### Barrel

Wired to the same counter, which resolves the existing TODO in `BarrelBlock`
covering its `open` state, open/close sounds, and recheck.

### Loot tables

Chests carry a worldgen loot table through `ContainerLoot` from #460 and unpack
it where vanilla does: opening the menu and reading the comparator output. A
structure chest now fills with its actual loot instead of opening empty.

## Notable deviations

A double chest maps its halves to **two consecutive menu sections** rather than
wrapping them in a `CompoundContainer`. Steel keeps each container behind its own
mutex, so a compound wrapper would nest locks; two sections keep each half
independently lockable while preserving vanilla's slot order and
`stillValid` semantics.

The viewer recheck walks the world's players instead of an entity AABB, because
Steel keeps players outside the entity manager. Vanilla's range and
own-container checks are applied unchanged. Copper golems, vanilla's other
container user, are not implemented yet.

Both are commented at their call sites.

## Not implemented

- **Cat-on-chest check** (`isCatSittingOnChest`) — awaits the Cat entity. The
  conducting-block-above check is implemented.
- **`PiglinAi.angerNearbyPiglins`** and the container `LockCode`, neither of
  which exists in Steel yet (the barrel already lacked both).

## Tests

New: double-chest geometry on placement and in `updateShape` (4), and the
combined comparator signal across both halves (1).

Three chunk tests used `CHEST` as an example of a block with a block entity but
no registered factory. Since chests now have one, they use `BEACON` instead,
keeping their original intent.

`cargo test`, `cargo clippy -r --all-targets --all-features` and
`cargo fmt --all --check` all pass.

Manually verified on a local server: placing a chest, opening it, and moving
items in and out.


JunkyDeveloper:
The loot table needs to be done before chests, so I will put this PR in draft until that is done

coco875:
ok:
* ask on discord before doing pr, we haven't done chest because of loot table and we want to do it properly
* you clearly didn't use the template that we made
* I seen claude as co-author, are you in the loop or it's vibe coded ? Because it feel that this pr have been automated

kokspudding:
Sorry for that, will ask on discord first next time.
Also going to use the template

Code is created with the help of Claude, but I am in the loop

coco875:
there is too many proof that it's manage by an ai so until you didn't exprimate on discord I will just close your pr.

PR #460, opened 2026-08-23
feat: fill containers from loot tables on vanilla's random source

kokspudding:
Prerequisite for #457, which a review correctly asked to land first.

## The gap

Steel's loot pipeline evaluated on `rand::Rng`. Its draws do not match Java's
`RandomSource`, so a stored loot seed could not reproduce vanilla contents —
the gap `LootTable.getRandomItems`' TODO already described. Two call sites
(brushable blocks, entity death loot) even seeded `StdRng` from a vanilla seed,
which looks deterministic but produces different items than vanilla.

## Porting the pipeline

All 17 draw sites now run on `steel_utils::random::Random`:

| site | vanilla |
|---|---|
| weighted entry selection | `random.nextInt(totalWeight)` |
| random-chance conditions, explosion survival | `random.nextFloat()` |
| `UniformGenerator` | `Mth.nextInt` / `Mth.nextFloat` |
| `BinomialDistributionGenerator` | `nextFloat() < p` per trial |
| `ApplyBonusCount` ore/uniform/binomial | `nextInt` per formula |
| instrument and tag-expand picks | `nextInt(size)` |

Two helpers are new because the distinction is load-bearing:

- `next_int_between` / `next_float_between` mirror `Mth.nextInt` / `Mth.nextFloat`,
  which return `min` **without drawing** when `min >= max`.
  `Random::next_i32_between` always draws, so using it on a degenerate range —
  which `shuffleAndSplitItems` hits whenever one stack is left — would shift
  every later value in a seeded sequence.
- `random::shuffle` mirrors `Util.shuffle`.

Seeded call sites now use `LegacyRandom::from_seed`, matching
`RandomSource.create(seed)`. Unseeded ones take a `LegacyRandom` from entropy:
vanilla would use the level's shared source, which Steel does not have yet, so
the algorithms are vanilla's and only the seed is incidental.

`Random::next_i32_between` itself is left alone — worldgen relies on it heavily
and its parity suite covers those paths, so tightening it belongs in its own
change.

## Filling containers

`LootTable.fill`: roll the table, spread the results over the container's
shuffled empty slots, split stacks through `shuffleAndSplitItems`.

`ContainerLoot` holds `RandomizableContainer`'s pending table and its NBT
round-trip; `unpack_container_loot` rolls and clears it.

**Deviation:** vanilla routes every container accessor (`getItem`, `isEmpty`,
`removeItem`, …) through `unpackLootTable`. Steel's `Container` methods are
storage-local by contract and must not reach the registry, so callers unpack at
the boundaries where contents become observable — opening the menu and reading
the comparator output. Both are covered for barrels here; chests follow in #457.

## Fixes along the way

- **Barrels dropped their `LootTable` NBT on load**, so a worldgen-placed barrel
  silently lost its loot. They now round-trip and unpack it.
- **The explosion-survival test was measuring the wrong thing.** It sampled one
  draw from each of 100 sequential seeds; vanilla's LCG barely moves its first
  `nextFloat` across those (all ≈0.731), so it measured seeding, not
  distribution, and only passed because `StdRng` scrambles its seed. It now
  samples repeatedly from one generator.

## Tests

- A degenerate `Mth.nextInt` range consumes no randomness.
- `shuffle` consumes exactly one draw per element above the first.
- Splitting a 64-stack across 8 slots preserves the total count.
- The same seed fills a dungeon chest identically, with content.

`cargo test`, `cargo clippy -r --all-targets --all-features` and
`cargo fmt --all --check` pass.

## Still open

Table-level `random_sequence` entries fall back to the caller's source; vanilla
resolves them through a `RandomSequences` registry Steel does not have yet.


coco875:
it's the second time https://github.com/Steel-Foundation/SteelMC/pull/457, the last time you said: 
> Sorry for that, will ask on discord first next time.
> Also going to use the template
> 
> Code is created with the help of Claude, but I am in the loop

You still didn't ask on how we want do it or why we use ThreadRng instead of Legacy random.
You still didn't use the template.

so it only confirm that you aren't in the loop...

PR #462, opened 2026-08-23
feat: implement shears item

Coderinsightmaven:
## Type of change

- [ ] Block implementation
- [x] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

Implements vanilla `ShearsItem.useOn` for Minecraft 26.2:

- Crops growing plant heads that are not at max age to age 25
- Plays the growing-plant crop sound and emits `BLOCK_CHANGE`
- Damages the shears by 1
- Registers the behavior for the vanilla shears item

Kelp, cave vines, weeping vines, and twisting vines identify as growing plant heads through `BlockBehavior::is_growing_plant_head()`, matching vanilla `instanceof GrowingPlantHeadBlock`.

## How this was tested

Minecraft 26.2

- `cargo test -p steel-core --lib shears`
- `cargo test -p steel-core --lib max_age_state_is_age_25`
- In-game: used shears on kelp, cave vines, weeping vines, and twisting vines. The growing tip stopped growing and shears took 1 durability. Using shears on an already cropped tip or a plant body did nothing.

## Checklist

- [x] Self-reviewed the diff

## Additional notes

This PR does **not** implement vanilla `ShearsItem.mineBlock`.

Steel still uses the shared mining tool-damage path (`destroy_time != 0`). The shears-only override (skip durability on `#minecraft:fire`, still take durability on 0-hardness non-fire blocks) is left for a follow-up that adds `ItemBehavior::mine_block` dispatch.

Sheep shearing and pumpkin carving already live on the entity/block and are unchanged.


steel-core/src/behavior/blocks/vegetation/growing_plant_head_block.rs
```diff
@@ -287,4 +303,16 @@ mod tests {
 
         assert_eq!(state, vanilla_blocks::CAVE_VINES_PLANT.default_state());
     }
+
+    #[test]
+    fn max_age_state_is_age_25() {
+        init_vanilla_registry();
+
+        let young = vanilla_blocks::KELP.default_state().set_value(AGE, 7);
```
kdcokenny: `7` is an unexplained age here. Could this use a named value such as `let non_max_age = MAX_AGE - 1;` so the test intent and validity are clear?

steel-core/src/behavior/items/shears.rs
```diff
@@ -0,0 +1,99 @@
...
+        let pos = context.hit_result.block_pos;
+        let state = context.world.get_block_state(pos);
+        if !BLOCK_BEHAVIORS
+            .get_behavior(state.get_block())
+            .is_growing_plant_head()
+            || GrowingPlantHeadBlock::is_max_age(state)
+        {
+            return InteractionResult::Pass;
+        }
+
```
kdcokenny, line 50: Add a todo comment for triggering ITEM_USED_ON_BLOCK advancement.

steel-core/src/behavior/blocks/vegetation/growing_plant_head_block.rs
```diff
@@ -110,6 +111,18 @@ impl GrowingPlantHeadBlock {
             .set_value(AGE, rng.random_range(0..25))
```
kdcokenny: forgot to use MAX_AGE here

steel-core/src/behavior/blocks/vegetation/growing_plant_head_block.rs
```diff
@@ -287,4 +303,16 @@ mod tests {
 
         assert_eq!(state, vanilla_blocks::CAVE_VINES_PLANT.default_state());
     }
+
+    #[test]
+    fn max_age_state_is_age_25() {
```
kdcokenny: i would personally prefer to just check if it's consistent with our constant value than a hardcoded "25".

steel-core/src/behavior/block/mod.rs
```diff
@@ -610,6 +610,11 @@ pub trait BlockBehavior: Send + Sync {
         false
     }
 
+    /// Returns whether this behavior implements vanilla `GrowingPlantHeadBlock`.
+    fn is_growing_plant_head(&self) -> bool {
```
kdcokenny: This boolean encodes the `GrowingPlantHeadBlock` class check without exposing its associated behavior. `ShearsItem` consequently calls `GrowingPlantHeadBlock` AGE helpers on every implementation returning `true`; an incompatible plugin implementation could attempt invalid property access. Could this be a capability such as `as_growing_plant_head() -> Option<&dyn GrowingPlantHeadBehavior>` exposing `is_max_age` and `get_max_age_state`? That also follows Steel capability-trait guidance for class-hierarchy checks.

steel-core/src/behavior/items/shears.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use steel_registry::{init_vanilla_registry, vanilla_blocks};
+
+    use crate::behavior::{BLOCK_BEHAVIORS, init_behaviors};
+
+    #[test]
+    fn growing_plant_heads_are_identified() {
```
kdcokenny: These assertions only verify the new classification flag and never execute `ShearsItem::use_on`. Please add focused behavior tests for a successful crop and an already-max/pass case so regressions in the sound, block update/game event, and durability path are covered.

kdcokenny (changes requested):
check for more instances of 25 being hardcoded as well

coco875:
for shear can you explain how it block the growth ? because I didn't find nothing much related to that in the pr

steel-core/src/behavior/blocks/vegetation/growing_plant_head_block.rs
Coderinsightmaven: Resolved in 6d384de62. I removed the helper-only test that used the unexplained `7`; the end-to-end crop test now names the setup explicitly as `let non_max_age = MAX_AGE - 1`.

steel-core/src/behavior/items/shears.rs
Coderinsightmaven: Added in 6d384de62: `TODO: Trigger ITEM_USED_ON_BLOCK once advancement criteria dispatch is implemented.`

steel-core/src/behavior/blocks/vegetation/growing_plant_head_block.rs
Coderinsightmaven: Fixed in 6d384de62. `get_head_state` now uses `0..MAX_AGE`, and I also replaced the remaining growth/bonemeal `25` literals in this module with `MAX_AGE`.

steel-core/src/behavior/blocks/vegetation/growing_plant_head_block.rs
Coderinsightmaven: Resolved in 6d384de62. The direct helper test was replaced by `ShearsItem::use_on` behavior tests; the successful case asserts the resulting AGE against `MAX_AGE`.

steel-core/src/behavior/block/mod.rs
Coderinsightmaven: Implemented in 6d384de62. `BlockBehavior` now exposes `as_growing_plant_head() -> Option<&dyn GrowingPlantHeadBehavior>`, and that capability owns `is_max_age` plus `get_max_age_state`. The concrete head adapters return their embedded base behavior, so `ShearsItem` no longer assumes an AGE property from a boolean claim.

steel-core/src/behavior/items/shears.rs
Coderinsightmaven: Added in 6d384de62. The tests now execute `ShearsItem::use_on` against a real test world. The success case verifies AGE becomes `MAX_AGE`, the exact growing-plant-crop sound packet, `BLOCK_CHANGE`, and one durability damage; the max-age case verifies Pass with no state, sound, event, or durability effects.

Coderinsightmaven:
The shears stop further growth by changing the clicked growing-head block state to `AGE = 25` (the vanilla maximum). `GrowingPlantHeadBlock::random_tick` only extends the plant while `AGE < MAX_AGE`, so after cropping that head no longer creates the next segment. This matches vanilla 26.2: `ShearsItem.useOn` calls `getMaxAgeState`, and `GrowingPlantHeadBlock.randomTick` gates growth on age being below 25. I also pushed 6d384de62 to make that shared `MAX_AGE` relationship explicit and cover it end-to-end in tests.

kdcokenny (approved):

steel-core/src/behavior/items/shears.rs
```diff
@@ -0,0 +1,297 @@
...
+        );
+        let new_state = plant.get_max_age_state(state);
+        context
+            .world
+            .set_block(pos, new_state, UpdateFlags::UPDATE_ALL);
+        context.world.game_event(
+            &vanilla_game_events::BLOCK_CHANGE,
+            pos,
+            &GameEventContext::new(Some(context.player), Some(new_state)),
+        );
```
coco875: hmm would prefer if we put it under a clear function call block_age for example, to make it more clear that it's how we block plant grow

coco875 (changes requested):

steel-core/src/behavior/items/shears.rs
Coderinsightmaven: Addressed in 2c3ca92fb: `ShearsItem::use_on` now calls `ShearsItem::block_age`, which owns the max-age state update and `BLOCK_CHANGE` game event. That keeps the crop path named after the growth-blocking step instead of inlining `get_max_age_state` + `set_block` in `use_on`.

coco875 (dismissed):

JunkyDeveloper (dismissed):

JunkyDeveloper (dismissed):

coco875 (dismissed):

JunkyDeveloper (approved):

coco875 (approved):

PR #467, opened 2026-08-24
feat: implement vanilla turtle egg block behavior

darkstarworks:
## Type of change

- [x] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Implements `TurtleEggBlock` for MC 26.2.

Turtle eggs are placed in clusters of one to four in a single block space. While sitting on sand they crack through three stages and then hatch. Most living entities that step or fall onto them trample them.

* `EGGS` (1–4) and `HATCH` (0–2) states, including stacking eggs onto an existing cluster.
* Random-tick cracking and hatching, only on sand (`#sand`).
* Trampling via step-on (1/100) and fall-on (1/3). Sneaking avoids it, turtles and bats never trample, non-players require `mobGriefing`, and zombies always trample when landing.
* Breaking a cluster removes one egg at a time.
* Collision shape changes with the egg count.

## How this was tested

Added tests for cracking through to hatching on sand, no hatch progression off sand, and the egg-count-dependent collision shape.

`cargo fmt --all --check`, `cargo clippy -r --all-targets --all-features`, and `typos` all pass.

## Screenshots / logs

Not applicable. Block behavior is covered by the unit tests above.

## Checklist

- [x] Self-reviewed the diff

## Additional notes

A couple of implementation details are worth sanity-checking during review. These are not firm proposals.

* **Hatch chance.** In 26.2, hatching uses the `turtle_egg_hatch_chance` environment attribute (base `0.002`, rising to `1.0` during the pre-dawn window via a `maximum` float modifier). Steel's timeline sampler in `world::environment` only handles multiply/replace modifiers and exposes sky light and sun angle, so it cannot resolve this attribute yet. I've reproduced the vanilla curve directly from the overworld day clock. If a more general approach is preferred, the follow-up is to add the `maximum` modifier and expose a public environment-attribute getter.
* **Zombie-family trampling.** Vanilla excludes `instanceof Zombie` from fall-on trampling, which also covers husks, drowned, zombie villagers, and zombified piglins. None of those exist in Steel yet, so the current single-type check is effectively exact; it should widen to the whole zombie family once they're added.

Deferred: spawning baby turtles when an egg hatches is stubbed until the turtle entity lands (#490), and will be wired up in a follow-up.


JunkyDeveloper:
Please use the template because it will add now the labels to your PR

darkstarworks:
done. thank you

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
```diff
@@ -0,0 +1,419 @@
...
+            .get_block_state(pos.below())
+            .get_block()
+            .has_tag(&BlockTag::SAND)
+    }
+
+    /// Rolls whether an egg should advance its cracking stage this random tick.
+    ///
+    /// DESIGN NOTE (sanity check for us, not a proposal we are attached to):
+    /// In 26.2 this value comes from the `gameplay/turtle_egg_hatch_chance`
+    /// environment attribute: a base of 0.002, raised to 1.0 by a day-timeline
```
coco875: somone already report that we don't have environmenet atrribute, I think keep it like that but add a todo would be better

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
```diff
@@ -0,0 +1,419 @@
...
+                pos,
+                state.set_value(EGGS, eggs - 1),
+                UpdateFlags::UPDATE_CLIENTS,
+            );
+            world.game_event(
+                &vanilla_game_events::BLOCK_DESTROY,
+                pos,
+                &GameEventContext::default(),
+            );
+            world.level_event(2001, pos, i32::from(state.0), None);
```
coco875: create a const for 2001

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
```diff
@@ -0,0 +1,419 @@
...
+        &self,
+        _state: BlockStateId,
+        world: &Arc<World>,
+        pos: BlockPos,
+        _old_state: BlockStateId,
+        _moved_by_piston: bool,
+    ) {
+        // Play the "placed on sand" effect (level event 2012, data 15).
+        if Self::on_sand(world, pos) {
+            world.level_event(2012, pos, 15, None);
```
coco875: same

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
```diff
@@ -0,0 +1,419 @@
...
+            world.remove_block(pos, false);
+            world.game_event(
+                &vanilla_game_events::BLOCK_DESTROY,
+                pos,
+                &GameEventContext::default(),
+            );
+
+            let eggs = state.get_value(EGGS);
+            for _ in 0..eggs {
+                world.level_event(2001, pos, i32::from(state.0), None);
```
coco875: same

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
```diff
@@ -0,0 +1,419 @@
...
+        if let Some(entity) = context.source_entity()
+            && entity.entity_type() != &vanilla_entities::ZOMBIE
+        {
+            // DESIGN NOTE (sanity check): vanilla excludes `instanceof Zombie`,
+            // which also covers husks, drowned, zombie villagers, and zombified
+            // piglins because they extend Zombie. None of those entities exist in
+            // Steel yet, so this single-type check is currently exact; when the
+            // zombie family is added, this should exclude the whole set (a shared
+            // "is a zombie" predicate or an entity-type tag) so falling zombies do
+            // not double up with their seek-and-trample AI.
```
coco875: a todo would be better

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
```diff
@@ -0,0 +1,419 @@
...
+        pos: BlockPos,
+        state: BlockStateId,
+        _block_entity: Option<&SharedBlockEntity>,
+        _tool: &ItemStack,
+    ) {
+        // Vanilla calls super.playerDestroy (loot/stats) and then decreaseEggs so a
+        // cluster is broken one egg at a time. Steel's break pipeline
+        // (game_mode::block_breaking::destroy_block) already removes the block
+        // before invoking player_destroy, exactly like vanilla, so decrease_eggs
+        // re-places the cluster with one fewer egg when more than one remained.
```
coco875, line 324: it's necessary ?

coco875 (changes requested):

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
darkstarworks: Kept the inline curve and added a `// TODO(environment-attributes)` marker.

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
darkstarworks: Steel already has named level-event constants, so I used `level_events::PARTICLES_DESTROY_BLOCK`, plus `encode_block_state_data` for the data arg.

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
darkstarworks: Now using `level_events::PARTICLES_TURTLE_EGG_PLACEMENT`.

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
darkstarworks: Same, switched to `level_events::PARTICLES_DESTROY_BLOCK`.

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
darkstarworks: Reworded it to a `// TODO(zombie-family)` marker.

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
darkstarworks: It reproduces vanilla `playerDestroy` calling `decreaseEggs`, so breaking a cluster of 2 to 4 eggs removes one egg (re-places EGGS-1) instead of the whole stack. Steel's break pipeline removes the block and then calls `player_destroy` with the old state (block_breaking.rs:438), so `decrease_eggs` re-places the cluster with one fewer egg. Without it, breaking would destroy all the eggs at once. Happy to switch if there is a cleaner hook for count-decrementing blocks.

JunkyDeveloper (dismissed):

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
```diff
@@ -0,0 +1,426 @@
...
+    /// environment attribute: base 0.002, raised to 1.0 by a day-timeline
+    /// `maximum` modifier during the pre-dawn window (ticks 21062 to 21904).
+    /// Steel's timeline sampler in `world::environment` only handles
+    /// `multiply`/replace modifiers and exposes sky light and sun angle, so it
+    /// cannot resolve this attribute yet; the turtle curve is reproduced inline
+    /// off the overworld day clock.
+    // TODO(environment-attributes): replace this inline curve with a single
+    // attribute lookup once the timeline sampler learns the `maximum` modifier
+    // and exposes a public environment-attribute getter.
+    fn should_update_hatch_level(world: &Arc<World>) -> bool {
```
kdcokenny, line 101: this bypasses the timeline set for the current dimension. `should_update_hatch_level` always reads the `OVERWORLD` clock, and Steel gives every world that clock. when it is between 21062 and 21905, an egg in the nether or end advances one hatch stage on every random tick it receives.

vanilla resolves `TURTLE_EGG_HATCH_CHANCE` through the environment attributes for the current level. only `#minecraft:in_overworld` includes the day timeline with the 1.0 modifier. `#minecraft:in_nether` and `#minecraft:in_end` do not, so eggs there can still hatch, but only with the default 0.002 chance per random tick.

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
```diff
@@ -0,0 +1,426 @@
...
+        let eggs = state.get_value(EGGS);
+        if eggs <= 1 {
+            world.destroy_block(pos, false);
+        } else {
+            world.set_block(
+                pos,
+                state.set_value(EGGS, eggs - 1),
+                UpdateFlags::UPDATE_CLIENTS,
+            );
+            world.game_event(
```
kdcokenny, line 171: vanilla passes `GameEvent.Context.of(state)` for this event, and does the same for crack and hatch. using the default context drops the affected state. Steel already has `GameEventContext::new(None, Some(state))`, can we use that at all three call sites?

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
```diff
@@ -0,0 +1,426 @@
...
+        if eggs == 1 {
+            SHAPE_SINGLE
+        } else {
+            SHAPE_MULTIPLE
+        }
+    }
+}
+
+impl BlockBehavior for TurtleEggBlock {
+    fn get_collision_shape(
```
kdcokenny: these shapes are already in the generated turtle egg block data, and the default `get_collision_shape` reads them. copying them here gives us a second source of truth, plus the test below only checks the copy against itself. can we remove the constants, override, and shape test?

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
```diff
@@ -0,0 +1,426 @@
...
+        &self,
+        _state: BlockStateId,
+        world: &Arc<World>,
+        pos: BlockPos,
+        _old_state: BlockStateId,
+        _moved_by_piston: bool,
+    ) {
+        // Play the "placed on sand" particle effect (data is the particle count).
+        if Self::on_sand(world, pos) {
+            world.level_event(level_events::PARTICLES_TURTLE_EGG_PLACEMENT, pos, 15, None);
```
kdcokenny: seems like a magic number to me curious what @coco875 thinks

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
```diff
@@ -0,0 +1,426 @@
...
+
+        chance > 0.0 && rand::random::<f32>() < chance
+    }
+
+    /// Vanilla `TurtleEggBlock.canDestroyEgg`: which entities are able to trample
+    /// an egg. Turtles and bats never do, and only living entities can; players
+    /// always may, other mobs only when `mobGriefing` is enabled.
+    fn can_destroy_egg(world: &Arc<World>, entity: &dyn Entity) -> bool {
+        let entity_type = entity.entity_type();
+        if entity_type == &vanilla_entities::TURTLE || entity_type == &vanilla_entities::BAT {
```
coco875, line 123: Why bat ?

coco875 (dismissed):
Just one question else everything seems fine

kdcokenny (changes requested):

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
darkstarworks: Because bats don't trample turtle eggs in vanilla

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
darkstarworks: You are right. Now gates on `world.dimension_type.timelines == Some("#minecraft:in_overworld")`, so only the overworld (and caves variant) apply the boost and the nether/end stay at the `0.002` base, matching how vanilla resolves the attribute per dimension. Than kyou

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
darkstarworks: Agreed and switched all three (`crack`, `hatch`, and the break in `decrease_eggs`) to `GameEventContext::new(None, Some(state))` so that they carry the affected state, like vanilla `GameEvent.Context.of(state)`.

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
darkstarworks: You're right; The generated static shape already depends on the egg count, and the default `get_collision_shape` uses it. I've removed the override, the `SHAPE_*` constants, and the redundant shape test.

steel-core/src/behavior/blocks/vegetation/turtle_egg_block.rs
darkstarworks: Fair point. Renamed to `TURTLE_EGG_PLACEMENT_PARTICLE_COUNT` instead of a bare `15`.

kdcokenny:
oh no @coco875 is gonna be upset that you rebased 😭

kdcokenny (approved):
🔥 🔥 🔥

coco875 (approved):

JunkyDeveloper (approved):

PR #471, opened 2026-08-24
chore: award some stats x2

GreenedDev:
## Type of change

- [ ] Block implementation
- [x] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [x] Chore / tooling

## Description
awards stats in items.

## Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/steel-core/src/behavior/items/bucket.rs
+++ b/steel-core/src/behavior/items/bucket.rs
@@ -22,6 +22,7 @@ use steel_registry::fluid::FluidState;
 use steel_registry::item_stack::ItemStack;
 use steel_registry::level_events;
 use steel_registry::sound_events;
+use steel_registry::stat::vanilla_stat_types;
 use steel_registry::vanilla_blocks;
 use steel_registry::vanilla_fluids;
 use steel_registry::vanilla_game_events;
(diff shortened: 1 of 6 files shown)


JunkyDeveloper (dismissed):

JunkyDeveloper:
removed for ender eye because didn't found at vanilla and added bucket second if statement

Laptop59 (changes requested):
What if the item gets fully consumed before the stat gets awarded? Then won't it award ITEM_USED for AIR?

GreenedDev:
> removed for ender eye because didn't found at vanilla and added bucket second if statement

?
image

GreenedDev:
> What if the item gets fully consumed before the stat gets awarded? Then won't it award ITEM_USED for AIR?

vanilla consumes first and then awards
image

JunkyDeveloper:
> > removed for ender eye because didn't found at vanilla and added bucket second if statement
> 
> ? image

Your method is use and you added the award to use_on so different method if I didn't made a mistake

Laptop59:
> > What if the item gets fully consumed before the stat gets awarded? Then won't it award ITEM_USED for AIR?
> 
> vanilla consumes first and then awards
> image
> 

True, but `this` is the item whose behavior is called, so it won't change with `consume`, hence the item should be fetched before shrinking because we don't have that here.

coco875 (approved):

JunkyDeveloper (approved):

PR #473, opened 2026-08-24
perf: Cut encrypted outbound batch time by up to 73%

kdcokenny:
## Type of change

- [x] Refactor / code cleanup
- [x] Performance improvement

## Description

Removes bottlenecks from encrypted play-state output without changing the packet bytes or order a Vanilla client receives.

- Encrypts reusable byte groups instead of passing ciphertext to Tokio one byte at a time.
- Gives the play sender sole ownership of its writer, removing the per-packet lock.
- Sends bundles and chunk batches as one ordered group, so unrelated packets cannot appear inside them.
- Flushes after 32 packets or 256 KiB, then gives other connections a turn.
- Keeps queued packet order during graceful disconnects while retaining one bounded close deadline for stalled clients.
- Drops an incomplete encrypted stream instead of reusing invalid packet or cipher state.

Single packets still go out immediately, and batching never waits for future packets.

Release-mode encrypted loopback A/B results used fixed Linux CPU affinity, 20 samples, three-second measurements, and both benchmark orders:

| Workload | Old behavior | Optimized | Result |
| --- | ---: | ---: | ---: |
| 8 small packets | 19.48 µs | 5.17 µs | 73.5% faster |
| 868-packet synthetic burst | 4.51 ms | 2.82 ms | 37.5% faster |
| 9-chunk group | 1.72 ms | 1.69 ms | No regression (~2% faster) |
| Small-packet encryption | 1.46 ms | 1.12 ms | 23.3% faster |
| Mixed-size encryption | 3.80 ms | 2.94 ms | 22.5% faster |

In two paired 256-sample runs with another connection busy, batching reduced both median and 95th-percentile control-packet delay. The old path's tail varied widely, so this PR does not claim a fixed latency percentage.

## How this was tested

Parity tests cover continuous CFB8 output, partial writes, pending flushes, absolute close deadlines, FIFO disconnect ordering, batch contiguity, writer handoff, and login encryption transitions.

All Steel Core, Steel Login, and Steel Protocol tests passed. The change was also validated with a Vanilla Minecraft 26.2 client on the latest Windows release.

## Screenshots / logs

Not applicable.

## Checklist

- [x] Self-reviewed the diff


steel-core/src/player/connection/java.rs
```diff
@@ -826,42 +850,85 @@ impl JavaConnection {
...
+                        let closing_writes = {
+                            let write_result = Self::write_queued(network_writer, queued_write);
+                            tokio::pin!(write_result);
+                            select! {
+                                biased;
+                                () = close.as_mut() => {
+                                    let Some(closing_writes) = Self::close_and_take_writes_through_disconnect(sender_recv) else {
+                                        break;
+                                    };
+                                    if let Err(err) = write_result.as_mut().await {
```
Joniii11: I think disconnect can stuck here if client stop reading. After close happen this still await the current write without any timeout and after this also write queued packets before disconnect. If socket have backpressure and never become writable sender cannot finish the test also show this because after connection.close() sender is still not finished until writer get manually released

Should this have some timeout or bound and then hard close/ drop remaining queue so client cannot keep connection cleanup waiting forever?

Joniii11 (changes requested):

steel-core/src/player/connection/java.rs
Joniii11: kenny says that in #370 this should be fixed

banocean: > kenny says that in #370 this should be fixed

So for now this PR is blocked?

kdcokenny: > > kenny says that in #370 this should be fixed
> 
> So for now this PR is blocked?

no, not really. it's a pre-existing bug. not due to this pr

kdcokenny: Fixed by reconciling #370 with the owned-writer sender. One absolute graceful-close deadline now covers queue scanning, the exact in-flight write, the queued FIFO prefix, and the disconnect packet. If that deadline expires or a write fails, the encoder/socket is dropped and never reused; player cleanup remains independent of transport completion. Added deterministic partial-write, pending-flush, timeout, and lifecycle regression tests.

MarshY2k:
[macro_refactor.txt](https://github.com/user-attachments/files/31493213/macro_refactor.txt)

Joniii11 (approved):
lgtm

steel-core/src/behavior/blocks/decoration/jukebox_tests.rs
```diff
@@ -68,6 +68,10 @@ impl NetworkConnection for RecordingConnection {
         self.packets.lock().push(packet);
     }
 
+    fn send_encoded_batch(&self, packets: Vec<EncodedPacket>) {
+        self.packets.lock().extend(packets);
+    }
+
     fn send_encoded_bundle(&self, packets: Vec<EncodedPacket>) {
         self.packets.lock().extend(packets);
     }
```
coco875, line 77: what are the difference ?

steel-core/src/player/connection/mod.rs
```diff
@@ -127,6 +138,10 @@ impl NetworkConnection for Box<dyn NetworkConnection> {
         (**self).send_encoded(packet);
     }
 
+    fn send_encoded_batch(&self, packets: Vec<EncodedPacket>) {
+        (**self).send_encoded_batch(packets);
+    }
+
     fn send_encoded_bundle(&self, packets: Vec<EncodedPacket>) {
         (**self).send_encoded_bundle(packets);
     }
```
coco875, line 147: look duplicate

steel-core/src/player/player_inventory/tests.rs
```diff
@@ -768,6 +768,10 @@ impl NetworkConnection for LockProbeConnection {
         self.record_if_armed();
     }
 
+    fn send_encoded_batch(&self, _packets: Vec<EncodedPacket>) {
+        self.record_if_armed();
+    }
+
     fn send_encoded_bundle(&self, _packets: Vec<EncodedPacket>) {
         self.record_if_armed();
     }
```
coco875, line 777: look duplicate

steel-core/src/player/tests.rs
```diff
@@ -184,6 +184,10 @@ impl NetworkConnection for RecordingConnection {
         self.sent_packets.lock().push(packet);
     }
 
+    fn send_encoded_batch(&self, packets: Vec<EncodedPacket>) {
+        self.sent_packets.lock().extend(packets);
+    }
+
     fn send_encoded_bundle(&self, packets: Vec<EncodedPacket>) {
         self.sent_packets.lock().extend(packets);
     }
```
coco875, line 193: look duplicate

steel-core/src/server/tests.rs
```diff
@@ -85,6 +85,10 @@ impl NetworkConnection for TestConnection {
         self.sent_packets.lock().push(packet);
     }
 
+    fn send_encoded_batch(&self, packets: Vec<EncodedPacket>) {
+        self.sent_packets.lock().extend(packets);
+    }
+
     fn send_encoded_bundle(&self, packets: Vec<EncodedPacket>) {
         self.sent_packets.lock().extend(packets);
     }
```
coco875, line 94: look duplicate

steel-core/src/server/tests.rs
```diff
@@ -120,6 +124,10 @@ impl NetworkConnection for RecordingConnection {
         self.packets.lock().push(packet);
     }
 
+    fn send_encoded_batch(&self, packets: Vec<EncodedPacket>) {
+        self.packets.lock().extend(packets);
+    }
+
     fn send_encoded_bundle(&self, packets: Vec<EncodedPacket>) {
         self.packets.lock().extend(packets);
     }
```
coco875, line 133: look duplicate

steel-protocol/benches/outbound_transport.rs
coco875, line 1: it's no the same as steel-core/benches/outbound_transport.rs ?

coco875 (changes requested):

coco875:
hmm look like that your batch overlap with bundle that was made before and https://github.com/Steel-Foundation/SteelMC/pull/545 also try something similar ?

JunkyDeveloper:

PR #474, opened 2026-08-24
chore(entity): add `BlockAttachedEntity` trait and update some entities to use it

SomeYellowGuy:
## Type of change

- [x] Entity implementation
- [x] Refactor / code cleanup
- [x] Chore / tooling

## Description

This PR adds the intermediate trait `BlockAttachedEntity` (which extends `Entity`) for Vanilla's block-attached entities. It also refactors the leash knot and item frame to implement this trait.

It also adds a justified TODO to the item frame entity (it's technically not complete).

## How this was tested

The leash knot and item frame work as expected (like before). Teleporting a leash knot also works like Vanilla.

## Screenshots / logs
Not applicable.

## Checklist

- [x] Self-reviewed the diff

## Additional notes

I'm not sure if what I did is the best way to add this trait, so please let me know if there's anything wrong.


steel-core/src/entity/mod.rs
```diff
@@ -1353,7 +1354,6 @@ pub fn leashables_leashed_to_holder_in_area_near_position(
         })
     })
 }
-
```
banocean, line 1388: I think this change got here by accident

steel-core/src/entity/entities/objects/display_ui/item_frame.rs
```diff
@@ -128,24 +136,30 @@ impl ItemFrameEntity {
             .set(direction);
         self.base
             .set_rotation(Self::rotation_for_direction(direction));
-        self.recalculate_position();
+        self.recalculate_position_or_warn();
     }
 
-    fn recalculate_position(&self) {
-        let block_pos = *self.block_pos.lock();
+    fn recalculate_position(&self) -> Result<(), EntityMoveError> {
```
banocean: I have a question if anyone has time to answer, when do we name stuff with `try_` prefix and when no. It seems to be different everywhere and I struggle with finding any rule it could be done by 😓

SomeYellowGuy: I should add `try_`, actually. Thanks for pointing that out!

SomeYellowGuy: Usually, `try` is used for functions that return `Result`s or `Option`s for handling errors

steel-utils/src/types/position.rs
```diff
@@ -528,6 +528,16 @@ impl BlockPos {
...
     }
+
+    /// Returns the squared distance between this `BlockPos` (at the low corner)
+    /// and `pos`.
+    #[must_use]
+    pub fn distance_sqr(self, pos: DVec3) -> f64 {
+        let dx = f64::from(self.x()) - pos.x;
+        let dy = f64::from(self.y()) - pos.y;
+        let dz = f64::from(self.z()) - pos.z;
+        dx * dx + dy * dy + dz * dz
```
JunkyDeveloper: What is the difference to this?
```
fn block_center_distance_sqr(pos: BlockPos, target: DVec3) -> f64 {
    let (x, y, z) = pos.get_center();
    DVec3::new(x, y, z).distance_squared(target)
}
```
because it feels like it is equal, so should be equal

JunkyDeveloper (dismissed):

steel-utils/src/types/position.rs
SomeYellowGuy: > block_center_distance_sqr

While you're technically wrong (I want the distance between the corners of both positions), I found `block_pos_distance_sqr`, which seems to be exactly what I need.

JunkyDeveloper: Nice, yeah I only searched square roots and good you found something fitting :)

coco875 (dismissed):

suprohub (approved):

steel-core/src/entity/block_attached_entity.rs
```diff
@@ -0,0 +1,206 @@
...
+                log::warn!("Block-attached entity at invalid position: {block_pos:?}");
+            }
+        }
+    }
+
+    fn refresh_dimensions_block_attached_entity(&self) {}
+
+    fn try_set_position_block_attached_entity(&self, pos: DVec3) -> Result<(), EntityMoveError> {
+        self.block_attached_entity_base()
+            .set_pos(BlockPos(pos.as_ivec3()));
```
teknostom: This is wrong. It should be 
            .set_pos(BlockPos::containing(pos.x, pos.y, pos.z));

.as_ivec3 rounds towards 0, containing floors. ivec3 is wrong for all coords containing a negative coord.

teknostom (changes requested):

coco875 (approved):

steel-core/src/entity/block_attached_entity.rs
```diff
@@ -0,0 +1,206 @@
...
+            NbtTag::IntArray(vec![block_pos.x(), block_pos.y(), block_pos.z()]),
+        );
+    }
+
+    fn load_block_attached_entity(&self, nbt: BorrowedNbtCompoundView<'_, '_>) {
+        if let Some(block_pos_vec) = nbt.int_array("block_pos")
+            && block_pos_vec.len() == 3
+        {
+            let block_pos = BlockPos::new(block_pos_vec[0], block_pos_vec[1], block_pos_vec[2]);
+            self.block_attached_entity_base().set_pos(block_pos);
```
CuzImClicks: set pos after validation and return early if check fails

CuzImClicks (changes requested):

PR #478, opened 2026-08-24
perf(worldgen): cache parsed vanilla structure templates

carabistouflette:
## Type of change

- [x] Performance improvement

## Description

Vanilla structure templates (used for jigsaw pieces in trial chambers, ancient cities, villages, fossils, and configured template features) were being decompressed from bundled gzip NBT and parsed from scratch on **every single placed piece**.

Since bundled vanilla templates are immutable after load, this PR parses them at most once into an `Arc<StructureTemplate>` process-wide cache (`steel-core/src/worldgen/template/loading.rs`). Placements now share the cached template via `load_vanilla()`.

Non-vanilla/plugin-owned paths (`load_gzip_nbt`) remain unshared and owned.

### Test suite
```bash
cargo test -p steel-core --lib worldgen::template
cargo test -p steel-core --lib worldgen::structure
```

### Benchmark reproduction
Build with the DHAT heap profiler on a pinned seed:
```bash
```
Check `Total: ... bytes` in DHAT logs after server shutdown.

## Screenshots / logs

|---|---|---|---|
| **Total Allocations** | 3.378 GB | **3.207 GB** | **-171 MB (-5.1%)** |
| **Allocated Blocks** | 8.25M blocks | **7.88M blocks** | **-378K blocks (-4.6%)** |

## Checklist

- [x] Self-reviewed the diff

## Additional notes

- Zero gameplay / worldgen determinism change (structures generate identically).


Joniii11 (approved):
i like it!

steel-core/src/worldgen/template/loading.rs
```diff
@@ -1,11 +1,33 @@
...
-        Self::load_gzip_nbt(registry, bytes, &key.to_string())
+    pub(crate) fn load_vanilla(registry: &Registry, key: &Identifier) -> Result<Arc<Self>, String> {
+        if let Some(template) = VANILLA_TEMPLATE_CACHE.read().get(key) {
+            return Ok(Arc::clone(template));
+        }
+
+        let mut cache = VANILLA_TEMPLATE_CACHE.write();
+        if let Some(template) = cache.get(key) {
+            return Ok(Arc::clone(template));
+        }
```
coco875, line 24: why doing two time get ?

coco875 (changes requested):

steel-core/src/worldgen/template/loading.rs
coco875, line 24:
```suggestion
        let mut cache = VANILLA_TEMPLATE_CACHE.write();
        if let Some(template) = cache.get(key) {
            return Ok(Arc::clone(template));
        }
```

coco875: ok after discussion with parking_lot you can lock as read, do the look up and if it fail use upgradable_read to switch to read without another look up

coco875 (approved):

JunkyDeveloper (approved):

PR #479, opened 2026-08-24
perf(worldgen): sparse per-chunk cache for ore bulk section access

carabistouflette:
## Type of change

- [x] Performance improvement

## Results

- ~24 % less memory allocated during worldgen (DHAT, 1089 chunks)
- Faster, not slower: -6.6 % pregen time, -6.8 % CPU, -7.1 % peak memory (median of 30 runs each vs PR base)
- Vanilla parity: 0 mismatched chunks (2500 chunks x 3 dimensions, every stage)

## Description

`WorldGenBulkSectionAccess` pre-allocated a dense 17x17 chunk cache (~6.9 KB, zero-initialized) for every ore placement attempt, ~105 times per chunk (measured: 114,327 constructions across the 1089-chunk DHAT census below; ore placements sum their `count`s, e.g. 90 attempts for `ore_iron_upper`). This PR replaces it with a sparse `FxHashMap<(i32, i32), CachedWorldGenChunk>` that stays empty until a chunk is actually touched, while keeping the same out-of-bounds panics and the verified-status upgrade (single `entry()` lookup).

For completeness: the old radius is 8 (from the `Features` step's direct dependencies). My original 529-slot claim mixed that up with the full dependency halo (11) seen in pregen logs. The reviewer was right.

## Benchmark (reproducible)

```bash
cargo build --profile profiling
```

Compared merge-base `bfd6882`, this PR `38e27a3`, and master `36156e9`: fixed seed, 4225 chunks per run, 12 s cooldown, interleaved run order (on a hybrid laptop, run order and thermals swing single runs by +/-8 %, so this reports medians over 30 runs/side). Times come from the server's own `Spawn area prepared` log line.

| Metric | merge-base | This PR | Delta |
|---|---|---|---|
| Pregen wall (s) | 13.68 | 12.78 | -6.6 % |
| CPU time (s) | 56.05 | 52.27 | -6.8 % |
| Max RSS (MB) | 984.8 | 914.8 | -7.1 % |

Master ~ merge-base here (13.35 vs 13.55 s median), expected since nothing since the merge-base touches chunk scheduling. So the improvement is from this PR, and the added `entry()` lookup does not cost anything measurable.

## DHAT heap census (1089 chunks, seed `steelmc-perf`)

| Metric | Before | After | Delta |
|---|---|---|---|
| Total allocated | 3.378 GB | 2.585 GB | -23.5 % |
| Allocated blocks | 8.25M | 8.09M | -2.0 % |

## Vanilla parity

`cargo test -p steel-core --lib --release chunk_stage_hashes -- --ignored` -> pass, 0 mismatched chunks. 2500 chunks per dimension (overworld, nether, end) match vanilla at every generation stage, including features, so the changed ore path is covered.

## Checklist

- [x] Self-reviewed the diff

## Notes

- While benchmarking I hit an unrelated, pre-existing pregen crash (~2 in ~75 runs, also happens on the merge-base), filed as #561.


steel-core/src/worldgen/region.rs
```diff
@@ -1294,59 +1292,49 @@ impl<'region, 'world, 'profile> WorldGenBulkSectionAccess<'region, 'world, 'prof
...
+                    chunk: chunk.chunk,
+                    verified_status: status,
+                });
+            return cached;
+        }
+        if let Some(cached) = self.chunks.get(&cache_key)
+            && status > cached.verified_status
+        {
             self.with_ore_profile(OreFeatureStats::record_chunk_status_upgrade);
             let _ = self.region.chunk(chunk_x, chunk_z, status);
```
Joniii11: Are we missing cached.verified_status = status here? Master updated it after status validation, Without it every later access with same higher status does the upgrade check again

steel-core/src/worldgen/region.rs
```diff
@@ -1294,59 +1292,49 @@ impl<'region, 'world, 'profile> WorldGenBulkSectionAccess<'region, 'world, 'prof
...
-        let Some(cache_index) = self.chunk_cache_index(chunk_x, chunk_z) else {
+        let Some(cache_key) = self.chunk_cache_key(chunk_x, chunk_z) else {
             panic!(
                 "Worldgen bulk section requested chunk ({chunk_x}, {chunk_z}) outside the region cache centered on ({}, {})",
                 self.region.center.0.x, self.region.center.0.y
             );
         };
 
-        let cache_needs_insert = self.chunks.get(cache_index).is_none_or(Option::is_none);
+        let cache_needs_insert = !self.chunks.contains_key(&cache_key);
```
Joniii11: On cache hit we currently probe the hashmap 3 times (contains_key, get, get). Could we use one get_mut/entry lookup here? It would also make updating verified_status after validation straightforward

Joniii11 (changes requested):
couldn't test if it is still 100% vanilla par but this is what i found

steel-core/src/worldgen/region.rs
carabistouflette: that was dropped in the rewrite. The upgrade path now writes cached.verified_status = status after validation

steel-core/src/worldgen/region.rs
carabistouflette: one entry() lookup now. `Occupied` checks and upgrades in place and writes the status back, `Vacant` inserts.

Joniii11:
Please tell me if the vanilla parity test says 100% vanilla par and did not find mismatched chunks

steel-core/src/worldgen/region.rs
```diff
@@ -905,16 +910,10 @@ impl<'region, 'world, 'profile> WorldGenBulkSectionAccess<'region, 'world, 'prof
...
-        let chunk_cache_len =
-            usize::try_from(chunk_cache_size.saturating_mul(chunk_cache_size)).unwrap_or(0);
-        let chunks = (0..chunk_cache_len).map(|_| None).collect();
-
         Self {
             region,
-            chunk_cache_radius,
-            chunks,
+            chunk_cache_radius: region.chunk_cache_radius,
+            chunks: FxHashMap::default(),
```
Joniii11: The PR cachesize calculation appears wrong for the features path. the radius is 8 so the old cache is 17×17 = 289 entries not 529 and uh  DHAT also proves fewer allocated bytes not faster worldgen while this adds an FxHashMap::entry() lookup to the hot path.

please tell me in a comment or whatever that this is indeed faster and make it reproducible or like wallclock/cpu base bs mastee branch

Joniii11 (changes requested):

steel-core/src/worldgen/region.rs
carabistouflette: Good catch on both:

- Cache size: fixed. The radius is 8 (the `Features` step's direct deps), so the old cache is 17^2 = 289 slots (~6.9 KB) per construction, not 529. The "11" I used was the full dependency halo from the pregen logs. Description corrected.
- "faster?": measured it. 30 runs/side vs merge-base (4225 chunks, fixed seed, interleaved so laptop thermals do not skew it): pregen -6.6 % wall, -6.8 % CPU, -7.1 % RAM. Master ~ merge-base, so the win is from this PR, and the `entry()` lookup costs nothing measurable. Reproduce: `cargo build --profile profiling && env PREGEN_SIZE=65 target/profiling/steel`. Full numbers in the description.

Vanilla parity also verified below

carabistouflette:
Yes, 100 % vanilla parity, 0 mismatched chunks.

`cargo test -p steel-core --lib --release chunk_stage_hashes -- --ignored` on this branch: pass. All 2500 chunks x 3 dimensions (overworld, nether, end) match vanilla at every stage, including features, so the changed ore path is covered.

coco875 (dismissed):
I have confirm the performance improvement, to vulgarize why it increase performance, before we use an array and over evaluate the place needed to have an O(1) lookup but not all index was use. Also to calculate this index we make a kind of relatif index generator. So now instead we use an hashmap to be more compact while retain the O(1) lookup and simplify the index look up and like element are closer more cache hit, so gain in memory and memory latency and cpu

coco875 (dismissed):

coco875:
I have also re-run test just and it case and it pass

CuzImClicks:
we can also just use a FxSmallMap since 91.6% of caches ever only have <=2 elements or at most 6 in them. this allows us to have zero allocations and we dont need the custom hasher etc.

ore_cache_occupancy


```rust
const BULK_SECTION_INLINE_CHUNKS: usize = 8;

type BulkSectionChunkMap<'region> =
    FxSmallMap<BULK_SECTION_INLINE_CHUNKS, i64, CachedWorldGenChunk<'region>>;
```

JunkyDeveloper:
@CuzImClicks only 1 review is needed if you depend on it please finish the review

CuzImClicks:
@JunkyDeveloper i requested changes on this pr and they are still current

coco875 (approved):

CuzImClicks (approved):

PR #480, opened 2026-08-24
perf(worldgen): batch sculk spread reads into one section lock

carabistouflette:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [x] Performance improvement
- [ ] Chore / tooling

## Description

`Chunk::get_block_state` acquires a section read lock on every query, and the sculk checks issue several of them for adjacent blocks. This PR adds `WorldGenRegion::block_states_for::<N>`, which fetches multiple positions under one lock per section. The spread predicate and `can_sculk_spread_from` now use it, dropping up to seven lock acquisitions per check to one.

## How this was tested

- `cargo test -p steel-core --lib worldgen`
- `cargo test -p steel-core --release --lib chunk_stage_hashes -- --ignored` (passes)
- Criterion on builds differing only in the changed code:

| check | before | after |
| spread predicate read pair | 30.4 ns | 17.0 ns |
| `can_sculk_spread_from` | 287 ns | 133 ns |
| full patch placement (64 origins) | 41.7 ms | 40.3 ms |

- I tried running benchmarks e2e, but to be honest for such micro optimization it's not the proper way to do so. The full pregen stays within noise either way

- `PREGEN_SIZE=65` smoke test, zero panics

### Reproducing

The benchmark stack (instrumentation + criterion wrappers, no gameplay changes) is on the `bench/pr480-numbers` branch in this fork. Check out `c76447527` for the before side and the branch tip `875d407db` for the after side, then on an idle machine:

```bash
cargo bench -p steel-core --bench worldgen --features benchmark-support -- --save-baseline <name> sculk
```

For pregen: `cargo build --profile profiling`, pin `seed = "42"` in `config/worlds.toml`, then run `PREGEN_SIZE=65 target/profiling/steel` seven times per stack with `saves/` wiped in between and compare the `Spawn area prepared` timings. Run with `STEEL_FEATURE_READ_PROFILE=1` to get the read and placement counters.

## Screenshots / logs

```
[Info] Spawn area prepared: 4225 chunks in 10.41s (405.8 chunks/s)
[Info] Started Steel Server
```

## Checklist

- [x] Self-reviewed the diff

- `WorldGenRegion` (steel-core/src/worldgen/region.rs)
- `FeatureDecorationRunner` sculk patch checks (steel-core/src/worldgen/feature/features/sculk_patch.rs)

## Additional notes

- Out-of-height positions resolve to air; out-of-dependency positions panic like `block_state`.
- Zero gameplay change: identical placement counts per seed, vanilla chunk stage hashes unchanged.


Joniii11:
Do we have before/after timing for this? Current hot call only has 2 positionsbut we always build sort group them even when they are in different sections and no lock is saved. Maybe a same section fast path + normal reads otherwise would be cheaper?

Joniii11:
Also please prove to us that this still results in 100% vanilla parity for worldgen!

Joniii11:
and also is it safe to reeuse attach_state here? cause features can run parallel and write radius is 1 so another feature task can change this block between checks

carabistouflette:
> Do we have before/after timing for this? Current hot call only has 2 positionsbut we always build sort group them even when they are in different sections and no lock is saved. Maybe a same section fast path + normal reads otherwise would be cheaper?

Added a same-section fast path in new version

carabistouflette:
> Also please prove to us that this still results in 100% vanilla parity for worldgen!

chunk_stage_hashes passes on this branch

carabistouflette:
> and also is it safe to reeuse attach_state here? cause features can run parallel and write radius is 1 so another feature task can change this block between checks

It is not reused across placements. Both states are read fresh per spread attempt and consumed inside that one predicate call. The old code had the same interleaving window between its two locked reads, the batched read just makes the pair consistent with each other

coco875 (commented):
I think it would be good to try find other spot where this optimisation can also benefit

coco875:
also describe how you confirm that it improve performance

JunkyDeveloper:
I let claude run all your improvments PRs: https://claude.ai/code/artifact/6319dc53-cae9-4dfe-96b8-6f7a9ba7e20d
and for this PR it says 0.x% and for me that is kind of noise

So please provide numbers, so I can test again otherwise I would close this PR

JunkyDeveloper:
2 people found no improvements please reopen if you present benchmark numbers

carabistouflette:
Numbers, measured on two builds identical except for the changed read pair.

The spread check reads two adjacent blocks and used to acquire the section lock three times, because the attach position was read twice. Batched into one call: 30.4 to 17.0 ns per check. A full sculk patch placement over 64 origins gets about 4% faster (41.7 to 40.3 ms).

End to end it does not show up: about 230k reads move into batched calls, which is 1 or 2 ms on a 10 second pregen, and seven pregen runs per stack land within noise.

So the per call gain is real, it just cannot move a whole pregen on its own. Follow up: the batching now also covers `can_sculk_spread_from`, see my next comment.

carabistouflette:
Vanilla parity is also checked so it's all fine

carabistouflette:
The batching now also covers `can_sculk_spread_from`: one `block_states_for` call for the origin and its six neighbors instead of up to seven locks. Decision logic unchanged.

Back to back on identical builds, the check drops from 287 ns to 133 ns. Placement and pregen stay unchanged since it runs once per placement.

PR #481, opened 2026-08-24
perf(light): lease propagation queues from a thread-local pool

carabistouflette:
## Type of change

- [x] Performance improvement

## Description

Every block and sky light propagation step constructed fresh `PackedLightPropagationQueues` — two `Vec`s with 4,096 entries each (~64 KiB) per chunk pass.

This PR introduces `PooledPackedLightQueues`, an RAII guard that leases empty queues from a thread-local pool. When dropped, the queues are cleared and returned to the pool (queues exceeding a 128K-entry ceiling are discarded rather than pinned to worker threads).

The 9 production light propagation entry points now lease pooled buffers with zero allocation churn; queue bodies remain untouched via `Deref`/`DerefMut`.

### Test suite
```bash
cargo test -p steel-core --lib light
```

### Benchmark reproduction
```bash
```

## Screenshots / logs

|---|---|---|---|
| **Total Allocations** | 3.378 GB | **3.130 GB** | **-248 MB (-7.4%)** |
| **Allocated Blocks** | 8.25M blocks | **8.22M blocks** | **-30K blocks** |

## Checklist

- [x] Self-reviewed the diff

## Additional notes

- All 128 light engine tests pass with identical lighting output.


steel-core/src/chunk/light/queue.rs
```diff
@@ -473,6 +475,75 @@ impl PackedLightPropagationQueues {
...
     }
+
+    const fn total_capacity(&self) -> usize {
+        self.increase.entries.capacity() + self.decrease.entries.capacity()
+    }
+}
+
+/// Retained-capacity ceiling for the recycled queue pair. Larger buffers are
+/// dropped instead of pinned to a worker thread (2 MiB of packed entries).
+const POOLED_PACKED_QUEUES_MAX_ENTRIES: usize = 128 * 1024;
```
Joniii11: this is just about ~1 MiB but the comment says 2 so maybe use 256 * 1024 lol

Joniii11 (approved):
besides the midleading comment LGTM

steel-core/src/chunk/light/queue.rs
```diff
@@ -473,6 +475,75 @@ impl PackedLightPropagationQueues {
...
+
+thread_local! {
```
Joniii11, line 504: Do we have retained/ peak memory numbers to? This reduces allocations yes but each worker can now keep pooled queue memory alive well even tho its just 1 mib atm (if my thing gets fixed then 2mib)

Joniii11:
please read my feedback in the approved thing thanks ^^

steel-core/src/chunk/light/queue.rs
carabistouflette: Entry is 8 bytes so the 128K ceiling was ~1 MiB. Raised to 256K entries in fe6d4d3f so retention matches the comment

steel-core/src/chunk/light/queue.rs
carabistouflette: Worst case is one pair per worker thread capped at the ceiling (~2 mib). The pool holds a single slot so it cannot grow beyond that. Allocated bytes drop 248 mb over the pregen census while resident growth stays under 2 mib per worker once warm

steel-core/src/chunk/light/queue.rs
```diff
@@ -1,4 +1,6 @@
+use std::cell::RefCell;
 use std::iter::FusedIterator;
+use std::ops::{Deref, DerefMut};
```
Joniii11, line 4: smol import needed for the guard replacement below

```suggestion
use std::cell::RefCell;
use std::iter::FusedIterator;
use std::mem;
use std::ops::{Deref, DerefMut};
```

steel-core/src/chunk/light/queue.rs
```diff
@@ -473,6 +475,76 @@ impl PackedLightPropagationQueues {
...
+        let Some(queues) = self.inner.take() else {
+            return;
+        };
+        let recycled =
+            (queues.total_capacity() <= POOLED_PACKED_QUEUES_MAX_ENTRIES).then_some(queues);
+        POOLED_PACKED_QUEUES.with(|pool| {
+            *pool.borrow_mut() = recycled;
+        });
+    }
 }
```
Joniii11, line 575: Could we encode this invariant in the type instead of using Option + production expect()? This keeps Deref infallible and still lets Drop return the buffers to TLS without reallocating the 64 KiB queue pair

```suggestion
    /// Creates an empty queue pair without reserving propagation scratch capacity.
    fn empty_without_capacity() -> Self {
        Self {
            increase: PackedLightPropagationQueue {
                entries: Vec::new(),
                read_index: 0,
            },
            decrease: PackedLightPropagationQueue {
                entries: Vec::new(),
                read_index: 0,
            },
        }
    }

    const fn total_capacity(&self) -> usize {
        self.increase.entries.capacity() + self.decrease.entries.capacity()
    }
}

/// Retained-capacity ceiling for the recycled queue pair, summed over both
/// queues (~2 MiB of packed 8-byte entries). Larger buffers are dropped
/// instead of pinned to a worker thread.
const POOLED_PACKED_QUEUES_MAX_ENTRIES: usize = 256 * 1024;

thread_local! {
    static POOLED_PACKED_QUEUES: RefCell<Option<PackedLightPropagationQueues>> =
        const { RefCell::new(None) };
}

/// A packed propagation queue pair leased from a thread-local pool.
///
/// Light chunk work constructs fresh queues per chunk; leasing recycles the
/// worker's previous buffers instead of reallocating ~64 KiB each time. The
/// guard returns its buffers to the pool on drop, so early returns and panics
/// stay safe.
#[must_use]
pub struct PooledPackedLightQueues {
    inner: PackedLightPropagationQueues,
}

impl PooledPackedLightQueues {
    /// Leases empty queues from the worker's pool, allocating on first use.
    pub fn take() -> Self {
        let inner = POOLED_PACKED_QUEUES.with(|pool| match pool.borrow_mut().take() {
            Some(queues) => queues,
            None => PackedLightPropagationQueues::new(),
        });

        Self { inner }
    }
}

impl Default for PooledPackedLightQueues {
    fn default() -> Self {
        Self::take()
    }
}

impl Deref for PooledPackedLightQueues {
    type Target = PackedLightPropagationQueues;

    #[inline]
    fn deref(&self) -> &Self::Target {
        &self.inner
    }
}

impl DerefMut for PooledPackedLightQueues {
    #[inline]
    fn deref_mut(&mut self) -> &mut Self::Target {
        &mut self.inner
    }
}

impl Drop for PooledPackedLightQueues {
    fn drop(&mut self) {
        let mut queues = mem::replace(
            &mut self.inner,
            PackedLightPropagationQueues::empty_without_capacity(),
        );

        queues.clear();

        let recycled =
            (queues.total_capacity() <= POOLED_PACKED_QUEUES_MAX_ENTRIES).then_some(queues);

        POOLED_PACKED_QUEUES.with(|pool| {
            *pool.borrow_mut() = recycled;
        });
    }
}
```

Joniii11 (changes requested):

Joniii11:
these are suggestions i would do

Joniii11 (approved):
lgtm

steel-core/src/chunk/light/queue.rs
carabistouflette: Added use std::mem and applied your type-level version

steel-core/src/chunk/light/queue.rs
carabistouflette: you have falcon eyes bro, I've done your improvements and thanks for your help

Joniii11 (approved):
lgtm

carabistouflette:
@kdcokenny looks good to you?

kdcokenny:
detailed explanation from author that I suggest reviewers to read through: https://discord.com/channels/1428487339759370322/1428643159755325581/1542976732519206993

kdcokenny (approved):

coco875 (approved):
instead of create and remove new one each time, each thread have their queue, so it can be re-use between two iteration, no significant perfomance amelioration but avoid a frequent allocation/deallocation

JunkyDeveloper:
If another maintainer can reproduce the issues, I would close it because I have worse performance.

from claude: https://claude.ai/code/artifact/6319dc53-cae9-4dfe-96b8-6f7a9ba7e20d?via=auto_preview

mine:
Master:
dhat: Total:     3,438,902,575 bytes in 6,268,039 blocks
dhat: At t-gmax: 264,616,067 bytes in 234,403 blocks
dhat: At t-end:  21,055,017 bytes in 47,197 blocks

PR 481:
dhat: Total:     4,272,621,024 bytes in 11,919,688 blocks
dhat: At t-gmax: 244,221,573 bytes in 219,373 blocks
dhat: At t-end:  19,857,760 bytes in 40,062 blocks

coco875:
the branch are not update in master and clumzclik have made a pr to reduce allocation so check that both are upstream or at the same level

JunkyDeveloper (approved):

PR #482, opened 2026-08-24
perf(chunk-saver): pack section palettes without per-block linear scans

carabistouflette:
## Type of change

- [x] Performance improvement

## Description

When saving chunk sections (`section_to_persistent` and `biomes_to_persistent`), the code performed a linear scan (`palette.iter().position(...)`) for all 4,096 blocks in every section (up to 65,000 scans per section). It then collected all indices into an intermediate 16 KiB `Vec<u32>` before bit-packing.

This PR:
1. Inverts the palette into an `FxHashMap` once per section.
2. Adds `pack_indices_from_iter`, streaming packed bits directly from the cube iterator into the destination slice without allocating an intermediate `Vec<u32>`.

### Test suite
```bash
cargo test -p steel-core --lib chunk_saver
```

### Benchmark reproduction
```bash
```

## Screenshots / logs

|---|---|---|---|
| **Total Allocations** | 3.378 GB | **3.156 GB** | **-222 MB (-6.6%)** |
| **Allocated Blocks** | 8.25M blocks | **8.09M blocks** | **-167K blocks (-2.0%)** |

## Checklist

- [x] Self-reviewed the diff

## Additional notes

- Packed format output is 100% byte-identical (same LSB-first bit packing).


steel-core/src/chunk_saver/storage/chunk.rs
```diff
@@ -338,20 +342,21 @@ impl ChunkStorage {
...
                     .collect();
-
-                let biome_data = pack_indices(&indices, bits);
+                let biome_data = pack_indices_from_iter(
+                    data.cube
+                        .iter()
+                        .flatten()
+                        .flatten()
+                        .map(|biome_id| palette_indices.get(biome_id).copied().unwrap_or(0)),
+                    data.cube.len() * 64,
```
Joniii11: Isn't this supposed to be 64 total? Biome cube is 4x4x4, but data.cube.len() * 64 gives 256. Iterator only gives 64 values and loader also expects 64, so heterogeneous biome chunk saved here cannot load again

Joniii11 (changes requested):

steel-core/src/chunk_saver/storage/chunk.rs
carabistouflette: fixed the count to derive from all three cube dimensions and added a save load roundtrip regression test

steel-core/src/chunk_saver/bit_pack.rs
```diff
@@ -19,33 +19,35 @@ pub const fn bits_for_palette_len(palette_len: usize) -> Option<u8> {
...
         "bits must be 1, 2, 4, 8, or 16"
     );
-    if indices.is_empty() {
+    if entry_count == 0 {
         return Box::new([]);
     }
     let bits = bits as usize;
     let values_per_u64 = 64 / bits;
-    let num_u64s = indices.len().div_ceil(values_per_u64);
+    let num_u64s = entry_count.div_ceil(values_per_u64);
```
Joniii11, line 46: Please dont pass item count separately. Wrong count already broke biome saves once. Too small can panic &too large writes extra words Make the packer get its length from the iterator.

Joniii11 (changes requested):

steel-core/src/chunk_saver/bit_pack.rs
carabistouflette: done now

Joniii11 (approved):

steel-core/src/chunk_saver/storage/chunk.rs
```diff
@@ -286,23 +288,24 @@ impl ChunkStorage {
...
-
-                let block_data = pack_indices(&indices, bits);
+                let block_data = pack_indices_from_iter(
+                    data.cube
+                        .as_flattened()
+                        .as_flattened()
+                        .iter()
+                        .map(|block_id| palette_indices.get(block_id).copied().unwrap_or(0)),
+                    bits,
+                );
```
coco875, line 308: I don't get what we gain of doing this because at the end we still do
```rust
data.cube
                        .as_flattened()
                        .as_flattened()
                        .iter()

**Review by coco875 — CHANGES_REQUESTED** [review]
don't get why it's a performance improvement if at the end we do the same iteration

File: steel-core/src/chunk_saver/storage/tests/chunks_sections.rs [hunk 4]
```diff
@@ -311,3 +314,52 @@ fn full_chunk_postprocessing_roundtrips_through_persistent_chunk() {
...
+        0,
+        16,
+        Weak::new(),
+    );
+    let loaded = loaded.expect("heterogeneous biome data must keep its exact packed entry count");
+    assert_eq!(
+        loaded.chunk.sections.read_all_biomes().as_ref(),
+        expected.as_ref()
+    );
+}
```
JunkyDeveloper, line 365: What does this test do?

steel-core/src/chunk_saver/bit_pack.rs
```diff
@@ -28,24 +32,21 @@ pub const fn bits_for_palette_len(palette_len: usize) -> Option<u8> {
...
     let mut data = vec![0u64; num_u64s];
-    for (i, chunk) in indices.chunks(values_per_u64).enumerate() {
-        let mut word = 0u64;
-        for (j, &index) in chunk.iter().enumerate() {
-            word |= u64::from(index) << (j * bits);
-        }
-        data[i] = word;
+    for (i, index) in indices.enumerate() {
+        data[i / values_per_u64] |= u64::from(index) << ((i % values_per_u64) * bits);
```
JunkyDeveloper, line 49: Please add a comment and explain what bit shifts and or do you do and why

JunkyDeveloper (dismissed):

steel-core/src/chunk_saver/storage/chunk.rs
carabistouflette: we still visit every cell, but removed the palette scan per cell

carabistouflette: The inverse map is built once for lookups, and packing directly from the iterator avoids the intermediate Vec<u32> and extra pass
Also DHAT shows allocation savings only

steel-core/src/chunk_saver/storage/tests/chunks_sections.rs
carabistouflette: it regresses the biome entry-count bug Joniii11 found. It saves mixed biomes, reloads the section, and compares every biome cell

the two corners are desert and the rest keep default biome

steel-core/src/chunk_saver/bit_pack.rs
carabistouflette: added

steel-core/src/chunk_saver/storage/chunk.rs
```diff
@@ -286,23 +288,24 @@ impl ChunkStorage {
...
                     .collect();
 
-                // Pack block indices (indices into section-local palette)
+                // Pack block indices (indices into section-local palette),
+                // inverting the palette once instead of scanning it per block.
                 let bits = bits_for_palette_len(palette.len())
                     .expect("Heterogeneous section should have palette length >= 2");
-                let indices: Vec<u32> = data
-                    .cube
+                let palette_indices: FxHashMap<BlockStateId, u32> = data
```
coco875: hmm did you try with a SmallHashMap ? to try increate a bit performance and also alocation if it don't contain much item

steel-core/src/chunk_saver/storage/chunk.rs
```diff
@@ -338,20 +341,20 @@ impl ChunkStorage {
 
                 let bits = bits_for_palette_len(palette.len())
                     .expect("Heterogeneous biome data should have palette length >= 2");
-                let indices: Vec<u32> = data
-                    .cube
+                let palette_indices: FxHashMap<u16, u32> = data
```
coco875: same as above

steel-core/src/chunk_saver/bit_pack.rs
```diff
@@ -28,24 +32,26 @@ pub const fn bits_for_palette_len(palette_len: usize) -> Option<u8> {
...
-        let mut word = 0u64;
-        for (j, &index) in chunk.iter().enumerate() {
-            word |= u64::from(index) << (j * bits);
-        }
-        data[i] = word;
+    // Integer division selects the destination word; the remainder selects its
+    // entry slot. Shift the index into that slot, starting at the least significant
+    // bits, then OR it into the zero-initialized word without changing earlier slots.
+    // Each index fits in `bits` bits, and the supported widths divide 64, so slots
+    // do not overlap or cross word boundaries.
```
coco875: a bit long the comment

coco875 (changes requested):

steel-core/src/chunk_saver/storage/chunk.rs
carabistouflette: replaced with small fxsmall map

steel-core/src/chunk_saver/storage/chunk.rs
carabistouflette: same

steel-core/src/chunk_saver/bit_pack.rs
carabistouflette: done

coco875 (approved):

CuzImClicks (approved):

PR #485, opened 2026-08-24
perf(worldgen): reuse trilerp y/x-stage partials across z columns

carabistouflette:
## Type of change

- [x] Performance improvement

## Description

In `NoiseChunk::fill`, the first two stages of the per-block trilinear interpolation (the y-lerp and x-lerp) do **not** depend on `factor_z`. Their results are therefore identical for the four blocks sharing a cell-Y row across a cell's four z columns — but they were being recomputed for every block.

This PR computes the y/x-stage partials once per column row into flat reusable buffers sized by the active channel count, then finishes only the z-stage per block:

- Vector work per block drops from ~21 to ~9 ops (amortized).
- Cell-slice loads drop from 8 to 2 per block (amortized).

Per-block arithmetic keeps the original operand order and `place_block` keeps its exact call order, so worldgen output is **bit-identical**.

### Bit-parity verification
Full-chunk block arrays were checksummed (FNV-1a over all `BlockStateId`s after `create_biomes` + `fill_from_noise`) for two seeds on both the previous kernel and this one — identical hashes:

| Seed | Checksum |
|---|---|
| `0x0` | `0x63fba173375d2590` |
| `0x5eed1234abcd` | `0x117723c9cad416d9` |

### Test suite
```bash
cargo test -p steel-worldgen
cargo test -p steel-core
```

### Benchmark reproduction
```bash
cargo build --profile profiling
```
Pinned seed (`steelmc-perf`), 4,225 chunks, 5 interleaved runs per variant.

## Screenshots / logs

Wall-clock pregen throughput (5 runs each, interleaved to control drift):

| Variant | Median | Delta |
|---|---|---|
| `master` | 443–445 chunks/s | baseline |
| This PR | 450 chunks/s | **+1.1–1.5%** |

## Checklist

- [x] Self-reviewed the diff

## Additional notes

- The two scratch buffers add one allocation per chunk in `fill` (~50 KB for overworld-sized columns), amortized by the allocator across chunks.
- Zero gameplay / worldgen determinism change (verified bit-identical).


carabistouflette:
not good enough :(

carabistouflette:
Mind you, that’s a win in itself, given how much A/B testing I did to confirm it wasn't just noise.

steel-worldgen/src/noise/noise_chunk.rs
```diff
@@ -295,117 +295,165 @@ impl<N: DimensionNoises> NoiseChunk<N> {
...
+        // `cell_y_idx * cell_height + y_in_cell`.
+        //
+        // The per-block arithmetic (operand order included) is unchanged, and
+        // `place_block` keeps its original call order, so results and
+        // column-batched write behavior stay bit-identical.
+        let column_len = cell_count_y * cell_height as usize;
+        // Flat row-major storage sized by the ACTIVE channel count (not
+        // MAX_INTERP), so dimensions with fewer channels don't over-allocate.
+        let mut d0_col = vec![0.0f64; column_len * interp_count];
+        let mut d1_col = vec![0.0f64; column_len * interp_count];
```
Joniii11: didn't u say 1 alloc lol, well whatever

Could these use one reusable scratch allocation instead? split_at_mut would avoid the second allocation, and reusing the buffer between chunks could avoid most of the allocation churn completely

Joniii11 (changes requested):
does the parity check also cover nether or end dims? cause they use interp_count = 1

also please provide that the chunk gen is still vanilla par

steel-worldgen/src/noise/noise_chunk.rs
carabistouflette: One allocation now, my bad. I kept it at one allocation per fill call instead of pooling across chunks, since the size varies per dimension and cross-call retention would buy little next to the noise sampling itself

carabistouflette: For interp_count = 1:nether and end run entirely through the scalar tail, which keeps master's exact operand order (d00/d10/d01/d11, then d0/d1, then the z lerp), so they are covered by construction as well

carabistouflette: also I already ran the repo's chunk_stage_hashes regression and it passes

Joniii11: could you still uh prove it that it is 100% correct?

We got a test for that which checks 3500 chunks across all 3 dims for 100% vanilla parity

carabistouflette: tell me if it's all good to you now

Joniii11: > also I already ran the repo's chunk_stage_hashes regression and it passes

on all prs? if so then you get my lgtm after i review again

carabistouflette: ye I run it each time so each PR

carabistouflette: also if I do english mistakes my bad I'm using google translate as it's not my primary language

Joniii11: alrighty will keep that in mind i'll tomorrow when i got time review them all again and then give my generic LGTM and afterwards the maintainers coco and junky will maybe also give lgtm's lol

carabistouflette: perfect

Joniii11 (approved):
lgtm

coco875 (approved):
with
overworld_fill_from_noise
nether_fill_from_noise
end_fill_from_noise

I have seen:
I build | Dimension | improvement |
| --- | --- | --- |
| Default | Overworld | 9.9% |
| Default | Nether | 16.4 % |
| Default | End | 15.6 % |
| Native | Overworld | 10.1% |
| Native | Nether | 17.1% |
| Native | End | 24% |

I have try to change the vec with a smallVec (with a size of 256) but didn't see performance improvement

coco875:
I have also double check that it match it pass worldgen test

JunkyDeveloper (approved):

PR #488, opened 2026-08-24
perf(worldgen): position-batched SIMD lane selection for density corner fill

carabistouflette:
# NEED HELP WITH THIS ONE IF YOU HAVE AN AVX-512 CAPABLE CPU
## Type of change

- [x] Performance improvement

## Description

The transpiled density-function corner-fill backend was hardcoded to 4-lane
`f64x4`. This PR makes the lane width a **build-time setting**
(`STEEL_DENSITY_LANES`, default `4`) so the transpiler can emit 8-lane
(`f64x8`) corner evaluation where the target CPU favours it:

- `codegen_expr` emits a `__LaneV` alias instead of literal `f64x4`; all node
  forms (splat, select, `from_array`, scalar-fallback lanes, rarity mapper)
  become width-agnostic.
- Named `compute_*` helpers get width-suffixed idents; the dimension impl
  overrides `fill_cell_corner_densities_8x` or `_4x` to match the setting and
  exposes `simd_lanes()`.
- `NoiseChunk::fill_slice_into` dispatches on `simd_lanes()` and batches
  cell-corner Y values accordingly (8-lane path halves per-column call count).

**Status: draft.** Bit-parity is proven for both widths, but on the test
machine (Lunar Lake, 4P+4E) the measured throughput is flat-to-slightly-negative
at 8 lanes — see the benchmark table below. Opening as draft so CI/maintainers
can evaluate on wider CPUs (AVX-512) and advise whether to keep the
infrastructure with default 4, tune per-target, or drop it.

### Bit-parity (both widths)
Full-chunk block checksums after `create_biomes` + `fill_from_noise` are
**identical to master** at `STEEL_DENSITY_LANES=4` and `=8`:

| Seed | Checksum |
|---|---|
| `0x0` | `0x9cdaed00b4c40375` |
| `0x5eed1234abcd` | `0xcda05a50b795f863` |

### Test suite
```bash
cargo test --workspace          # all green at STEEL_DENSITY_LANES=4
cargo fmt --all --check
cargo clippy --workspace --all-targets --all-features   # 0 warnings
```

### Benchmark reproduction (pinned seed, 4,225 chunks, interleaved runs)
```bash
cargo build --profile profiling                      # LANES=4 default
STEEL_DENSITY_LANES=8 cargo build --profile profiling
```

## Screenshots / logs

Median chunks/s over 5 interleaved runs per variant (Lunar Lake 4P+4E):

| Variant | Median cps | Δ vs master |
|---|---|---|
| `master` | 444.8 | baseline |
| This PR, `LANES=4` | 445.4 | +0.1% (noise) |
| This PR, `LANES=8` | 441.5 | -0.7% |

## Checklist

- [x] Self-reviewed the diff

## Additional notes

- The 8-lane regression is consistent with register-pressure spills in deep
  tree nodes on 256-bit targets; AVX-512 hosts may flip the result, which is
  why the knob exists rather than a hard-coded choice.
- Zero gameplay/worldgen determinism change (checksums above).

### DIFF
--- a/steel-worldgen/build/density/functions.rs
+++ b/steel-worldgen/build/density/functions.rs
@@ -1,7 +1,9 @@
 use proc_macro2::TokenStream;
+use quote::format_ident;
 use quote::quote;
 use serde::Deserialize;
 use std::collections::BTreeMap;
+use std::env;
 use std::path::Path;
 use std::string::String;
 use std::sync::Arc;
@@ -685,6 +687,7 @@ fn transpile_dimension(
     dimension: &str,
     prefix: &str,
     registry: &BTreeMap<String, DensityFunction>,
+    simd_lanes: usize,
 ) -> TokenStream {
     let settings = read_noise_settings(dimension);
     let router_entries = router_to_entries(&settings.noise_router);
@@ -696,9 +699,10 @@ fn transpile_dimension(
         prefix: prefix.to_string(),
         cell_width,
         legacy_random_source: settings.legacy_random_source,
+        simd_lanes,
     };
 
-    transpile(&input)
+    transpile(&input, simd_lanes)
 }
 
 /// Generate noise settings constants and trait impls for a dimension.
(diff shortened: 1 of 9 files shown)


Joniii11:
@kdcokenny don't you have an avx 512 cpu?

Joniii11:
to maybe test this

kdcokenny:
> @kdcokenny don't you have an avx 512 cpu?

I don't, no

tn-lorenz:
Would it be possible to rent a server with an AVX-512 capable CPU? Or do these generally not have that capability?

carabistouflette:
> Would it be possible to rent a server with an AVX-512 capable CPU? Or do these generally not have that capability?

it should work I guess

PR #491, opened 2026-08-25
feat: Implement scaffolding block and item behavior

kdcokenny:
## Type of change

- [x] Block implementation
- [x] Item implementation
- [x] New feature

## Description

Implements Minecraft 26.2 scaffolding behavior, including directional placement, stability distance, waterlogging, support removal, falling, drops, collision, and build-height handling.

Scaffolding items now use the normal block placement pipeline while applying their vanilla placement rules.

## How this was tested

- 14 scaffolding-focused tests and the absolute world-bound regression test passed.
- The Fabric client GameTest suite passed all 10 cases against a live Steel server.
- Manually tested placement, climbing upward, and descending through scaffolding with a Minecraft Java 26.2 client.

## Screenshots / logs

```text
SCAFFOLD_GT_SUITE_PASS cases=10
```

## Checklist

- [x] Self-reviewed the diff

## Additional notes

Behavior was checked against regenerated Minecraft 26.2 sources.

No generated Rust files were edited directly. Item registration is generated from the new `#[item_behavior]` implementation.

Steel still cannot change scaffolding's outline based on the held item because outline raycasts do not currently carry held-item context.


steel-core/src/behavior/block/context.rs
```diff
@@ -217,6 +217,12 @@ impl BlockCollisionContext {
     /// Collision context for vanilla pre-move collision validation.
     #[must_use]
     pub const fn pre_move(entity_bottom: f64, descending: bool) -> Self {
+        Self::placement(entity_bottom, descending)
+    }
+
+    /// Collision context for vanilla block-placement obstruction checks.
+    #[must_use]
+    pub const fn placement(entity_bottom: f64, descending: bool) -> Self {
```
coco875: why have two function that do the same thing ?

coco875 (changes requested):
Look pretty good, just one weird thing

steel-core/src/behavior/block/context.rs
kdcokenny: i added it bc BlockItem needs `placementContext` and `pre_move` didn't look right there. looking into it more, `pre_move` is actually just vanilla's `withPosition` with a different name, and with the fields we have rn they're the same. i'll rm the duplicate from here. are you cool with me opening another pr to rename `pre_move` to `with_position`? i think that's better bc `pre_move` describes where we use it, not the context itself.

coco875: yes

coco875 (dismissed):

Joniii11 (approved):
lgtm did not find anything

steel-core/src/behavior/block/context.rs
```diff
@@ -227,6 +227,14 @@ impl BlockCollisionContext {
         }
     }
 
+    /// Placement obstruction context when no entity initiated the placement.
+    ///
+    /// This matches vanilla `CollisionContext.placementContext(null)`.
```
CuzImClicks: example of a good place to reference vanilla since this is not obvious

steel-core/src/behavior/blocks/building/scaffolding_block.rs
```diff
@@ -29,23 +34,53 @@ const SHAPE_STABLE: VoxelShape = VoxelShape::from_boxes(SHAPE_STABLE_BOXES);
 const SHAPE_UNSTABLE_BOTTOM: VoxelShape = VoxelShape::from_boxes(SHAPE_UNSTABLE_BOTTOM_BOXES);
 const SHAPE_BELOW_BLOCK: VoxelShape = VoxelShape::from_boxes(SHAPE_BELOW_BLOCK_BOXES);
 
-/// Vanilla scaffolding collision-shape behavior.
-///
-/// TODO: Add vanilla placement, stability distance updates, falling conversion, and waterlogging.
+const TICK_DELAY: i32 = 1;
+
+/// Vanilla scaffolding placement, stability, falling, and collision behavior.
```
CuzImClicks: example of a bad one. "placement stability" etc are expected to be implemented since this is the block implementation of it.

steel-core/src/behavior/blocks/building/scaffolding_block.rs
```diff
@@ -29,23 +34,53 @@ const SHAPE_STABLE: VoxelShape = VoxelShape::from_boxes(SHAPE_STABLE_BOXES);
...
     #[must_use]
     pub const fn new(block: BlockRef) -> Self {
         Self { block }
     }
+
+    fn is_bottom(&self, world: &dyn LevelReader, pos: BlockPos, distance: u8) -> bool {
+        distance > 0 && world.get_block_state(pos.below()).get_block() != self.block
+    }
+
+    /// Vanilla `ScaffoldingBlock.getDistance`.
```
CuzImClicks: what does this function compute? give an actual comment. @coco875 should these functions be public?

coco875 (approved):

steel-core/src/behavior/blocks/building/scaffolding_block.rs
```diff
@@ -29,23 +34,53 @@ const SHAPE_STABLE: VoxelShape = VoxelShape::from_boxes(SHAPE_STABLE_BOXES);
...
 pub struct ScaffoldingBlock {
     block: BlockRef,
 }
 
 const BOTTOM: &BoolProperty = &BlockStateProperties::BOTTOM;
 const STABILITY_DISTANCE: &IntProperty = &BlockStateProperties::STABILITY_DISTANCE;
+const WATERLOGGED: &BoolProperty = &BlockStateProperties::WATERLOGGED;
 
 impl ScaffoldingBlock {
+    pub(crate) const STABILITY_MAX_DISTANCE: u8 = 7;
```
JunkyDeveloper, line 50: why here and not at the other consts?

JunkyDeveloper (changes requested):

JunkyDeveloper (approved):

steel-core/src/behavior/blocks/building/scaffolding_block.rs
JunkyDeveloper: found the reason

PR #504, opened 2026-08-26
feat: implement vanilla FrostedIceBlock behavior

Coderinsightmaven:
## Type of change

- [x] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Implements vanilla `FrostedIceBlock` melting for Minecraft 26.2.

- Scheduled ticks age frosted ice `0 → 3`, then melt to water (or air when water evaporates)
- Isolated patches melt immediately when a frosted-ice neighbor changes
- Inherits ice light melting and silk-touch break behavior
- Pick-block returns nothing, matching vanilla

## How this was tested

Minecraft 26.2

- `cargo test -p steel-core frosted_ice_block`

## Checklist

- [x] Self-reviewed the diff

## Classes modified:

FrostedIceBlock

## Additional notes

Frost Walker is not implemented yet, so this block is created with `/setblock` (or later by that enchantment). Melting, aging, neighbor isolation, and pick-block were checked against `FrostedIceBlock.java` / `IceBlock.java`.


steel-core/src/behavior/blocks/building/frosted_ice_block.rs
```diff
@@ -0,0 +1,348 @@
...
+            true
+        }
+    }
+
+    fn melt_tick_delay() -> i32 {
+        rand::random_range(MELT_TICK_MIN..=MELT_TICK_MAX)
+    }
+
+    fn melt_brightness(world: &Arc<World>, pos: BlockPos, state: BlockStateId) -> bool {
+        let brightness = if world.dimension_type == &vanilla_dimension_types::THE_END {
```
kdcokenny: So vanilla checks the world key with level.dimension() == Level.END ([26.2 source](https://github.com/366862732/DirectXmod/blob/9157005cd307c8f986c77884d07f4c29f2c457ce/docs/official-262/net/minecraft/world/level/block/FrostedIceBlock.java#L55-L59)). Level.END is a ResourceKey<Level> ([declaration](https://github.com/366862732/DirectXmod/blob/9157005cd307c8f986c77884d07f4c29f2c457ce/docs/official-262/net/minecraft/world/level/Level.java#L112-L114)), while vanilla_dimension_types::THE_END is a dimension type. Steel has no equivalent End world-key constant yet, so add one for minecraft:the_end and compare it with world.key.

steel-core/src/behavior/blocks/building/frosted_ice_block.rs
```diff
@@ -0,0 +1,348 @@
...
+        _state: BlockStateId,
+        world: &Arc<World>,
+        pos: BlockPos,
+        _old_state: BlockStateId,
+        _moved_by_piston: bool,
+    ) {
+        world.schedule_block_tick_default(
+            pos,
+            self.block,
+            rand::random_range(PLACE_TICK_MIN..=PLACE_TICK_MAX),
```
kdcokenny, line 107: When I tested this with fabric gametest, I measured 95 and 109 ticks between ages. I think it's bc steel only invokes on_place for property state changes. slightly_melt changes the age, so this schedules 60–120 ticks before tick schedules the intended 20–40 delay; duplicate suppression then discards the latter.

kdcokenny (changes requested):

kdcokenny:
This is great, but I made a mistake earlier: my earlier test was contaminated by an earlier review i made which caused the 60–120 cadence. Please keep the End world-key fix, but remove the same-block `on_place` guard and its new test. Once that’s done, it’ll be fully approved on my side!

kdcokenny (approved):

coco875 (dismissed):

steel-core/src/behavior/blocks/building/frosted_ice_block.rs
Coderinsightmaven: Addressed in 486c5f812: Steel now has `world::END` (`minecraft:the_end`) and `FrostedIceBlock` compares `world.key` against that world key, matching vanilla `level.dimension() == Level.END`.

steel-core/src/behavior/blocks/building/frosted_ice_block.rs
Coderinsightmaven: Kept the End world-key fix and restored vanilla `on_place` scheduling in 75dc0dd80 after the later review note: the same-block `on_place` guard and its test are gone, so placement still schedules the 60–120 tick delay with no extra age-update skip.

steel-core/src/world/mod.rs
```diff
@@ -227,6 +227,9 @@ pub struct WorldConfig {
     pub difficulty: Difficulty,
 }
 
+/// Vanilla `Level.END` world key (`minecraft:the_end`).
+pub const END: Identifier = Identifier::vanilla_static("the_end");
```
JunkyDeveloper: I think we have already something like that and this will not work at the end. because other:the_end is also an end dimension. So what I know we have other functions which will help you here better but this const is the wrong way

JunkyDeveloper (changes requested):

steel-core/src/world/mod.rs
Coderinsightmaven: I checked this against the local 26.2 source. `FrostedIceBlock.tick` specifically compares `level.dimension() == Level.END`; `Level.dimension()` is the loaded world key (`ResourceKey<Level>`), not its `DimensionType`. Therefore a custom world such as `other:the_end`, even if it uses the vanilla End dimension type, does not take this branch in vanilla. Steel’s `world.key == minecraft:the_end` is the direct equivalent. I also searched Steel for an existing loaded-world-key constant/helper and found none; the existing `vanilla_dimension_types::THE_END` is a different concept and would diverge here.

JunkyDeveloper: Yes and that is exactly the problem

Coderinsightmaven: Understood—the issue is Steel’s domain-scoped world keys: `other:the_end` can use the vanilla End dimension type even though its loaded-world key is not `minecraft:the_end`. Fixed in a4c482d3d. I removed the hardcoded `world::END` key, promoted the existing server dimension-type check to `World::is_end_dimension_type()`, and now frosted ice plus End portal/gateway routing share it. Added a regression test for `other:the_end` using `vanilla_dimension_types::THE_END`.

JunkyDeveloper (approved):

steel-core/src/behavior/blocks/building/frosted_ice_block.rs
```diff
@@ -0,0 +1,361 @@
...
+                if count >= limit {
+                    return false;
+                }
+            }
+        }
+        true
+    }
+
+    /// Vanilla `FrostedIceBlock.slightlyMelt`.
+    ///
```
CuzImClicks: remove claude comment

JunkyDeveloper:
made some changes some days ago

coco875 (approved):

PR #446, opened 2026-08-22
feat: implement dropper block, block entity, container, and 3x3 menu

SanpuiRonak:
## Type of change

- [x] Block implementation
- [x] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description

Implements vanilla **Dropper** and **Dispenser** blocks (`DropperBlock`, `DispenserBlock`), block entities (`DropperBlockEntity`, `DispenserBlockEntity`), shared 3x3 container, and menu (`Generic3x3Kind`).

### Key Features
- **3x3 Menu & Inventory**: Standard 3x3 container grid (9 slots) with bi-directional shift-click transfer routing with player inventory.
- **Block Entity & Container**:
  - 9 inventory slots capped to 64 items per stack.
  - Uniform non-empty slot selection (`get_random_non_empty_slot`).
  - Single item decrement extraction (`take_single_item`).
  - Drop item spill on block destruction (`pre_remove_side_effects`).
  - Full NBT serialization (`save_additional` & `load_additional`) with `Items` list and slot indexes.
- **Block Behaviors**:
  - `FACING` (all 6 directions) and `TRIGGERED` (`bool`) block state properties.
  - Directional placement facing away from the player.
  - Opens 3x3 menu on right click without an item (`INSPECT_DROPPER` / `INSPECT_DISPENSER` stats).
  - Redstone activation: on neighbor power signal rising edge, triggers `dispense_from` and sets `TRIGGERED = true`. On falling edge, resets `TRIGGERED = false`.
  - Item transfer & dispense logic:
    - **Dropper**: Transfers item into adjacent container if present; otherwise drops into world.
    - **Dispenser**: Integrates with `DispenseItemBehavior` system, supporting armor auto-equipping onto living entities (`can_equip_with_dispenser`) and fallback to dropping.
    - Emits sound event `SOUND_DISPENSER_DISPENSE` (1000) on success, or `SOUND_DISPENSER_FAIL` (1001) if empty.
  - Redstone comparator output: calculates analog output signal (0..15) based on container fullness.

## How this was tested

Tested with automated unit tests and in-game client verification:
- `cargo test -p steel-core` (Unit tests covering menu, entity, NBT persistence, container transfer, world drop, comparator signal, redstone edge triggering, and dispense behaviors).
[...]

PR #505, opened 2026-08-26
fix: Correct dispenser and dropper placement orientation

kdcokenny:
## Type of change

- [x] Block implementation
- [x] Bug fix

## Description

Dispenser and dropper placement used `DefaultBlockBehavior`, leaving both blocks facing north regardless of player orientation.

This adds exact-class behaviors matching Minecraft 26.2. Placed dispensers and droppers now face opposite the player's nearest look direction.

## How this was tested

- Compared against generated Minecraft 26.2 source.
- Tested both blocks across all six look directions.
- Fabric client GameTest passed all 12 placements using real client look/use input.

## Screenshots / logs

Fabric client GameTest completed with `BUILD SUCCESSFUL`.

## Checklist

- [x] Self-reviewed the diff

## Classes modified:

- DispenserBlock
- DropperBlock

## Additional notes

This change is unaffected by the planned loot-table overhaul. It only selects the block state used during placement and does not touch loot generation, inventories, block entities, redstone activation, or dispensing behavior.

Steel registers behaviors by exact Java class name, so `DropperBlock` needs its own Rust behavior even though vanilla inherits placement behavior from `DispenserBlock`.

### DIFF
--- a/steel-core/src/behavior/blocks/container/placement_tests.rs
+++ b/steel-core/src/behavior/blocks/container/placement_tests.rs
@@ -1,0 +1,64 @@
+use glam::DVec3;
+use steel_registry::blocks::BlockRef;
+use steel_registry::blocks::block_state_ext::BlockStateExt as _;
+use steel_registry::blocks::properties::{BlockStateProperties, Direction, EnumProperty};
+use steel_registry::item_stack::ItemStack;
+use steel_registry::{init_vanilla_registry, vanilla_blocks, vanilla_items};
+use steel_utils::BlockPos;
+use steel_utils::types::InteractionHand;
+
+use crate::behavior::{
+    BLOCK_BEHAVIORS, BlockHitResult, BlockPlaceContext, PlacementOrientation, PlacementSource,
+    init_behaviors,
+};
+use crate::test_support::test_world;
+
+const FACING: &EnumProperty<Direction> = &BlockStateProperties::FACING;
+
+fn assert_faces_away_from_player_look(block: BlockRef) {
+    let behavior = BLOCK_BEHAVIORS.get_behavior(block);
+
+    for look_direction in Direction::ALL {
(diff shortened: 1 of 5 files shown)


steel-core/src/behavior/blocks/container/placement_tests.rs
```diff
@@ -0,0 +1,64 @@
...
+
+use crate::behavior::{
+    BLOCK_BEHAVIORS, BlockHitResult, BlockPlaceContext, PlacementOrientation, PlacementSource,
+    init_behaviors,
+};
+use crate::test_support::test_world;
+
+const FACING: &EnumProperty<Direction> = &BlockStateProperties::FACING;
+
+fn assert_faces_away_from_player_look(block: BlockRef) {
```
Joniii11, line 18: i mean is this really needed its like a test which says this constants has the number 8 or whatever.

steel-core/src/behavior/blocks/container/placement_tests.rs
```diff
@@ -0,0 +1,64 @@
...
+            .get_state_for_placement(&context)
+            .expect("dispenser-family placement should produce a state");
+
+        assert_eq!(state.get_block(), block);
+        assert_eq!(state.get_value(FACING), look_direction.opposite());
+    }
+}
+
+#[test]
+fn dispenser_and_dropper_face_away_from_player_look() {
```
Joniii11, line 58: same here

Joniii11 (approved):
LGTM

steel-core/src/behavior/blocks/container/placement_tests.rs
kdcokenny: This is regression coverage for the reported always-north bug, not a constant test.

JunkyDeveloper:
Please check next time the PR list will be closed because #446 will be preferred because almost full block implementation

PR #506, opened 2026-08-26
Implement compass lodestone binding and periodic target invalidation

MarshY2k:
## Type of change

- [ ] Block implementation
- [x] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description
CompassItem
This PR fully implements Lodestone compass mechanics to parity with vanilla 1.21.1+.

What changed:
•Added inventory_tick to ItemBehavior to the player's tick loop so inventory items can receive regular updates.
•Added the use_on logic for compasses on Lodestones. It handles the sounds, sets the LODESTONE_TRACKER component, and compass tacks correctly.
•Added Invalidation and dimension checks: the compass now checks if the Lodestone still exists once the chunk loads. If the block is destroyed, the tracker resets. The needle will also spin randomly if you carry the compass into a different dimension.

## How this was tested
- Added an automated unit test compass_binds_to_lodestone_and_invalidates_when_removed to compass.rs that validates binding, ticking when present, and invalidating once replaced by air.
- Checked full compilation, clippy checks, and formatting:
  - Unit tests: cargo test -p steel-core --lib behavior::items::compass
  - Full test suite: cargo test
  - Clippy check: cargo clippy -r --all-targets --all-features
  - Formatting: cargo fmt --all


steel-core/src/behavior/items/compass.rs
```diff
@@ -24,4 +31,173 @@ impl ItemBehavior for CompassItem {
...
+        let Some(target) = tracker.target() else {
+            return;
+        };
+
+        if world.key != *target.dimension() {
+            return;
+        }
+
+        let target_pos = target.pos();
+        if !world.is_full_chunk_loaded_at(target_pos) {
```
Joniii11, line 118: Should world bounds be checked before this? Vanilla clears lodestone target if position is outside world bounds, but here we return early on unloaded chunk, so invalid target can stay forever

Joniii11 (changes requested):

steel-core/src/behavior/items/compass.rs
MarshY2k: I just tested this in-game with a compass pointing to X: 40,000,000, and you are right ill fix this bug soon.

JunkyDeveloper (dismissed):

steel-core/src/behavior/items/compass.rs
```diff
@@ -24,4 +31,207 @@ impl ItemBehavior for CompassItem {
...
+            return InteractionResult::Pass;
+        }
+
+        context.world.play_block_sound(
+            &sound_events::ITEM_LODESTONE_COMPASS_LOCK,
+            pos,
+            1.0,
+            1.0,
+            None,
+        );
```
CuzImClicks:
```suggestion
        context.world.play_sound(
            &sound_events::ITEM_LODESTONE_COMPASS_LOCK,
            SoundSource::Players,
            pos,
            1.0,
            1.0,
            None,
        );
```
this is a player sound not a block sound

`level.playSound(null, blockPos, SoundEvents.LODESTONE_COMPASS_LOCK, SoundSource.PLAYERS, 1.0F, 1.0F);`

steel-core/src/behavior/items/compass.rs
```diff
@@ -24,4 +31,207 @@ impl ItemBehavior for CompassItem {
             default_name(stack)
         }
     }
```
CuzImClicks, line 35:
```rust

// add this

static LODESTONE_NAME: TextComponent =
    TextComponent::translated(translations::ITEM_MINECRAFT_LODESTONE_COMPASS.msg());

// and change this function to

    fn get_name<'a>(&self, stack: &'a ItemStack) -> Cow<'a, TextComponent> {
        if stack.has(LODESTONE_TRACKER) {
            Cow::Borrowed(&LODESTONE_NAME)
        } else {
            default_name(stack)
        }
    }
```

steel-core/src/behavior/items/compass.rs
```diff
@@ -24,4 +31,207 @@ impl ItemBehavior for CompassItem {
...
+        let Some(target) = tracker.target() else {
+            return;
+        };
+
+        if world.key != *target.dimension() {
+            return;
+        }
+
+        let target_pos = target.pos();
+        if !world.is_in_valid_bounds(target_pos) {
```
CuzImClicks:
```suggestion
        if !world.is_in_world_bounds(target_pos) {
```

steel-core/src/behavior/items/compass.rs
```diff
@@ -24,4 +31,207 @@ impl ItemBehavior for CompassItem {
...
+                context.hand,
+                result_stack,
+                context.player.has_infinite_materials(),
+                true,
+            )
+        });
+
+        if !leftover.is_empty() {
+            let _ = context.player.drop_item(leftover, false, true);
+        }
```
CuzImClicks:
```suggestion
        let has_infinite_materials = context.player.has_infinite_materials();

        let leftover = context.inv.with_inventory(|inv| {
            let held_count = inv.get_item_in_hand(context.hand).count();

            if !has_infinite_materials && held_count == 1 {
                inv.mutate_item_in_hand(context.hand, |item| item.set(LODESTONE_TRACKER, tracker));
                return ItemStack::empty();
            }

            let mut result_stack = inv.get_item_in_hand(context.hand).clone();
            result_stack.set_count(1);
            result_stack.set(LODESTONE_TRACKER, tracker);

            if !has_infinite_materials {
                inv.shrink_item_in_hand(context.hand, 1);
            }

            if inv.add(&mut result_stack) {
                ItemStack::empty()
            } else {
                result_stack
            }
        });

        if !leftover.is_empty() {
            let _ = context.player.drop_item(leftover, false, false);
        }
```

fix for a bug in creative mode and this also only locks the player inventory once

steel-core/src/behavior/item.rs
```diff
@@ -184,6 +184,17 @@ pub trait ItemBehavior: Send + Sync {
...
     }
 
+    /// Called every tick for each item stack in a player's inventory.
+    fn inventory_tick(
+        &self,
+        _stack: &mut ItemStack,
+        _world: &Arc<World>,
+        _player: &Player,
+        _slot: usize,
+        _selected: bool,
```
CuzImClicks, line 194: please merge all of this into a Context struct. please refer to `UseOnContext` and to `UseItemContext` on how to do that. the context should contain a mutable reference to the locked inventory inside

`fn inventory_tick(&self, _context: &mut InventoryTickContext<'_>)`

CuzImClicks (changes requested):

CuzImClicks:
this pr is blocked until #587 is merged. but i verified vanilla behavior and after #587 is merged we only need to change to `inv.mutate_item_in_hand(|item| item.consume_and_return(1, has_infinite_materials))`. and then this pr is good to go

JunkyDeveloper:
please fix lint

steel-core/src/behavior/items/compass.rs
```diff
@@ -24,4 +33,215 @@ impl ItemBehavior for CompassItem {
...
+
+        if !world.is_full_chunk_loaded_at(target_pos) {
+            return;
+        }
+
+        let block_state = world.get_block_state(target_pos);
+        if block_state.get_block() != &vanilla_blocks::LODESTONE {
+            let new_tracker = LodestoneTracker::new(None, true);
+            stack.set(LODESTONE_TRACKER, new_tracker);
+        }
```
JunkyDeveloper, line 126: This is wrong, because if a lodestone get's removed the tracker still points to that block in an unloaded chunk.

Because the tracker asks the poi system, which I already have but I think not good enough for your usecase  @Purdze did the PoI maybe he can help, so either add a TODO or make the foundation first

JunkyDeveloper (changes requested):

steel-core/src/behavior/items/compass.rs
Purdze: when the chunk is loaded, check world.poi_storage for the lodestone type instead of the block state.., when it isn't loaded, keep the early return and add a todo noting poi persistence is missing. the persistence layer is a separate task, the compass shouldn't block on it

 get_type(pos) == LODESTONE.try_id().ok()

PR #507, opened 2026-08-26
feat(entity): implement display entities

SomeYellowGuy:
## Type of change

- [x] Entity implementation

## Description

This PR adds all three types of display entities (block, item and text displays). It also replaces the preexisting block display code.

## How this was tested

I added tests for composing and decomposing transformations, which is very similar to how *Minecraft* does it.

## Checklist

- [x] Self-reviewed the diff

## Classes modified:

BlockDisplay
ItemDisplay
TextDisplay


steel-core/src/entity/entities/objects/technical/display/transformation.rs
```diff
@@ -0,0 +1,411 @@
...
+        compound.insert("right_rotation", self.0.right_rotation.to_nbt_tag());
+        NbtTag::Compound(compound)
+    }
+}
+
+impl FromNbtTag for NormalTransformation {
+    fn from_nbt_tag(tag: simdnbt::borrow::NbtTag) -> Option<Self> {
+        let compound = tag.compound()?;
+        Some(Self(Transformation {
+            translation: Vector3f::from_nbt_tag(compound.get("transformation")?)?,
```
Joniii11: transformation? should be probably translation lol

above irs translation lol
decide

steel-core/src/entity/entities/objects/technical/display/text_display.rs
```diff
@@ -0,0 +1,394 @@
...
+    #[must_use]
+    pub fn background_color(&self) -> i32 {
+        *self.0.background_color.get()
+    }
+
+    /// Sets the background color of this text display to `color`.
+    ///
+    /// **Note:** This property is interpolated.
+    pub fn set_background_color(&mut self, color: i32) {
+        self.0.line_width.set(color);
```
Joniii11: should have been background color which should have been set not the line

steel-core/src/entity/entities/objects/technical/display/mod.rs
```diff
@@ -0,0 +1,450 @@
...
+    }
+    /// Sets the display entity's *start interpolation delay* (the delay in starting an interpolation), in ticks, to `duration`.
+    ///
+    /// If this is set to `0`, interpolation starts immediately.
+    ///
+    /// **Note:** This property is not saved to disk.
+    fn set_transformation_interpolation_delay(&mut self, duration: i32) {
+        self.display_data_mut()
+            .transformation_interpolation_start_delta_ticks
+            .set(duration);
```
Joniii11: Vanilla sends start_interpolation again even when value is same steel skips it so the client may not restart the animation correctly

steel-core/src/entity/entities/objects/technical/display/text_display.rs
```diff
@@ -0,0 +1,394 @@
...
+                "see_through",
+                view.flags().contains(TextDisplayFlags::SEE_THROUGH),
+            );
+            nbt.insert(
+                "default_background",
+                view.flags()
+                    .contains(TextDisplayFlags::USE_DEFAULT_BACKGROUND),
+            );
+
+            nbt.insert("alignment", view.text_opacity());
```
Joniii11: opacity as alignment? hold on a minute

steel-core/src/entity/entities/objects/technical/display/text_display.rs
```diff
@@ -0,0 +1,394 @@
...
+                flags.insert(TextDisplayFlags::USE_DEFAULT_BACKGROUND);
+            }
+            view.set_flags(flags);
+
+            let alignment = nbt.get("alignment").and_then(Alignment::from_nbt_tag);
+            if let Some(alignment) = alignment {
+                view.set_alignment(alignment);
+            }
+
+            // TODO: resolve text component
```
Joniii11: ee already have the foundations to resolve that

steel-core/src/entity/entities/objects/technical/display/transformation.rs
```diff
@@ -0,0 +1,411 @@
...
+    /// This function uses *singular value decomposition* (or SVD) to split the matrix
+    /// into a set of transformation values.
+    #[expect(clippy::similar_names, reason = "matches the stages of decomposition")]
+    #[must_use]
+    pub fn decompose(matrix: Matrix4f) -> Self {
+        let scale_factor = 1.0 / matrix.0.col(3)[3];
+        let input = matrix.0 * scale_factor;
+
+        // Extract the translation.
+        let translation = input.w_axis.xyz() * scale_factor;
```
Joniii11, line 120: input was already multiplied twice i believe vanilla does that jus tonce

steel-registry/src/entity_data/math.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+impl FromNbtTag for Quaternionf {
+    fn from_nbt_tag(tag: BorrowedNbtTag) -> Option<Self> {
+        // One of the two: 4 floats or AxisAngle4f
+        if let Some(l) = tag.list()
+            && let Some(floats) = l.floats()
+            && floats.len() == 4
+        {
+            return Some(Quaternionf::new(floats[0], floats[1], floats[2], floats[3]));
```
Joniii11: vanilla bormalizes quaternions here

steel-core/src/entity/entities/objects/technical/display/transformation.rs
```diff
@@ -0,0 +1,411 @@
...
+
+#[derive(Clone, Copy)]
+struct Givens {
+    sin_half: f32,
+    cos_half: f32,
+}
+
+impl Givens {
+    pub const G: f32 = 3.0 + 2.0 * consts::SQRT_2;
+    const SIN_PI_8: f32 = 0.382_683_43;
```
Joniii11, line 200: wrong constant howndid you do that did you misstype?
should have been 0.382_683_46

steel-core/src/entity/entities/objects/technical/display/text_display.rs
```diff
@@ -0,0 +1,394 @@
...
+    fn load_additional(&self, nbt: BorrowedNbtCompoundView<'_, '_>) {
+        self.with_view(|mut view| {
+            <Self as Display>::load_display(&mut view, nbt);
+
+            view.set_line_width(nbt.int("line_width").unwrap_or(DEFAULT_LINE_WIDTH));
+            view.set_text_opacity(nbt.byte("text_opacity").unwrap_or(DEFAULT_TEXT_OPACITY) as u8);
+            view.set_background_color(nbt.int("background").unwrap_or(DEFAULT_BACKGROUND_COLOR));
+
+            let mut flags = TextDisplayFlags::empty();
+            if view.shadow() {
```
Joniii11, line 142: This reads current entity flags not NBT ones! Should use shadow see_through and default_background values from nbt here

steel-core/src/entity/entities/objects/technical/display/mod.rs
```diff
@@ -0,0 +1,450 @@
...
+    }
+    /// The base `isIgnoringBlockTriggers()` method for display entities.
+    fn is_ignoring_block_triggers_display(&self) -> bool {
+        true
+    }
+
+    /// Provides a view to the synced data of this entity, accessible via the function `f`.
+    ///
+    /// This allows accessing and modifying any required data.
+    fn with_view(&self, f: impl FnOnce(Self::View<'_>));
```
Joniii11, line 81: This callback runs while entity data mutex is locked calling with_view again inside callback will deadlock can this be avoided or at least documented i mean probably its just a picky comment but yk you never know

steel-registry/src/entity_data/math.rs
```diff
@@ -0,0 +1,161 @@
+use glam::{Mat4, Quat, Vec3};
+use simdnbt::owned::{NbtCompound, NbtList, NbtTag};
+use simdnbt::{FromNbtTag, ToNbtTag};
+
+type BorrowedNbtTag<'a, 'tape> = simdnbt::borrow::NbtTag<'a, 'tape>;
+
+/// A 3D vector (for display entities).
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Vector3f {
```
Joniii11: why do qw even make our own structs here when we have glam? 
Could use glam types directly and keep NBT codec helpers in steel-utils::codec::glam?

steel-core/src/entity/entities/objects/technical/display/block_display.rs
```diff
@@ -0,0 +1,155 @@
...
+        });
+    }
+
+    fn load_additional(&self, nbt: BorrowedNbtCompoundView<'_, '_>) {
+        self.with_view(|mut view| {
+            <Self as Display>::load_display(&mut view, nbt);
+
+            view.set_block_state(
+                nbt.compound("block_state")
+                    .and_then(block_state_nbt::load)
```
Joniii11, line 111: vanilla does by invalid properties use the default bloxk properties and not just air like we do here
cab we have a fallback property like vanilla did with blockstate.codec

steel-core/src/entity/entities/objects/technical/display/item_display.rs
```diff
@@ -0,0 +1,270 @@
...
+    fn new_with_base(base: EntityBase, entity_type: EntityTypeRef) -> Self {
+        Self {
+            base,
+            entity_type,
+            entity_data: SyncMutex::new(ItemDisplayEntityData::new()),
+        }
+    }
+}
+
+impl Entity for ItemDisplayEntity {
```
Joniii11, line 72: in vanilla the item display has a slot 0 and is beeing used by the /item replace entity 
Now steel has no slot here so maybe do a // TODO cause the foundations are missing or implement it idk

steel-core/src/entity/entities/objects/technical/display/mod.rs
```diff
@@ -0,0 +1,450 @@
...
+/// - A brightness and glow color override.
+/// - A maximum and minimum height and width (if set).
+/// - Interpolation properties, like the duration of a transformation interpolation, its delay and the duration
+///   of a teleport interpolation.
+/// - A shadow radius and strength.
+/// - A maximum view range.
+///
+/// To **access** or **modify** the data of a display entity, you will need to use [`Display::with_view`].
+/// This method takes a function with a [`DisplayView`] as a parameter, which can be used within the function.
+pub trait Display: Entity {
```
Joniii11, line 52: here the default impl of gravity is missing default would be noPhysics = true

Joniii11 (changes requested):
heh

steel-core/src/entity/entities/objects/technical/display/transformation.rs
SomeYellowGuy: You're right, thanks!

steel-core/src/entity/entities/objects/technical/display/text_display.rs
SomeYellowGuy: Yeah, but how do I do it with a `World`? I only see how to do it for players

steel-core/src/entity/entities/objects/technical/display/text_display.rs
SomeYellowGuy: Thanks for pointing that out.

steel-core/src/entity/entities/objects/technical/display/text_display.rs
SomeYellowGuy: Thanks for pointing that out.

steel-core/src/entity/entities/objects/technical/display/mod.rs
SomeYellowGuy: Good catch!

steel-core/src/entity/entities/objects/technical/display/text_display.rs
Joniii11: hm true, in vanilla that works well we have a missing foundation then ig its fine

SomeYellowGuy: I'll still make the TODO more clear.

steel-registry/src/entity_data/math.rs
SomeYellowGuy: Should I keep `AxisAngle4f`?

Joniii11: Only keep it private for NBT decode runtime should use glam quat xause vanilla wraps angle and does not normalize axis so cannot blindly use Quat::from_axis_angle

steel-core/src/entity/entities/objects/technical/display/item_display.rs
SomeYellowGuy: It's probably not implemented yet (I can't find `setItemStack` and `getItemStack`), so I'll put a TODO.

steel-registry/src/entity_data/math.rs
SomeYellowGuy: The file you are pointing to is for `WriteTo` and `ReadFrom` implementations, which are defined by Steel, but I can't directly implement the NBT traits (outside trait for another outside crate). Should I add bare functions instead?

steel-core/src/entity/entities/objects/technical/display/mod.rs
SomeYellowGuy: Good catch!

steel-core/src/entity/entities/objects/technical/display/mod.rs
SomeYellowGuy: I think documenting it is enough; calling this again within the callback should be obvious to spot as being wrong.

steel-core/src/entity/entities/objects/technical/display/block_display.rs
SomeYellowGuy: I don't think the `block_state_nbt` implementation is completely right, but at the same time, this feels out of scope for this PR.

steel-registry/src/entity_data/math.rs
Joniii11: Yeah bare NBT encode/decode functions are better because of orphan rule put them in steel_utils::nbt::glam or whatever you tjink ist best & axis angle can stay private inside quaternion decode

steel-core/src/entity/entities/objects/technical/display/block_display.rs
Joniii11: i mean it adds those block displays idk ehat the Maintainers (@coco875 or @junkydeveloper ) thinks bout that but to be honest i would maybe make a helper or just document it with a // FIXME or a // TODO

SomeYellowGuy: Other places in Minecraft's code use the same codec, so it doesn't make much sense to create a helper.

steel-registry/src/entity_data/math.rs
SomeYellowGuy: I think I did everything now.

steel-core/src/entity/entities/objects/technical/display/block_display.rs
SomeYellowGuy: I can add the TODO at the function itself.

coco875: then just add todo

steel-core/src/entity/entities/objects/technical/display/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+/// The default shadow radius of a display entity.
+pub const DEFAULT_SHADOW_RADIUS: f32 = 0.0;
+/// The default shadow strength of a display entity.
+pub const DEFAULT_SHADOW_STRENGTH: f32 = 1.0;
+/// The default width of a display entity.
+pub const DEFAULT_WIDTH: f32 = 0.0;
+/// The default height of a display entity.
+pub const DEFAULT_HEIGHT: f32 = 0.0;
+/// The default glow color override of a display entity.
+///
```
coco875, line 34: the empty line are useful ?

steel-core/src/entity/entities/objects/technical/display/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+    fn transformation(&self) -> Transformation {
+        let data = self.display_data().display();
+        Transformation {
+            translation: *data.translation.get(),
+            left_rotation: *data.left_rotation.get(),
+            scale: *data.scale.get(),
+            right_rotation: *data.right_rotation.get(),
+        }
+    }
+    /// Sets the [`Transformation`] of the display entity to `transformation`.
```
coco875, line 189:
```suggestion
    }

    /// Sets the [`Transformation`] of the display entity to `transformation`.
```

steel-core/src/entity/entities/objects/technical/display/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+    /// Gets the display entity's *interpolation duration* (the time to interpolate to a new transformation), in ticks.
+    fn transformation_interpolation_duration(&self) -> i32 {
+        *self
+            .display_data()
+            .transformation_interpolation_duration
+            .get()
+    }
+    /// Sets the display entity's *interpolation duration* (the time to interpolate to a new transformation), in ticks, to `duration`.
```
coco875, line 212:
```suggestion
    }

    /// Sets the display entity's *interpolation duration* (the time to interpolate to a new transformation), in ticks, to `duration`.
```

steel-core/src/entity/entities/objects/technical/display/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+    fn set_transformation_interpolation_duration(&mut self, duration: i32) {
+        self.display_data_mut()
+            .transformation_interpolation_duration
+            .set(duration);
+    }
+    /// Gets the display entity's *teleport duration* (the time to interpolate to a new position due to a teleport), in ticks.
```
coco875, line 218:
```suggestion
    }

    /// Gets the display entity's *teleport duration* (the time to interpolate to a new position due to a teleport), in ticks.
```

coco875: I let you do the same on other x)

steel-core/src/entity/entities/objects/technical/display/transformation.rs
```diff
@@ -0,0 +1,404 @@
...
+            Givens::from_unnormalized(sin_half, cos_half)
+        } else {
+            Givens::PI_4
+        }
+    }
+
+    fn qr_givens_quat(a: f32, b: f32) -> Givens {
+        let p = a.hypot(b);
+
+        let mut sin_half = if p > 1.0e-6 { b } else { 0.0 };
```
coco875, line 61: look like 1.0e-6 are use at multiple place look like an epsilon, so magic number x)

coco875 (changes requested):

steel-core/src/entity/entities/objects/technical/display/block_display.rs
SomeYellowGuy: Added a todo here.

steel-core/src/entity/entities/objects/technical/display/mod.rs
SomeYellowGuy: They are, but it's probably not necessary for a single sentence after.

steel-core/src/entity/entities/objects/technical/display/transformation.rs
SomeYellowGuy: Where did you get `0.382_683_46` from?

SomeYellowGuy: I printed the `PI_4` value from `MatrixUtil`. It turns out that its values are indeed `sinHalf=0.38268346, cosHalf=0.9238795`. Will fix that.

steel-core/src/entity/entities/objects/technical/display/text_display.rs
SomeYellowGuy: Fixed this.

steel-core/src/entity/entities/objects/technical/display/transformation.rs
SomeYellowGuy: Added a constant for that.

steel-registry/src/entity_data/math.rs
SomeYellowGuy: Didn't notice that, thanks.

steel-core/src/entity/entities/objects/technical/display/transformation.rs
SomeYellowGuy: Fixed that.

PR #511, opened 2026-08-26
feat: implement vanilla mob item pickup

darkstarworks:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Adds the vanilla mob item-pickup ("looting") step to the mob base.

Each server tick, a mob that is allowed to pick up loot scans for nearby dropped items and takes one into its equipment. The scan is gated on the `mobGriefing` game rule and uses vanilla's item pickup reach of one block on each horizontal axis.

* Overridable hooks so individual mobs can shape the behavior: `get_pickup_reach`, `wants_to_pick_up`, `can_hold_item`, and `pick_up_item`.
* The base `pick_up_item` runs the full vanilla `equipItemIfPossible`: it routes the item to the slot its equippable component asks for, swaps out worse gear it already wears, drops the replaced piece by that slot's drop chance, flags the taken item to always drop on death, plays the pickup animation, and removes the item entity.
* Gear is compared the way vanilla does it: armor by armor value then toughness, a main-hand weapon by attack damage, each falling back to a tiebreak on enchantment count, then damage, then a custom name.

This is a foundation for mobs that collect items, and a prerequisite for a Fox entity that overrides `pick_up_item` and `can_hold_item` for its carry-in-mouth behavior.

## How this was tested

Added tests covering a mob taking a nearby item into an empty main hand, an occupied main hand being left alone, a mob with looting disabled ignoring the item, and a mob with no AI still collecting loot through its ai step.

`cargo fmt --all --check`, `cargo clippy -r --all-targets --all-features`, `typos`, and `cargo test` all pass.

## Screenshots / logs

Not applicable. Server-side behavior is covered by the unit tests above.

## Checklist

- [x] Self-reviewed the diff

## Additional notes

A couple of things are worth sanity-checking during review. These are not firm proposals.

* **Equip routing.** `pick_up_item` ports the full vanilla `equipItemIfPossible` rather than only filling an empty main hand, including the armor-slot selection, the `canReplaceCurrentItem` preference between held and dropped gear, and the drop-chance drop of the replaced piece. This needed two small registry helpers: `ItemAttributeModifiers::compute` to fold an item's modifiers onto a base attribute value, and `EquipmentSlot::count_limit` and `limit` to split a stack down to what a slot holds. No armor-wearing mob exercises the armor path yet, so it is covered by unit tests rather than in-game.
* **Placement.** The looting loop runs from a shared `mob_ai_step` (vanilla `Mob.aiStep`) rather than `mob_server_ai_step`, so a mob with `NoAI` set still collects loot, matching vanilla keeping the loop out of `serverAiStep`. The passive mobs now call `mob_ai_step`.


kdcokenny:
@darkstarworks did you test in-game

darkstarworks:
> @darkstarworks did you test in-game

Not yet. one moment please as I finish the last part in Fox entity and test both in-game

kdcokenny:
> > @darkstarworks did you test in-game
> 
> Not yet. one moment please as I finish the last part in Fox entity and test both in-game

ok, i think that this prolly shouldn't be it's own pr then and instead part of that fox pr. Altho i'd request that you join the discord and ask a maintainer for verification.

darkstarworks:
> @darkstarworks did you test in-game

yes:

https://github.com/user-attachments/assets/24c2b503-cb75-4f50-b782-754a54bbc46c

coco875 (dismissed):

steel-core/src/entity/mob/mod.rs
```diff
@@ -570,6 +575,89 @@ pub trait Mob: LivingEntity + Leashable {
...
+
+    /// Vanilla `Mob.canHoldItem`: whether this mob is willing to hold the item.
+    /// The base mob holds anything; specific mobs narrow this down.
+    fn can_hold_item(&self, _item_stack: &ItemStack) -> bool {
+        true
+    }
+
+    /// Vanilla `Mob.pickUpItem`: take a dropped item into this mob's equipment.
+    ///
+    /// TODO(equip-armor): port the full vanilla `equipItemIfPossible` (armor-slot
```
kdcokenny: hmm, i don't think we should be shipping this if there are still todo's in the implemented function itself. The problem is that to properly test that you'd need a monster implemented which could use it - which we currently don't have any that do (that I know of)

steel-core/src/entity/mob/mod.rs
```diff
@@ -1099,6 +1187,7 @@ pub trait Mob: LivingEntity + Leashable {
         self.tick_move_control();
         self.tick_look_control();
         self.tick_jump_control();
+        self.tick_looting();
```
kdcokenny, line 1426: This looks wrong, isn't this skipped for NoAI mods? Vanilla applies looting regardless. Also, I noticed that the tests call tick_looting directly, which would hide this bug.

kdcokenny (changes requested):

steel-core/src/entity/mob/mod.rs
kdcokenny: mobs*

steel-core/src/entity/mob/mod.rs
coco875: like the todo are signalize it can be added later but still have the basic functionality

darkstarworks: Challenge accepted

JunkyDeveloper: I think adding now mobs to finish that is a bit much, so I also like it this way, maybe do the foundation with the other PR together, but please only 1 mob per PR

kdcokenny: > I think adding now mobs to finish that is a bit much, so I also like it this way, maybe do the foundation with the other PR together, but please only 1 mob per PR

so you want it to stay as a todo?

kdcokenny: or build up the mob entity over time based on mobs implemented rather than one shared fn pr?

steel-core/src/entity/mob/mod.rs
darkstarworks: > This looks wrong, isn't this skipped for NoAI mods? Vanilla applies looting regardless. Also, I noticed that the tests call tick_looting directly, which would hide this bug.

good catch. fixing

steel-core/src/entity/mob/mod.rs
darkstarworks: (content removed. double posted thanks to awesome internet)

darkstarworks: Went ahead and built this out rather than leaving the TODO or growing the function per-mob. `pick_up_item` now ports vanilla's `equipItemIfPossible`: items are routed by their equippable slot, better gear replaces worse gear, replaced items respect their drop chance, and `PreventArmorChange` is honored.

This also adds the supporting attribute and `equipment-slot` infrastructure needed for comparisons. The swap path is covered by unit tests, including replacing worse gear, keeping better gear, `curse-of-binding`, and partial stacks.

The TODO(equip-armor) is now gone. I deliberately skipped vanilla's `preferred-weapon-type` check since Mob currently returns none and no mob overrides it yet.

This PR still adds no mobs, so it remains foundation work.

steel-core/src/entity/mob/mod.rs
darkstarworks: Fixed. Vanilla runs looting from `Mob.aiStep`, not `serverAiStep`, so I moved the call into a new `mob_ai_step` (the `aiStep` foundation, then the looting loop) and pointed the passive mobs at it. It no longer sits behind the `isEffectiveAi` gate, so a `NoAI` mob still picks up loot.

I also reworked the test to drive `mob_ai_step` with `NoAI` set and assert the item is still collected, instead of calling `tick_looting` directly, which was hiding exactly this.

steel-core/src/entity/mob/mod.rs
kdcokenny: oh awesome

steel-core/src/entity/mob/mod.rs
```diff
@@ -570,6 +582,282 @@ pub trait Mob: LivingEntity + Leashable {
...
+            slot,
+        );
+        if new_toughness == old_toughness {
+            self.can_replace_equal_item(new_item_stack, current_item_stack)
+        } else {
+            new_toughness > old_toughness
+        }
+    }
+
+    /// Vanilla `Mob.compareWeapons`: prefer the piece dealing more attack damage,
```
kdcokenny: You're missing the preferred item type as far as I can tell. Skeletons for example, prefer bows over swords no matter the damage that sword deals.

kdcokenny (changes requested):

steel-core/src/entity/mob/mod.rs
darkstarworks: Good call, added it. `compare_weapons` now checks a `get_preferred_weapon_type` hook first: a mob keeps a weapon of its preferred type over one that is not, and takes a preferred weapon over a non-preferred held one, regardless of attack damage. The base `Mob` returns no preference, so this is inert until a mob overrides it (skeleton -> `skeleton_preferred_weapons`, and so on), but the foundation is there. Covered by two tests using a stand-in mob that prefers bows: it keeps a bow over a diamond sword, and takes a bow over a held stone sword.

steel-core/src/entity/mob/mod.rs
```diff
@@ -570,6 +582,304 @@ pub trait Mob: LivingEntity + Leashable {
...
+    /// [`equip_item_if_possible`](Self::equip_item_if_possible), then shrinks the
+    /// source stack by however much was equipped, only removing the item entity
+    /// once nothing is left.
+    fn pick_up_item(&self, world: &Arc<World>, item_entity: &ItemEntity) {
+        let equipped = self.equip_item_if_possible(item_entity.get_item());
+        if equipped.is_empty() {
+            return;
+        }
+
+        let count = equipped.count();
```
kdcokenny, line 615: One last thing this time I promise. Vanilla calls LivingEntity.onItemPickup here before sending the take packet. That eventually triggers THROWN_ITEM_PICKED_UP_BY_ENTITY. Since Steel does not have the advancement-criteria foundation yet, please leave a concise TODO here so that we remember to add it when that's ready.

kdcokenny (approved):

steel-core/src/entity/mob/mod.rs
darkstarworks: no need to promise(?). if something needs improving it just does. 1x, 500x, I don't mind.

darkstarworks: Done. Added a `TODO(advancements)` right at that spot noting vanilla calls `onItemPickup` before the take packet (which fires `THROWN_ITEM_PICKED_UP_BY_ENTITY`), to be wired up once Steel has the advancement-criteria foundation. Thanks for the thorough pass.

steel-core/src/entity/mob/tests.rs
```diff
@@ -4,12 +4,13 @@ use glam::DVec3;
...
 use steel_registry::item_stack::ItemStack;
 use steel_registry::vanilla_entities;
+use steel_registry::vanilla_item_tags::ItemTag;
 use steel_registry::vanilla_items;
 use steel_registry::{
     REGISTRY, init_vanilla_registry, vanilla_attributes, vanilla_blocks, vanilla_damage_types,
 };
 use steel_utils::locks::SyncMutex;
-use steel_utils::{BlockPos, BlockStateId};
+use steel_utils::{BlockPos, BlockStateId, ChunkPos, Downcast as _, Identifier};
```
banocean, line 13: Downcast as _? Why?
```suggestion
use steel_utils::{BlockPos, BlockStateId, ChunkPos, Identifier};
```

steel-registry/src/equipment.rs
```diff
@@ -122,6 +124,29 @@ impl EquipmentSlot {
         }
     }
 
+    /// Returns vanilla `EquipmentSlot.countLimit`: the maximum stack size this
+    /// slot will hold, or `0` for no limit. Hand slots are unlimited; armor,
```
banocean, line 129: Why i32? Are there any negative values? Also why making no limit as 0, I think it would be more clear with different type of value since 0 count mean not allowing any as a limit or something so its easy to overlook

steel-core/src/entity/mob/mod.rs
```diff
@@ -570,6 +582,307 @@ pub trait Mob: LivingEntity + Leashable {
...
+            item_entity.set_item(remaining);
+        }
+    }
+
+    /// Vanilla `Mob.equipItemIfPossible`: route `item_stack` to the slot its
+    /// equippable component asks for (main hand otherwise), swap up when the
+    /// target slot already holds worse gear, and drop the replaced piece by its
+    /// drop chance. Returns the stack that was equipped, or empty when nothing
+    /// was taken.
+    fn equip_item_if_possible(&self, mut item_stack: ItemStack) -> ItemStack {
```
banocean, line 640: I know is this is vanilla's name but I don't think it follows any rust conventions with that `_if_possible` suffix 😓

banocean (changes requested):
Very minor things on my end, good work 🫡

steel-core/src/entity/mob/tests.rs
darkstarworks: Good to check. It's actually used by `downcast_ref::<ItemEntity>()` (line 995), and the `as _` just pulls the trait's methods in without binding a name we never touch. Dropping it stops the build, which is why clippy never spoke up. 🙂

steel-registry/src/equipment.rs
darkstarworks: No negatives ever (it's just `0` or `1`). `i32` is so it slots straight into `ItemStack::split(i32)` without a cast. The `0`-means-unlimited is me mirroring vanilla's `EquipmentSlot.countLimit`, but I can move it to an `Option` if that reads clearer to you.. in my opinion keeping it as is, is fine.

steel-core/src/entity/mob/mod.rs
darkstarworks: Yeah, I hear you... I kept it because every method here mirrors its vanilla name, so it stays easy to diff against + `_if_possible` is used more often (`place_dripstone_block_if_possible`). `try_` felt a touch off since it reads as a `Result` in Rust(?) but this hands back the equipped `ItemStack`. Can rename if you'd rather, but imo this is okay as is.

steel-registry/src/equipment.rs
banocean: Now I wonder why ItemStack.count is i32 :sob:, but fair it's not an issue on your end

coco875 (dismissed):

banocean (approved):

coco875 (approved):

JunkyDeveloper (approved):

PR #411, opened 2026-08-21
Food / Potions foundations

tangjuyo:
## Type of change

- [ ] Block implementation
- [x] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Checklist

- [x] Code builds w/o errors or warnings
- [x] Self-reviewed the diff
- [x] Docs updated (if applicable)
- [x] No leftover debug code / comments

PR #513, opened 2026-08-26
feat: implement vanilla fox entity core

darkstarworks:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [x] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Adds the vanilla fox passive mob (MC 26.2).

Foxes are animals that:

* Come in a red or snow coat, chosen from the biome they spawn in
* Eat and breed with sweet berries, with the kit inheriting one parent's coat
* Carry a single item in the mouth, spitting out what they held and swapping a trinket for food
* Chew over the food they carry and swallow it, effects and all
* Sleep under cover by day, sit and look around, and screech at night when nobody is near

The entity follows the existing `CowEntity` stack and includes the synced variant, the vanilla behaviour-flag byte (sitting, crouching, interested, pouncing, sleeping, faceplanted, defending), the two trusted-player slots, the fox sounds, a 0.6 baby scale, and the sweet-berry food tag. Persistence round-trips the variant, the trusted players, and the sleeping, sitting, and crouching state.

On top of the item-pickup foundation from #511 the fox carries one item in its mouth, spits out the old one and drops any extra when it grabs another, and swaps a held non-food item for food once its feeding timer has run. After half a minute of carrying that food it chews over it and swallows it, so a fox that picked up a chorus fruit teleports and one that drank a honey bottle is left holding the bottle.

Since the last review this branch also covers:

* `Fox.checkFoxSpawnRules`: foxes want their own ground, the snow, grass and podzol of the forests they live in, rather than the wider set every other animal spawns on. Sand passes the general rule and fails this one
* The server half of `Fox.tick`: water, something to chase, or a thunderstorm ends a nap; a fox stops sitting up once it is wet or asleep; and a fox that has landed face-down keeps kicking up the block it is stuck in
* The five fox goal subclasses vanilla registers, replacing the stock goals the fox was borrowing. Each wraps the shared goal and adds only what the fox changes:
  * `FoxFloatGoal` starts swimming once the water is a quarter of a block deep, rather than waiting for the depth that makes most mobs jump
  * `FoxPanicGoal` and `FoxFollowParentGoal` hold their ground while the fox is defending something it trusts
  * `FoxBreedGoal`, `FoxFloatGoal` and `FoxFollowParentGoal` clear whatever the fox was in the middle of as they start, so it cannot court while asleep or swim while sat down. Courting clears both foxes
  * `FoxLookAtPlayerGoal` does not break off to watch a player while the fox is stalking something or lying face-down

* A few more `Fox` details: the sleep goal clears every state when it ends, stops the fox jumping and holds its move control in place when it starts, and a sleeping fox gets no movement input, as in `aiStep`. A dispenser can only fill the mouth slot, loading replaces the trusted list, and the item search stops chasing once the mouth is full

The hunting, avoid, and defend goals live in the stacked follow-ups, described under Additional notes.

## How this was tested

Unit tests cover the default red variant, item pickup and vanilla health, variant round-tripping, the behaviour flags sharing one byte independently, the sweet-berry food tag, variant and state and trusted-player persistence (loading replaces the trusted list rather than adding to it), a kit taking either parent's variant, the mouth-carry swap rules, spitting the held item out to grab another, a spawned fox holding one of the vanilla candidate items, the search goal standing down with a full mouth and going after an item once the mouth is empty, the sleep goal staying usable while asleep and clearing every pose when it ends, and a dispenser only filling the mouth slot.

Eating has its own tests: swallowing on time and not before, keeping the leftover bottle, holding a trinket forever, and not eating while asleep, in mid-air, or chasing something.

The newer behaviour is covered too: the fox's spawn rule across its ground tag and light gates including the spawner case, a sleeping fox woken by prey and not by a quiet night, clearing every state flag at once, a baby fox with mob drops off still dropping its mouth item on death, and the fox floating at a depth where the shared float goal declines while the fox's own accepts, which pins the difference between them rather than just the fox's own answer.

The faceplant particle is the one part driven but not asserted: it goes out as a level event, which only reaches a watching client, so there is nothing server-side to observe. The test exercises that path and asserts the sleeping and sitting rules around it.

`cargo fmt --all --check`, `RUSTFLAGS="-Dwarnings" cargo clippy -r --all-targets --all-features`, `typos` and `RUSTFLAGS="-Dwarnings" cargo test` all pass.

## Screenshots / logs

Not applicable. Server-side entity logic is covered by the unit tests above.

## Checklist

- [x] Self-reviewed the diff

## Classes modified:

Fox

## Additional notes

A few things are worth sanity-checking during review. These are not firm proposals.

* **Eating goes through the item's own behaviour.** Steel's item behaviours return the stack the slot should end up with, so the fox writes that result straight back into its mouth. Vanilla shrinks the held stack in place and therefore only writes back a leftover, which is why the `if (!remainingFood.isEmpty())` guard has no counterpart here.
* **The mouth item now persists for free.** Since #582 moved `save_equipment` and `load_equipment` into `save_mob` and `load_mob`, the fox's held item survives chunk and vehicle saves without the fox saving anything itself, which matches vanilla's `HandItems`. Rebasing onto current master is what picked that up.
* **The fox checks light unconditionally.** The shared animal rule waives the light check for spawners; vanilla's fox rule does not, the same way the turtle's does not. Nothing drives natural spawning in the tree yet, so the rule is exercised by its test rather than in play, as the shared rule it replaces already is.
* **`BreedGoal` gained a `partner()` accessor.** Vanilla's `FoxBreedGoal.start` clears the states of both foxes, which needs the shared goal to say who it paired with. One small read-only accessor rather than a behaviour change.
* **The variant-dependent target priorities do not need a `remove_goal`.** Vanilla's `setTargetGoals` only ever adds, and its goal set ignores a second add of the same goal, so the first registration wins. The stalk PR (#603) registers the prey targeting once from `finalize_spawn` and `load_additional` at the coat's priority.

Deferred: `FoxEatBerriesGoal`, which is picking berries off a sweet berry bush and off cave vines rather than the held-item path, in its own PR. `FoxStrollThroughVillageGoal` waits on village POI, and the wolf, polar bear, rabbit and schooling-fish goals wait on those mobs. Each carries a TODO at its vanilla priority.

### DIFF
--- a/steel-core/src/entity/entities/mobs/passive/chicken/mod.rs
+++ b/steel-core/src/entity/entities/mobs/passive/chicken/mod.rs
@@ -525,7 +525,7 @@ impl LivingEntity for ChickenEntity {
     }
 
     fn ai_step(&self) -> Option<MoveResult> {
-        let result = self.default_ai_step();
+        let result = Mob::mob_ai_step(self);
 
         AgeableMob::tick_ageable_mob(self);
         Animal::tick_animal_love(self);
--- a/steel-core/src/entity/entities/mobs/passive/cow/mod.rs
+++ b/steel-core/src/entity/entities/mobs/passive/cow/mod.rs
@@ -347,7 +347,7 @@ impl LivingEntity for CowEntity {
     }
 
     fn ai_step(&self) -> Option<MoveResult> {
-        let result = self.default_ai_step();
+        let result = Mob::mob_ai_step(self);
 
         AgeableMob::tick_ageable_mob(self);
         Animal::tick_animal_love(self);
--- a/steel-core/src/entity/entities/mobs/passive/pig/mod.rs
+++ b/steel-core/src/entity/entities/mobs/passive/pig/mod.rs
@@ -371,7 +371,7 @@ impl LivingEntity for PigEntity {
     }
 
     fn ai_step(&self) -> Option<MoveResult> {
-        let result = self.default_ai_step();
+        let result = Mob::mob_ai_step(self);
         AgeableMob::tick_ageable_mob(self);
         Animal::tick_animal_love(self);
         result
--- a/steel-core/src/entity/entities/mobs/passive/sheep/mod.rs
+++ b/steel-core/src/entity/entities/mobs/passive/sheep/mod.rs
@@ -446,7 +446,7 @@ impl LivingEntity for SheepEntity {
     }
 
     fn ai_step(&self) -> Option<MoveResult> {
-        let result = self.default_ai_step();
+        let result = Mob::mob_ai_step(self);
 
         AgeableMob::tick_ageable_mob(self);
         Animal::tick_animal_love(self);
--- a/steel-core/src/entity/mob/mod.rs
+++ b/steel-core/src/entity/mob/mod.rs
@@ -15,6 +15,7 @@ use glam::DVec3;
 use simdnbt::borrow::NbtCompound as BorrowedNbtCompoundView;
 use simdnbt::owned::{NbtCompound, NbtTag};
 use steel_math::fast_floor;
+use steel_protocol::packets::game::CTakeItemEntity;
 use steel_registry::blocks::block_state_ext::BlockStateExt as _;
 use steel_registry::enchantment_effect::EnchantmentEffectComponent;
 use steel_registry::item_stack::ItemStack;
@@ -23,11 +24,11 @@ use steel_registry::sound_event::SoundEventRef;
 use steel_registry::vanilla_block_tags::BlockTag;
 use steel_registry::{
     REGISTRY, RegistryExt, vanilla_attributes, vanilla_damage_types, vanilla_entities,
-    vanilla_game_events,
+    vanilla_game_events, vanilla_game_rules,
 };
 use steel_utils::locks::SyncMutex;
 use steel_utils::types::{Difficulty, InteractionHand};
-use steel_utils::{BlockPos, Identifier, WorldAabb, axis::Axis};
+use steel_utils::{BlockPos, ChunkPos, Downcast as _, Identifier, WorldAabb, axis::Axis};
 
 use crate::behavior::{BLOCK_BEHAVIORS, BlockCollisionContext, ITEM_BEHAVIORS, InteractionResult};
 use crate::enchantment_helper::{self, EnchantmentDamageContext, EnchantmentPostAttackContext};
@@ -41,6 +42,7 @@ use crate::entity::ai::sensing::Sensing;
 use crate::entity::ai::walk::WalkPathEvaluator;
 use crate::entity::attribute::{AttributeModifier, AttributeModifierOperation};
 use crate::entity::damage::DamageSource;
+use crate::entity::entities::objects::items::ItemEntity;
 use crate::entity::{
     Entity, EntitySpawnReason, LivingEntity, LivingTravelInput, RemovalReason, SharedEntity,
     SpawnGroupData, WeakEntity,
@@ -65,6 +67,9 @@ const DEFAULT_ATTACK_REACH_OFFSET: f32 = 0.6;
 const RANDOM_SPAWN_BONUS_ID: Identifier = Identifier::vanilla_static("random_spawn_bonus");
 const RANDOM_SPAWN_BONUS_SCALE: f64 = 0.114_850_000_000_000_01;
 const LEFT_HANDED_SPAWN_CHANCE: f32 = 0.05;
+/// Vanilla `Mob.ITEM_PICKUP_REACH`: the per-axis distance the item-pickup search
+/// box is inflated by when a mob looks for nearby dropped items to collect.
+const ITEM_PICKUP_REACH: DVec3 = DVec3::new(1.0, 0.0, 1.0);
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 struct DropChances {
@@ -568,6 +573,89 @@ pub trait Mob: LivingEntity + Leashable {
             .drop_chances()
             .lock()
             .set_guaranteed_drop(slot);
+    }
+
+    /// Vanilla `Mob.getPickupReach`: the per-axis distance the item-pickup search
+    /// box is inflated by. Overridable so mobs with a longer reach can widen it.
+    fn get_pickup_reach(&self) -> DVec3 {
+        ITEM_PICKUP_REACH
+    }
+
+    /// Vanilla `Mob.wantsToPickUp`: whether this mob wants to collect the item it
+    /// just walked over. Defaults to whatever the mob is willing to hold.
+    fn wants_to_pick_up(&self, item_stack: &ItemStack) -> bool {
+        self.can_hold_item(item_stack)
+    }
+
+    /// Vanilla `Mob.canHoldItem`: whether this mob is willing to hold the item.
+    /// The base mob holds anything; specific mobs narrow this down.
+    fn can_hold_item(&self, _item_stack: &ItemStack) -> bool {
+        true
+    }
+
+    /// Vanilla `Mob.pickUpItem`: take a dropped item into this mob's equipment.
+    ///
+    /// TODO(equip-armor): port the full vanilla `equipItemIfPossible` (armor-slot
+    /// selection via the item's equippable slot, the `canReplaceCurrentItem`
+    /// preference between held and dropped gear, and the drop-chance drop of any
+    /// replaced item) once armor-wearing mobs land. Until then this covers the
+    /// common case of taking an item into an empty main hand.
+    fn pick_up_item(&self, world: &Arc<World>, item_entity: &ItemEntity) {
+        let item_stack = item_entity.get_item();
+        if item_stack.is_empty()
+            || self.has_item_in_slot(EquipmentSlot::MainHand)
+            || !self.can_hold_item(&item_stack)
+        {
+            return;
+        }
+
+        let count = item_stack.count();
+        self.living_base()
+            .equipment()
+            .lock()
+            .set(EquipmentSlot::MainHand, item_stack);
+        self.set_guaranteed_drop(EquipmentSlot::MainHand);
+        self.set_persistence_required();
+
+        let chunk_pos = ChunkPos::from_entity_pos(item_entity.position());
+        world.broadcast_to_nearby(
+            chunk_pos,
+            CTakeItemEntity::new(item_entity.id(), self.id(), count),
+            None,
+        );
+        item_entity.set_removed(RemovalReason::Discarded);
+    }
+
+    /// Vanilla `Mob.aiStep` looting loop: while this mob may pick up loot and the
+    /// `mobGriefing` game rule allows it, collect nearby dropped items in reach.
+    fn tick_looting(&self) {
+        let Some(world) = self.level() else {
+            return;
+        };
+        if !self.can_pick_up_loot()
+            || !Entity::is_alive(self)
+            || !world.get_game_rule(&vanilla_game_rules::MOB_GRIEFING)
+        {
+            return;
+        }
+
+        let reach = self.get_pickup_reach();
+        let search_box = self.bounding_box().inflate_xyz(reach.x, reach.y, reach.z);
+        for entity in world.get_entities_in_aabb(&search_box) {
+            let Some(item_entity) = entity.downcast_ref::<ItemEntity>() else {
+                continue;
+            };
+            let item = item_entity.get_item();
+            if item_entity.is_removed()
+                || item.is_empty()
+                || item_entity.has_pickup_delay()
+                || !self.wants_to_pick_up(&item)
+            {
+                continue;
+            }
+
+            self.pick_up_item(&world, item_entity);
+        }
     }
 
     fn drop_custom_death_loot_mob(&self, _source: &DamageSource, killed_by_player: bool) {
@@ -1099,6 +1187,7 @@ pub trait Mob: LivingEntity + Leashable {
         self.tick_move_control();
         self.tick_look_control();
         self.tick_jump_control();
+        self.tick_looting();
     }
 
     fn tick_path_navigation(&self) {
--- a/steel-core/src/entity/mob/mod.rs
+++ b/steel-core/src/entity/mob/mod.rs
@@ -26,8 +26,8 @@ use steel_registry::loot_table::LootTableRef;
 use steel_registry::sound_event::SoundEventRef;
 use steel_registry::vanilla_block_tags::BlockTag;
 use steel_registry::{
-    REGISTRY, RegistryExt, vanilla_attributes, vanilla_damage_types, vanilla_entities,
-    vanilla_game_events, vanilla_game_rules,
+    REGISTRY, RegistryExt, TaggedRegistryExt, vanilla_attributes, vanilla_damage_types,
+    vanilla_entities, vanilla_game_events, vanilla_game_rules,
 };
 use steel_utils::locks::SyncMutex;
 use steel_utils::types::{Difficulty, InteractionHand};
@@ -748,9 +748,16 @@ pub trait Mob: LivingEntity + Leashable {
         }
     }
 
-    /// Vanilla `Mob.compareWeapons`: prefer the piece dealing more attack damage,
-    /// then the equal-item tiebreak. The base mob has no preferred weapon type, so
-    /// that branch is omitted here.
+    /// Vanilla `Mob.getPreferredWeaponType`: the item tag a mob favors for its
+    /// main hand regardless of raw damage (a skeleton keeps its bow over a
+    /// stronger sword). The base mob has no preference; specific mobs override it.
+    fn get_preferred_weapon_type(&self) -> Option<Identifier> {
+        None
+    }
+
+    /// Vanilla `Mob.compareWeapons`: keep a piece of the mob's preferred weapon
+    /// type over one that is not, otherwise prefer more attack damage, then the
+    /// equal-item tiebreak.
     #[expect(
         clippy::float_cmp,
         reason = "vanilla compares approximate attribute values for exact equality"
@@ -761,6 +768,21 @@ pub trait Mob: LivingEntity + Leashable {
         current_item_stack: &ItemStack,
         slot: EquipmentSlot,
     ) -> bool {
+        if let Some(preferred_weapon_type) = self.get_preferred_weapon_type() {
+            let current_is_preferred = REGISTRY
+                .items
+                .is_in_tag(current_item_stack.item(), &preferred_weapon_type);
+            let new_is_preferred = REGISTRY
+                .items
+                .is_in_tag(new_item_stack.item(), &preferred_weapon_type);
+            if current_is_preferred && !new_is_preferred {
+                return false;
+            }
+            if !current_is_preferred && new_is_preferred {
+                return true;
+            }
+        }
+
         let new_attack_damage = self.approximate_attribute_with(
             new_item_stack,
             vanilla_attributes::ATTACK_DAMAGE,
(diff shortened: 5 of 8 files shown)


steel-core/src/entity/entities/mobs/passive/fox/goals.rs
```diff
@@ -0,0 +1,255 @@
...
+    fn controls(&self) -> GoalControls {
+        GoalControls::MOVE | GoalControls::LOOK
+    }
+
+    fn can_use(&mut self, mob: &dyn PathfinderMob) -> bool {
+        let Some(fox) = as_fox(mob) else {
+            return false;
+        };
+        // Vanilla also gates on `!alertable()`, which scans for threatening mobs.
+        // None of those entities exist yet, so a fox is never alertable.
```
banocean: Seems like a good place for TODO comment

darkstarworks: Thank you! rather than leaving `alertable` as TODO. It turns out not to depend on the missing threat mobs: the selector's else-branch alerts on any awake, non-sneaking, untrusted living entity (players, 'farm' mobs). The monster branch keys off `MobCategory`, so it just works with what's in the tree today; Wired into both the perch-and-search and sleep goals, with a test.

While there, also expanded on `hasShelter` (the `getWalkTargetValue(pos) >= 0` check) via `PathfinderMob::get_walk_target_value`.

Spots that actually wait on a foundation now have grep-able TODOs: the untamed-tamable-animal branch, the hunting/avoid/defend goals, and eating picked-up food (like chorus fruit @tangjuyo ).

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+        AnimalBase::initialize_pathfinding_malus(&mob_base);
+        let mut entity_data = FoxEntityData::new();
+        living_base.initialize_synced_data(&mut entity_data);
+
+        {
+            // Fox goals at their vanilla priorities. The stock panic, breed, follow-parent,
+            // and look-at-player goals stand in for the fox-specific variants.
+            // TODO(fox-goals): swap in the bespoke panic/breed/follow-parent/look variants
+            // (they differ only in threat- and trust-driven behaviour) and add the hunting,
+            // avoid, and defend goals once the prey and threat mobs they need exist.
```
JunkyDeveloper: Please list all missing Goals I would make a each goal a TODO line because easier to keep track.
Also ClimpOnTopOfPowderSnowGoal already exists, so please add it

JunkyDeveloper (changes requested):

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: Adding after some sleep

darkstarworks: Done. Added `ClimbOnTopOfPowderSnowGoal` at priority `0` (matching vanilla), 

and added `// TODO(fox-goals):` as requested.

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+impl Animal for FoxEntity {
+    fn animal_base(&self) -> &AnimalBase {
+        &self.animal_base
+    }
+
+    fn is_food(&self, item_stack: &ItemStack) -> bool {
+        FoxEntity::is_food(item_stack)
+    }
+
+    fn initialize_breed_offspring(&self, partner: &dyn Animal, offspring: &dyn Animal) {
```
kdcokenny, line 754: this only copies the variant, but vanilla also makes the kit trust each parent's `getLoveCause()` player if present, as long as theyre different. Steel already records `love_cause_uuid`, so player feeding drives this behavior right now.

i tested breeding two player fed foxes and the baby had no trusted player. could we add the vanilla trust behavior here too?

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+            for id in ids {
+                if let Some(uuid) = Uuid::from_int_array(&id.to_vec()) {
+                    self.add_trusted(uuid);
+                }
+            }
+        }
+    }
+}
+
+impl LivingEntity for FoxEntity {
```
kdcokenny, line 663: foxes need to drop the item in their mouth before the normal loot checks. vanilla takes the main hand item, drops it, clears the slot, then calls super.dropAllDeathLoot.
right now it goes through the generic mob equipment path, so the item is lost when mob_drops is false or the fox is a baby. i tested mob_drops=false and got 0 berries instead of 1. could we override the death loot behavior here?

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+                } else {
+                    FoxVariant::Red
+                }
+            });
+        self.set_variant(variant);
+
+        // Vanilla `populateDefaultEquipmentSlots`: a fifth of foxes spawn holding an item.
+        // (Vanilla shares the variant across a spawn group; picking it from the biome per
+        // fox gives the same uniform result, since a group shares one spawn biome.)
+        if rand::random::<f32>() < FOX_SPAWN_HELD_ITEM_CHANCE {
```
kdcokenny, line 852: this doesnt seem equivalent to vanilla spawn groups. vanilla stores the first fox's variant in `FoxGroupData`, reuses it for the whole group, disables the normal baby roll, then makes the 3rd and later foxes babies.

this delegates to the generic ageable data, which gives the 2nd and later foxes a 5% baby chance and doesnt carry the variant. could we represent the fox group data here?

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+            .lock()
+            .take(EquipmentSlot::MainHand);
+        self.spit_out_item(world, held);
+
+        let one = item_stack.split(1);
+        self.living_base()
+            .equipment()
+            .lock()
+            .set(EquipmentSlot::MainHand, one);
+        self.set_guaranteed_drop(EquipmentSlot::MainHand);
```
kdcokenny, line 906: vanilla resets ticksSinceEaten to 0 after every successful pickup, but this never resets it. the old timer survives when the fox picks up a new item, so the swap timing seems to be wrong.

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+        self.refresh_dimensions();
+    }
+}
+
+impl Animal for FoxEntity {
+    fn animal_base(&self) -> &AnimalBase {
+        &self.animal_base
+    }
+
+    fn is_food(&self, item_stack: &ItemStack) -> bool {
```
kdcokenny, line 746: fox still needs to override `play_eating_sound`. this is separate from the held item eating TODO because feeding an adult or baby through `Animal` already calls this hook. the default does nothing, while vanilla plays `ENTITY_FOX_EAT`

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+            Arc::downgrade(world),
+        );
+        let _ = world.try_add_entity(Arc::new(item));
+    }
+
+    /// Returns whether no player is close enough to suppress the fox screech.
+    fn no_player_within_screech_range(&self, world: &Arc<World>) -> bool {
+        let search = self.bounding_box().inflate(FOX_SCREECH_PLAYER_RANGE);
+        world
+            .get_entities_in_aabb_matching(&search, |entity| {
```
kdcokenny, line 431: this counts spectators as nearby players. vanilla uses EntitySelector.NO_SPECTATORS, so a spectator shouldnt stop a fox from screeching

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+
+    fn tick_path_navigation(&self) {
+        PathfinderMob::tick_pathfinder_path_navigation(self);
+    }
+
+    fn custom_server_ai_step(&self) {
+        Animal::custom_server_ai_step_animal(self);
+    }
+
+    fn ambient_sound(&self) -> Option<SoundEventRef> {
```
kdcokenny, line 808: vanilla overrides playAmbientSound and plays the screech at volume 2.0. this uses the generic ambient path, which plays it at 1.0. could we add the fox playback override too?

kdcokenny (changes requested):
there are so many todos here im not sure if this should wait until the foundations are set or if it's good to merge when all of the issues are fixed..

darkstarworks:
> there are so many todos here im not sure if this should wait until the foundations are set or if it's good to merge when all of the issues are fixed..

Let me get back to you on this in a day if that's alright

kdcokenny:
> > there are so many todos here im not sure if this should wait until the foundations are set or if it's good to merge when all of the issues are fixed..
> 
> Let me get back to you on this in a day if that's alright

yeah no rush

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: A few lines down from there:
```rust
    );
    item_entity.set_removed(RemovalReason::Discarded);
    *self.ticks_since_eaten.lock() = 0;
}
```

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: Good catch. `initialize_breed_offspring` will add each parent's `love_cause_uuid` to the kit's trusted slots when it is present and the two differ, matching vanilla. The slots and `love_cause` are already in the tree, so it is a small addition.

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: You're right; the mouth item goes through the generic equipment path today. I'll override `drop_custom_death_loot_mob` to drop the main-hand stack and clear the slot before the loot-table gate, so it still drops when `mob_drops` is false or the fox is a baby.

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: Agreed, the generic ageable data drops the shared variant and the group baby rule. I'll represent fox group data that stores the first fox's variant, reuses it for the group, skips the normal baby roll, and makes the third and later foxes babies. It touches the shared `SpawnGroupData` enum, so I [split it into its own PR](https://github.com/darkstarworks/SteelMC/pull/1).

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: Nice catch. The `Animal` feed path already calls the hook, so the fox just needs to override `play_eating_sound` to play `ENTITY_FOX_EAT`. Adding it.

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: Right, `no_player_within_screech_range` counts spectators. I'll exclude them to match `NO_SPECTATORS`, so a spectator no longer suppresses the screech.

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: Good point, the screech should play at `2.0`. Steel plays ambient sound at the entity-wide `sound_volume()` right now, so I'll add a small `play_ambient_sound` override hook and set the fox screech volume there without touching its other sounds.

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: Done in `10cb714d`. The kit now trusts each parent's love-cause player, and only once when the same player fed both.

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: Fixed in `4b4a7f45`. The fox drops its mouth item on death before the loot gate now, so it survives `mob_drops = false` and babies, via a new unconditional `drop_custom_death_equipment` hook.

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: Added in `7ba71288`. The fox overrides `play_eating_sound` to play `ENTITY_FOX_EAT`.

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: Done in `2e01deb7`. The screech check skips spectators now, matching `NO_SPECTATORS`.

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: Done in `404ee785`. A `play_ambient_sound` override plays the screech at `2.0`; other ambient sounds keep the normal path.

darkstarworks:
> > > there are so many todos here im not sure if this should wait until the foundations are set or if it's good to merge when all of the issues are fixed..
> > 
> > 
> > Let me get back to you on this in a day if that's alright
> 
> yeah no rush

The changes you proposed are done, as well as almost all features that were previously TODOs:

- [feat: implement the vanilla fox spawn group](https://github.com/darkstarworks/SteelMC/pull/1)
- [feat: gate the fox move and look controls and add the faceplant goal](https://github.com/darkstarworks/SteelMC/pull/2)
- [feat: make a fox flee untrusted players](https://github.com/darkstarworks/SteelMC/pull/3)
- [feat: give the fox its seek-shelter goal](https://github.com/darkstarworks/SteelMC/pull/4)
- [feat: implement the fox pounce physics and goal](https://github.com/darkstarworks/SteelMC/pull/5)
- [feat: give the fox its stalk-prey hunt and chicken targeting](https://github.com/darkstarworks/SteelMC/pull/6)
- [feat: give the fox its melee attack and defend-trusted-target goals](https://github.com/darkstarworks/SteelMC/pull/7)

remaining (blocked):

- Eating held food / `FoxEatBerriesGoal`: Awaits #411 
- Other prey (Rabbit, schoolin gfish, baby turtle): Not implemented / No PR yet
- Wolf / PolarBear (avoid goals): Not implemented / No PR yet
- and `StrollThroughVillage`: Awaits #249

darkstarworks:
> > > > there are so many todos here im not sure if this should wait until the foundations are set or if it's good to merge when all of the issues are fixed..
> > > 
> > > 
> > > Let me get back to you on this in a day if that's alright
> > 
> > 
> > yeah no rush
> 
> The changes you proposed are done, as well as almost all features that were previously TODOs:
>
> * [feat: implement the vanilla fox spawn group](https://github.com/darkstarworks/SteelMC/pull/1)
> * [feat: gate the fox move and look controls and add the faceplant goal](https://github.com/darkstarworks/SteelMC/pull/2)
> * [feat: make a fox flee untrusted players](https://github.com/darkstarworks/SteelMC/pull/3)
> * [feat: give the fox its seek-shelter goal](https://github.com/darkstarworks/SteelMC/pull/4)
> * [feat: implement the fox pounce physics and goal](https://github.com/darkstarworks/SteelMC/pull/5)
> * [feat: give the fox its stalk-prey hunt and chicken targeting](https://github.com/darkstarworks/SteelMC/pull/6)
> * [feat: give the fox its melee attack and defend-trusted-target goals](https://github.com/darkstarworks/SteelMC/pull/7)
> 
> remaining (blocked):
> 
> * Eating held food / `FoxEatBerriesGoal`: Awaits [Food / Potions foundations #411](https://github.com/Steel-Foundation/SteelMC/pull/411)
> * Other prey (Rabbit, schoolin gfish, baby turtle): Not implemented / No PR yet
> * Wolf / PolarBear (avoid goals): Not implemented / No PR yet
> * and `StrollThroughVillage`: Awaits [Brain system and villager #249](https://github.com/Steel-Foundation/SteelMC/pull/249)

Please let me know if those 7 PRs look good.

And if doing it like this is okay, or whether I shouldn't have done stacked PRs

coco875 (dismissed):

JunkyDeveloper:
> > > > > there are so many todos here im not sure if this should wait until the foundations are set or if it's good to merge when all of the issues are fixed..
> > > > 
> > > > 
> > > > Let me get back to you on this in a day if that's alright
> > > 
> > > 
> > > yeah no rush
> > 
> > 
> > The changes you proposed are done, as well as almost all features that were previously TODOs:
> > 
> > * [feat: implement the vanilla fox spawn group](https://github.com/darkstarworks/SteelMC/pull/1)
> > * [feat: gate the fox move and look controls and add the faceplant goal](https://github.com/darkstarworks/SteelMC/pull/2)
> > * [feat: make a fox flee untrusted players](https://github.com/darkstarworks/SteelMC/pull/3)
> > * [feat: give the fox its seek-shelter goal](https://github.com/darkstarworks/SteelMC/pull/4)
> > * [feat: implement the fox pounce physics and goal](https://github.com/darkstarworks/SteelMC/pull/5)
> > * [feat: give the fox its stalk-prey hunt and chicken targeting](https://github.com/darkstarworks/SteelMC/pull/6)
> > * [feat: give the fox its melee attack and defend-trusted-target goals](https://github.com/darkstarworks/SteelMC/pull/7)
> > 
> > remaining (blocked):
> > 
> > * Eating held food / `FoxEatBerriesGoal`: Awaits [Food / Potions foundations #411](https://github.com/Steel-Foundation/SteelMC/pull/411)
> > * Other prey (Rabbit, schoolin gfish, baby turtle): Not implemented / No PR yet
> > * Wolf / PolarBear (avoid goals): Not implemented / No PR yet
> > * and `StrollThroughVillage`: Awaits [Brain system and villager #249](https://github.com/Steel-Foundation/SteelMC/pull/249)
> 
> Please let me know if those 7 PRs look good.
> 
> And if doing it like this is okay, or whether I shouldn't have done stacked PRs

Wait, these are all on your side so on your profile, should we review these also by hand?

darkstarworks:
Done, they are all on the project now:

* #598 fox spawn group
* #599 move and look control gating, plus the faceplant goal
* #600 flee untrusted players
* #601 seek-shelter goal
* #602 pounce physics and goal
* #603 stalk-prey hunt and chicken targeting
* #604 melee attack and defend-trusted-target

They each target `master` because there is no `fox-entity` branch here to stack on, so every diff carries #513's commits along with its own change. Each PR is a single commit (#604 is two), so the commit tab shows the actual change, 82 to 273 lines each. A rebase drops the overlap once #513 lands.

Dependency order: #598 to #602 all sit directly on #513 and are independent of each other; #603 sits on #602, and #604 on #603.

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
JunkyDeveloper, line 1: You can now add eating, with the newest merge of food foundation

Hope tomorrow I can do the 3 PRs which I want to finish also please ping me on discord if you finished turtle and fox, so we can merge them soon, will be from Thursday not active

JunkyDeveloper (changes requested):

steel-core/src/entity/entities/mobs/passive/fox/mod.rs
darkstarworks: Thank you for the update. Fox eats what it holds now, chewing from 560 ticks with the `FoxEat` event, then swallows at 600 through the item's own behaviour, so a picked-up chorus fruit teleports it and a leftover bottle stays in its mouth. 
Gated like vanilla on awake, on the ground, and no target.

`FoxEatBerriesGoal` still carries a TODO: Picking berries off a bush rather than the held-item path, so that lands later as a PR

PR #528, opened 2026-08-27
feat: allow dropping items from Creative Menu

Laptop59:
## Type of change

- [x] Entity implementation

## Description

The dropping of items from the Creative Menu was left unimplemented. This PR implements it, because we now have the functions necessary to allow item dropping. It replicates Vanilla behavior with a dropping throttler.

## How this was tested

1. Join the server.
2. Switch to Creative Mode.
3. Drop items **with the Creative Menu open**. The items should drop normally. Also try dragging them before dropping them.
4. Take a few stacks of items in your hot-bar, and just hold the Drop key (by default `Q`) on them, placing your cursor on the hot-bar slots one by one. Eventually, you will reach throttling, but that is client-side. A throttler also exists server-side, but it is not triggered due to the client-side one.

## Screenshots / logs

N/A

## Checklist

- [x] Self-reviewed the diff

## Classes modified:
- Player


steel-core/src/player/chat/mod.rs
```diff
@@ -110,11 +109,12 @@ impl ChatState {
...
 
 impl Player {
-    /// Decays the per player chat and command spam counters once per server tick
-    pub fn tick_spam_throttlers(&self) {
+    /// Decays the throttlers of the player once per server tick.
+    pub fn tick_throttlers(&self) {
         let mut chat = self.chat.lock();
         chat.chat_spam_throttler.tick();
         chat.command_spam_throttler.tick();
+        self.drop_spam_throttler.lock().tick();
```
kdcokenny, line 120: Could we release the chat guard before acquiring drop_spam_throttler? This currently nests two unrelated mutexes, creating a future lock-order risk. A small scope around the two chat-throttler ticks would avoid that without changing behavior.

kdcokenny (commented):
oh also, i think a unit test proving 75 same-tick creative drop packets spawn only 74 drops, plus the statistics behavior would be nice

steel-core/src/player/chat/mod.rs
Laptop59: Something new for me to be careful about.

kdcokenny (approved):

steel-core/src/player/chat/mod.rs
```diff
@@ -110,11 +109,14 @@ impl ChatState {
...
-        chat.chat_spam_throttler.tick();
-        chat.command_spam_throttler.tick();
+    /// Decays the throttlers of the player once per server tick.
+    pub fn tick_throttlers(&self) {
+        {
+            let mut chat = self.chat.lock();
+            chat.chat_spam_throttler.tick();
+            chat.command_spam_throttler.tick();
+        }
+        self.drop_spam_throttler.lock().tick();
```
coco875:
```suggestion
        let mut chat = self.chat.lock();
        chat.chat_spam_throttler.tick();
        chat.command_spam_throttler.tick();
        drop(chat)
        self.drop_spam_throttler.lock().tick();
```
why not something like that ?

coco875 (changes requested):

steel-core/src/player/chat/mod.rs
Laptop59: > why not something like that ?

It's quite subjective, but sure

coco875 (approved):

suprohub (approved):

JunkyDeveloper (approved):

PR #545, opened 2026-08-28
perf: batch outbound writes with whole-buffer encryption and atomic bundles

carabistouflette:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [x] Performance improvement

## Description

Overhauls the Java outbound write path. It also adds the criterion benchmarks used for the
measurements below (`steel-protocol/benches/packet_codec.rs`: whole-buffer encryption
throughput, and per-packet-flush vs batched-single-flush socket write patterns), kept in-tree
for future regression tracking:

- `StreamEncryptor` encrypts whole buffers at once (mirroring vanilla `CipherBase.encipher`) instead of running one async `poll_write` per byte, and retains partially written encrypted bytes across polls so a partial socket write can never desync the CFB8 stream.
- New `OutboundPacket::Bundle` variant: bundle delimiters and sub-packets are queued as a single channel message, eliminating the multi-producer interleaving race where another packet could land between a bundle's delimiters.
- `TCPNetworkEncoder::write_packets` writes a batch of packets with a single flush; both sender tasks (pre-play and play) drain up to 128 queued packets per writer-lock acquisition via `recv_many`.
- A batch write that fails partway drops the encoder instead of returning it to the shared writer slot: mid-packet CFB8 cipher state must never be reused. `write_outbound_batch`/`OUTBOUND_BATCH_SIZE` are shared between the play and pre-play sender loops in `player::connection`.

## How this was tested

- `cargo test -p steel-core --lib connection::java` (16 tests, incl. new `send_encoded_bundle_sends_one_atomic_bundle_message`), `cargo test -p steel-core --lib player::connection` (adds `failed_batch_write_discards_the_encoder` and `successful_batch_write_restores_the_encoder` for the poison-on-failure behavior), `cargo test -p steel-protocol --lib` (new `encrypts_whole_buffers_like_vanilla_cipher` and `retains_unwritten_encrypted_bytes_across_partial_writes`), `cargo test -p steel-login`, `cargo clippy -r --all-targets` clean.
- Benchmarks: numbers and replication steps in [Measurements](#measurements) below.

## Measurements

Run back-to-back on the same idle machine (Linux x86_64 laptop, Intel Core Ultra 7 258V, Rust
nightly, criterion defaults: 3 s warmup, 5 s measurement, 100 samples; loopback TCP with
`TCP_NODELAY` and a concurrent drain task). Absolute numbers are machine-specific; the ratios
are what matter.

Outbound write pattern, 128 encrypted `CKeepAlive` packets per batch over loopback:

| sender write path | encryptor | mean time / batch | throughput |
|---|---|---|---|
| `master` (36156e984): write + flush per packet | byte-at-a-time | 238.6 µs | 0.54 Melem/s |
| this PR: write + flush per packet | whole-buffer | 223.1 µs | 0.57 Melem/s |
| this PR: batched, single flush | whole-buffer | **27.9 µs** | **4.59 Melem/s** |

Batching alone gives **~8.6x** more outbound throughput per writer-lock acquisition vs the
`master` send path — one write syscall per batch instead of one per packet. The encryptor
change contributes ~0: the per-packet rows show it is within noise of `master`.

Whole-buffer CFB8 encryption throughput into a non-backpressuring sink (isolates the encryptor
from the socket):

| buffer size | `master` byte-at-a-time | this PR whole-buffer |
|---|---|---|
| 64 B | 44.8 MiB/s | 45.0 MiB/s |
| 512 B | 45.6 MiB/s | 47.8 MiB/s |
| 4096 B | 48.1 MiB/s | 47.4 MiB/s |
| 64 KiB | 48.1 MiB/s | 48.0 MiB/s |

Throughput parity (AES dominates; ≤ run-to-run noise). The encryption win is elsewhere: one
task reschedule per buffer instead of per byte when the socket backpressures, and no dropped
bytes when the inner writer returns `Ok(0)`.

### How to replicate

On this branch (all three arms above except the `master` row):

```bash
cargo bench -p steel-protocol --bench packet_codec
```

For the `master` baseline row, run the same bench against the commit this PR branches from.
The bench needs `write_packets`, which only exists on this branch, so copy it onto a `master`
worktree and drop the batched arm:

```bash
git worktree add --detach /tmp/steel-master 36156e984
mkdir -p /tmp/steel-master/steel-protocol/benches
git show pr1-batched-outbound-writes:steel-protocol/benches/packet_codec.rs \
  > /tmp/steel-master/steel-protocol/benches/packet_codec.rs
cat >> /tmp/steel-master/steel-protocol/Cargo.toml <<'EOF'

[dev-dependencies]
criterion = { workspace = true, features = ["async_tokio"] }

[[bench]]
name = "packet_codec"
harness = false
EOF
# delete the `group.bench_function("batched_single_flush", ..)` block from
# /tmp/steel-master/steel-protocol/benches/packet_codec.rs (`write_packets` does not exist on master)
cd /tmp/steel-master && cargo bench -p steel-protocol --bench packet_codec
```

## Screenshots / logs

N/A (no visual surface).

## Checklist

- [x] Self-reviewed the diff
- [ ] Docs updated (if applicable) — N/A, no docs cover this area yet

## Additional notes

- Stack: `pr2-outbound-byte-budget` is stacked on this branch.


steel-core/src/player/connection/java.rs
```diff
@@ -848,47 +843,46 @@ impl JavaConnection {
...
-                        select! {
-                            biased;
-                            () = self.wait_for_close() => {
+                    let has_disconnect = batch
+                        .iter()
+                        .any(|outbound| matches!(outbound, OutboundPacket::Disconnect(_)));
+                    // Bundles stay contiguous, and everything up to (and including) a
+                    // disconnect is written under a single writer lock, in order. The batch
+                    // is drained by value, so it is empty once written.
+                    match write_outbound_batch(&self.network_writer, &mut batch).await {
```
kdcokenny, line 867: what happens if the client stops reading while this batch is being written? cancellation is only checked outside this await now, so the write can stay pending while holding the writer. as far as I can tell, kicks and timeouts cant stop it, and shutdown can get stuck waiting for the task. the login sender has the same issue in tcp_client.rs:381

steel-protocol/src/utils.rs
```diff
@@ -156,66 +164,99 @@ impl<W: AsyncWrite + Unpin> StreamEncryptor<W> {
...
-                        Err(err) => return Poll::Ready(Err(err)),
+                Poll::Ready(Err(err)) => return Poll::Ready(Err(err)),
+                Poll::Pending => {
+                    if this.written == 0 {
+                        // Keep the encrypted bytes buffered for the retry.
+                        return Poll::Pending;
                     }
+                    // The written prefix is consumed; the remainder stays buffered and is
+                    // flushed by the next call.
+                    return Poll::Ready(Ok(this.written));
```
kdcokenny: this cant return only this.written after encrypting and keeping the whole buffer. AsyncWrite only says that prefix was consumed, so the caller can drop the rest and write different bytes next. this still sends the old ciphertext and reports the new bytes as consumed
i reproduced it by writing abc, getting Ok(1), then writing XY. the stream contained abc instead of aXY. poll_flush and poll_shutdown also dont drain this buffer. can we fix the buffering so it follows the AsyncWrite contract?

kdcokenny (changes requested):

steel-core/src/player/connection/java.rs
carabistouflette: Fixed in eaab77f2b: `write_outbound_batch` now wraps the whole batch in a 30 s deadline (`OUTBOUND_WRITE_TIMEOUT`). If the client stops reading, the write future is dropped, the taken encoder stays dropped (mid-packet CFB8 state must not be reused), and the sender treats it like any failed batch, so kicks, keepalive timeouts, and shutdown all proceed. The login sender gets the same fix for free since it shares `write_outbound_batch`. Healthy clients drain a batch in microseconds, so only a genuinely stalled peer ever hits the deadline; `finish_disconnect` keeps its separate 1 s best-effort budget on top.

steel-protocol/src/utils.rs
carabistouflette: fixed in eaab77f2b. `StreamEncryptor` now uses commit-all buffering (`BufWriter` semantics): every `poll_write` reports the whole buffer consumed and retains its ciphertext until the inner writer accepts it, so a sink that eats a prefix and discards the rest can never desync the stream, and the retained bytes are drained by later writes, `poll_flush`, and `poll_shutdown` instead of sitting there. Your repro (write `abc`, `Ok(1)`, then `XY`) now yields `abcXY`. Pinned by `backpressured_sink_keeps_separate_writes_sequential` and `shutdown_drains_retained_ciphertext` in `steel-protocol/src/utils.rs`.

coco875:
I would like an confirmation that it don't do something similar to with https://github.com/Steel-Foundation/SteelMC/pull/473

carabistouflette:
@coco875 I checked #473 and there is some overlap but #545 uses a smaller shared-writer design

carabistouflette:
its 30 second batch timeout drops the encoder on stalls so the sender won’t get wedged

PR #547, opened 2026-08-28
feat: vanilla framing parity, read-path hardening, and malformed packet kicks

carabistouflette:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [x] Performance improvement

## Description

Hardens the read path to vanilla framing parity:

- `get_raw_packet` decodes the packet length `VarInt` with a manual 3-byte cap, matching vanilla `Varint21FrameDecoder` (`MAX_VARINT21_BYTES = 3`, "length wider than 21-bit"): lengths wider than 21 bits and zero-length frames ("Frame length cannot be zero") are rejected with `PacketError::MalformedLength` before consuming unbounded input (previously an unconstrained `VarInt::read_async` accepted up to 5 bytes). The now-provably-dead `packet_len > MAX_PACKET_SIZE` read check is removed (a 3-byte `VarInt` cannot exceed 2^21-1; vanilla has no extra read-side maximum), along with the `PacketError::OutOfBounds` variant that only it constructed.
- Compressed packets whose declared uncompressed size is nonzero and below the server threshold are rejected with a new `PacketError::BelowThreshold` before inflation, matching vanilla `CompressionDecoder` ("Badly compressed packet - size of N is below server threshold"). The check only applies to nonzero payloads, so legitimate uncompressed packets still pass.
- Malformed-packet kicks: pre-play connections are kicked with `multiplayer.disconnect.invalid_packet` when packet processing fails; play connections now do the same (a failed `process_packet` leaves the stream framing intact, so the vanilla invalid-packet disconnect still reaches the client instead of a bare close). Frame-level read failures close without a packet (a desynced stream cannot parse one). Unrecognized immediate packets log at debug level.

## How this was tested

- `cargo test -p steel-protocol --lib` — new vanilla-parity tests: `rejects_length_varint_wider_than_21_bits`, `rejects_zero_length_frames`, `rejects_compressed_size_below_server_threshold` (54 tests total). Full `steel-core` (2466 tests) and `steel-login` suites pass; `cargo clippy -r --all-targets` clean.
- Benchmark: numbers and replication steps in [Measurements](#measurements) below.

## Measurements

Back-to-back criterion runs on the same idle machine (Linux x86_64 laptop, Intel Core Ultra 7 258V, Rust nightly, criterion defaults: 3 s warmup, 5 s measurement, 100 samples), decoding 256 framed packets per iteration from an in-memory buffer:

| reader | mean time / 256 packets | throughput |
|---|---|---|
| `master` (36156e984): `VarInt::read_async` | 6.609 µs | 38.7 Melem/s |
| this PR: manual 3-byte length loop | **6.285 µs** | **40.7 Melem/s** |

**+5.2%** decode throughput (confidence intervals do not overlap; absolute numbers are machine-specific).

### How to replicate

On this branch:

```bash
cargo bench -p steel-protocol --bench packet_reader_framing
```

For the `master` row, copy the bench onto a `master` worktree (it uses only pre-existing APIs, so it compiles unchanged):

```bash
git worktree add --detach /tmp/steel-master 36156e984
mkdir -p /tmp/steel-master/steel-protocol/benches
git show pr3-read-path-hardening:steel-protocol/benches/packet_reader_framing.rs \
  > /tmp/steel-master/steel-protocol/benches/packet_reader_framing.rs
cat >> /tmp/steel-master/steel-protocol/Cargo.toml <<'EOF'

[dev-dependencies]
criterion = { workspace = true, features = ["async_tokio"] }

[[bench]]
name = "packet_reader_framing"
harness = false
EOF
cd /tmp/steel-master && cargo bench -p steel-protocol --bench packet_reader_framing
```

## Screenshots / logs

N/A.

## Checklist

- [x] Self-reviewed the diff
- [ ] Docs updated (if applicable) — N/A

## Additional notes

- The bench lives in `steel-protocol/benches/packet_reader_framing.rs` (renamed from `packet_codec.rs`) so it cannot add/add-conflict with #545's `packet_codec.rs` bench; the two PRs now merge cleanly in either order.
- Vanilla references verified against the local `minecraft-src` tree: `Varint21FrameDecoder.java` (3-byte cap, zero-length rejection) and `CompressionDecoder.java` (nonzero below-threshold rejection before inflation).
- Adds a `criterion` `async_tokio` dev-feature for `steel-protocol`.

### DIFF
--- a/steel-core/src/player/connection/java.rs
+++ b/steel-core/src/player/connection/java.rs
@@ -811,7 +811,7 @@ impl JavaConnection {
                     .lock()
                     .on_chunk_batch_received_by_client(packet.desired_chunks_per_tick);
             }
-            ImmediatePlayPacket::Unknown(id) => log::info!("play packet id {id} is not known"),
+            ImmediatePlayPacket::Unknown(id) => log::debug!("play packet id {id} is not known"),
         }
     }
 
@@ -830,10 +830,14 @@ impl JavaConnection {
                     match packet {
                         Ok(packet) => {
                             if let Some(player) = self.player.upgrade()
-                                && let Err(err) = self.process_packet(packet, player, &server) {
-                                log::warn!(
-                                    "Failed to get packet from client {}: {err}",
-                                    self.id
+                                && let Err(err) = self.process_packet(packet, player, &server)
+                            {
+                                // A packet failed to decode or dispatch. The stream framing
+                                // is still intact, so the vanilla invalid-packet disconnect
+                                // reaches the client before the connection closes.
+                                log::warn!("Failed to process packet from client {}: {err}", self.id);
+                                self.disconnect(
+                                    translations::MULTIPLAYER_DISCONNECT_INVALID_PACKET.msg(),
                                 );
                             }
                         }
--- a/steel-protocol/src/packet_reader.rs
+++ b/steel-protocol/src/packet_reader.rs
@@ -18,9 +18,7 @@ use steel_utils::codec::VarInt;
 use steel_utils::serial::ReadFrom;
 use tokio::io::{AsyncRead, AsyncReadExt, ReadBuf};
 
-use crate::utils::{
-    Aes128Cfb8Dec, MAX_PACKET_DATA_SIZE, MAX_PACKET_SIZE, PacketError, RawPacket, StreamDecryptor,
-};
+use crate::utils::{Aes128Cfb8Dec, MAX_PACKET_DATA_SIZE, PacketError, RawPacket, StreamDecryptor};
 
 /// A reader that can decrypt data.
 pub enum DecryptionReader<R: AsyncRead + Unpin> {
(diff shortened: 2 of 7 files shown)


steel-core/src/player/connection/java.rs
```diff
@@ -830,10 +830,14 @@ impl JavaConnection {
...
                             if let Some(player) = self.player.upgrade()
-                                && let Err(err) = self.process_packet(packet, player, &server) {
-                                log::warn!(
-                                    "Failed to get packet from client {}: {err}",
-                                    self.id
+                                && let Err(err) = self.process_packet(packet, player, &server)
+                            {
+                                // A packet failed to decode or dispatch. The stream framing
+                                // is still intact, so the vanilla invalid-packet disconnect
+                                // reaches the client before the connection closes.
```
coco875, line 837: does the comment are useful

steel-protocol/src/packet_reader.rs
```diff
@@ -108,10 +106,32 @@ impl<R: AsyncRead + Unpin> TCPNetworkDecoder<R> {
...
+            if byte[0] & 0x80 == 0 {
+                break;
+            }
+        }
+
+        if packet_len == 0 {
+            Err(PacketError::MalformedLength(
+                "packet length cannot be zero".to_string(),
+            ))?;
         }
```
coco875, line 135: the reader don't emit an error if you go out of bound already ?

steel-protocol/src/packet_reader.rs
```diff
@@ -108,10 +106,32 @@ impl<R: AsyncRead + Unpin> TCPNetworkDecoder<R> {
     /// - If the packet fails to decompress.
     #[expect(clippy::cast_sign_loss)]
     pub async fn get_raw_packet(&mut self) -> Result<RawPacket, PacketError> {
```
coco875, line 108: in this function have Err(...)? are a tricky return Err disguise, can you change that even if it's a bit out of scope of your pr ?

steel-protocol/src/packet_reader.rs
```diff
@@ -108,10 +106,32 @@ impl<R: AsyncRead + Unpin> TCPNetworkDecoder<R> {
...
+                ))?;
+            }
+
+            let mut byte = [0u8; 1];
+            self.reader.read_exact(&mut byte).await?;
+            packet_len |= ((byte[0] & 0x7F) as usize) << (7 * length_bytes);
+            length_bytes += 1;
 
-        if packet_len > MAX_PACKET_SIZE {
-            Err(PacketError::OutOfBounds)?;
```
coco875, line 114: why remove out of bound check ?

coco875 (changes requested):

PR #548, opened 2026-08-28
feat: pre-play lifecycle timeouts, config keepalive, and favicon caching

carabistouflette:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [x] Performance improvement

## Description

Implements the vanilla pre-play lifecycle:

- **Login deadline**: a client still in the login phase 30 seconds after the handshake (vanilla `ServerLoginPacketListenerImpl`, `MAX_TICKS_BEFORE_LOGIN = 600`) is kicked with `multiplayer.disconnect.slow_login`. The read loop wraps packet reads in `timeout_at` while the deadline is active.
- **Configuration keep-alive**: challenges are sent every 15 seconds since the last send, and a challenge left unanswered for a full interval is a `disconnect.timeout` kick at the next boundary — vanilla `ServerCommonPacketListenerImpl.keepConnectionAlive` timing. Out-of-order answers are a timeout kick, and accepted answers smooth the round-trip latency that is seeded into the play connection (vanilla `CommonListenerCookie.latency()`). Uses the existing common `CKeepAlive`/`SKeepAlive` packets — no new packet types.
- **Status favicon caching**: the favicon is loaded once into `Server::favicon` at startup (a single warning is logged when the configured path is unreadable), removing blocking filesystem I/O from the per-ping `handle_status_request` path. The file is validated like vanilla `PngInfo`: it must be a PNG with a 64x64 `IHDR`, otherwise it is rejected with the vanilla `Couldn't load server icon` error.

## How this was tested

- `cargo test -p steel-login` (22 tests) and `cargo test -p steel-core --lib` (2467 tests) pass; `cargo clippy -r --all-targets` clean.
- New tests: `keepalive_decisions_match_vanilla_boundaries` and `keepalive_answer_smooths_latency_and_rejects_out_of_order` pin the vanilla keep-alive boundaries (15 s cadence, unanswered → timeout, out-of-order/duplicate answers rejected, latency smoothing); `load_favicon_rejects_invalid_pngs_like_vanilla` covers the favicon validation matrix.
- Manual E2E per the plan: raw socket handshake with intent=Login stalls 32 s → server logs the slow-login kick at t=30 s and closes; status ping answers immediately with the cached favicon (no disk access after startup).

## Screenshots / logs

N/A (server-log behavior only).

## Checklist

- [x] Self-reviewed the diff
- [ ] Docs updated (if applicable) — N/A

## Additional notes

- Stack: `pr4-preplay-lifecycle` is stacked on `pr3-read-path-hardening`.

> **Note:** stacked on #547 (`pr3-read-path-hardening`). Until that merges, this PR's diff includes its commits; the lifecycle changes themselves are the final commits. Incremental diff: https://github.com/carabistouflette/SteelMC/compare/pr3-read-path-hardening...pr4-preplay-lifecycle

### DIFF
--- a/steel-core/src/player/connection/java.rs
+++ b/steel-core/src/player/connection/java.rs
@@ -811,7 +811,7 @@ impl JavaConnection {
                     .lock()
                     .on_chunk_batch_received_by_client(packet.desired_chunks_per_tick);
             }
-            ImmediatePlayPacket::Unknown(id) => log::info!("play packet id {id} is not known"),
+            ImmediatePlayPacket::Unknown(id) => log::debug!("play packet id {id} is not known"),
         }
     }
 
@@ -830,10 +830,14 @@ impl JavaConnection {
                     match packet {
                         Ok(packet) => {
                             if let Some(player) = self.player.upgrade()
-                                && let Err(err) = self.process_packet(packet, player, &server) {
-                                log::warn!(
-                                    "Failed to get packet from client {}: {err}",
-                                    self.id
+                                && let Err(err) = self.process_packet(packet, player, &server)
+                            {
+                                // A packet failed to decode or dispatch. The stream framing
+                                // is still intact, so the vanilla invalid-packet disconnect
+                                // reaches the client before the connection closes.
+                                log::warn!("Failed to process packet from client {}: {err}", self.id);
+                                self.disconnect(
+                                    translations::MULTIPLAYER_DISCONNECT_INVALID_PACKET.msg(),
                                 );
                             }
                         }
--- a/steel-core/src/player/connection/java.rs
+++ b/steel-core/src/player/connection/java.rs
@@ -397,13 +397,15 @@ pub struct JavaConnection {
 }
 
 impl JavaConnection {
-    /// Creates a new `JavaConnection`.
+    /// Creates a new `JavaConnection`, seeding the smoothed latency measured during the
+    /// configuration phase (vanilla carries it in `CommonListenerCookie`).
     pub const fn new(
         outgoing_packets: UnboundedSender<OutboundPacket>,
         cancel_token: CancellationToken,
         compression: Option<CompressionInfo>,
         network_writer: JavaNetworkWriter,
         id: u64,
+        latency: u32,
         player: Weak<Player>,
     ) -> Self {
         Self {
@@ -418,7 +420,7 @@ impl JavaConnection {
                 alive_pending: false,
                 alive_id: 0,
             }),
-            latency: SyncMutex::new(0),
+            latency: SyncMutex::new(latency),
         }
     }
 
--- a/steel-login/src/tcp_client.rs
+++ b/steel-login/src/tcp_client.rs
@@ -10,7 +10,7 @@ use std::{
     io::Cursor,
     net::SocketAddr,
     sync::Arc,
-    time::Duration,
+    time::{Duration, Instant, SystemTime, UNIX_EPOCH},
 };
 
 use crossbeam::atomic::AtomicCell;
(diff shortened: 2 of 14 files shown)


steel-login/src/tcp_client.rs
```diff
@@ -578,6 +662,35 @@ impl JavaTcpClient {
...
+        if self.login_deadline_expired() {
+            return LoginOperationResult::TimedOut;
+        }
+        let incoming_event = async {
+            select! {
+                packet = reader.get_raw_packet() => IncomingEvent::Packet(packet),
+                connection_update = connection_updates_recv.recv() => {
+                    IncomingEvent::ConnectionUpdate(connection_update)
+                }
+                _ = config_keepalive_tick.tick() => IncomingEvent::KeepAliveTick,
```
Joniii11, line 682: this tick can cancel get_raw_packet after some bytes was already read then remaining packet body is parsed as new frame anilla keeps incomplete frame bytes

steel-login/src/tcp_client.rs
```diff
@@ -482,26 +567,16 @@ impl JavaTcpClient {
 
         self.task_tracker.spawn(async move {
             let mut connection = None;
+            // One-second ticks give the vanilla 15-second send and timeout boundaries
+            // one-second resolution on a single shared timer.
+            let mut config_keepalive_tick = interval(Duration::from_secs(1));
```
Joniii11, line 572: vanilla checks config keepalive every server connection tick not with separate 1s timer timing is different here

steel-login/src/tcp_client.rs
```diff
@@ -780,6 +910,36 @@ impl JavaTcpClient {
...
 
+    /// Ticks the configuration keep-alive, executing the tracker's vanilla-timed
+    /// decision. Both writes are bounded: a wedged socket must close the connection
+    /// instead of stalling the read loop.
+    async fn tick_config_keepalive(&self) {
+        let decision = self.config_keepalive.lock().tick(Instant::now());
+        match decision {
+            KeepAliveDecision::None => {}
+            KeepAliveDecision::Send(challenge) => {
+                if timeout(
```
Joniii11, line 921: vanilla only queues this keepalive and has no 1s write timeout slow valid client can be closed too early

steel-core/src/player/connection/java.rs
```diff
@@ -811,7 +813,7 @@ impl JavaConnection {
                     .lock()
                     .on_chunk_batch_received_by_client(packet.desired_chunks_per_tick);
             }
-            ImmediatePlayPacket::Unknown(id) => log::info!("play packet id {id} is not known"),
+            ImmediatePlayPacket::Unknown(id) => log::debug!("play packet id {id} is not known"),
```
Joniii11, line 816: why is it now debug any reason for it?

steel-login/src/tcp_client.rs
```diff
@@ -192,6 +200,76 @@ impl ConnectionAction {
...
+    fn tick(&mut self, now: Instant) -> KeepAliveDecision {
+        if now.duration_since(self.last_sent) < Self::KEEP_ALIVE_INTERVAL {
+            return KeepAliveDecision::None;
+        }
+        if self.pending.is_some() {
+            return KeepAliveDecision::Timeout;
+        }
+
+        // Vanilla uses `Util.getMillis()` as the challenge id.
+        let challenge = SystemTime::now()
```
Joniii11, line 244: Util.getMillis() uses monotonic time, not unix time

steel-login/src/tcp_client.rs
```diff
@@ -192,6 +200,76 @@ impl ConnectionAction {
...
+            .duration_since(UNIX_EPOCH)
+            .expect("System time before UNIX EPOCH")
```
Joniii11, line 246: i mean hm would be an edge case where it could trigger so idk i dont lioe thaz

steel-login/src/tcp_client.rs
```diff
@@ -769,6 +882,23 @@ impl JavaTcpClient {
                     .await;
                 Ok(ConnectionAction::none())
             }
+            config::S_KEEP_ALIVE => {
+                let packet = SKeepAlive::read_packet(data)?;
```
Joniii11, line 886: this new handler accepts keepalive with extra trailing bytes because payload end is not checked vanilla rejects it before handler

Joniii11 (changes requested):

PR #557, opened 2026-08-29
refactor: Remove RawEntity and rename RawBlockEntity

### LINKED ISSUE #555: Rename `RawEntity` to `UnimplementedEntity`
`RawEntity` makes it sound like the raw/base type behind every entity, especially when we already have `EntityBase`. thats not what it is.

its only used when Steel doesnt have the actual entity implemented yet. it preserves the base data and NBT, but doesnt have proper gameplay behavior.

can we rename:

- `RawEntity` to `UnimplementedEntity`
- `RawBlockEntity` to `UnimplementedBlockEntity`
- the related functions, modules, tests, and docs

the loading fallback can still use it so we dont lose NBT. the name should just make it obvious that this isnt a proper entity implementation someone should use in place of implementing one.

kdcokenny:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [x] Breaking change
- [x] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

`RawEntity` kept unsupported entity NBT around by creating an inert runtime entity. That also meant worldgen could treat an entity Steel doesnt implement as if it was supported.

This removes `RawEntity`. Unsupported entities are now skipped when theyre loaded from disk or placed by structure templates, and structure mob spawning only creates registered mob implementations.

`RawBlockEntity` still has a useful job because it preserves block entity NBT until Steel has a typed implementation. This renames it to `UnimplementedBlockEntity`, including its module, registry methods, downcast key, call sites, tests, and docs.

Closes #555.

## How this was tested

Ran on Linux with Rust nightly-2026-08-21 against Minecraft 26.2:

- `cargo test -p steel-core unimplemented` (7 passed)
- `cargo test -p steel-core create_and_load_skips_unregistered_entity_type` (1 passed)
- `cargo test -p steel-core` (2,487 passed, 11 ignored)
- `cargo clippy -r --all-targets --all-features`
- `cargo fmt --all --check`
- `typos`
- `git diff --check`

## Screenshots / logs

N/A

## Checklist

- [x] Self-reviewed the diff

## Classes modified:

N/A

## Additional notes

This removes and renames public Rust APIs. Code using `RawEntity`, `RawBlockEntity`, or the old `*_or_raw` methods needs to update.

Unsupported saved entities are now skipped, so Steel no longer preserves their NBT. Unimplemented block entity NBT is still preserved. There is no on disk format change for data Steel still saves.


steel-core/src/entity/base/tests.rs
```diff
@@ -47,13 +47,13 @@ fn assert_f64_close(left: f64, right: f64) {
     );
 }
 
-fn raw_entity(id: i32) -> SharedEntity {
-    Arc::new(RawEntity::new(
```
coco875, line 51: the function can be just removed ? and directly use TestEntity::shared ?

coco875 (changes requested):

steel-core/src/entity/base/tests.rs
kdcokenny: yes, removed

steel-core/src/entity/registry.rs
```diff
@@ -170,25 +169,29 @@ impl EntityRegistry {
...
+        let entity_type = request.entity_type;
         let id = entity_type.id();
-        if let Some(load_factory) = self.entries.get(id).and_then(|entry| entry.load_factory) {
-            let entity = load_factory(entity_type, load);
-            Self::finish_registered_load(&entity, nbt);
-            return entity;
-        }
+        let Some(load_factory) = self.entries.get(id).and_then(|entry| entry.load_factory) else {
+            log::warn!(
+                "Skipping unsupported entity {} while loading",
```
carabistouflette, line 185: When this fires on a real save the entity is gone for good, so let's log enough to actually debug a "my pet disappeared" report later. Can we include the position and uuid here? `request` already has both, so it's just widening the message.

steel-core/src/worldgen/structure/piece_placer/mod.rs
```diff
@@ -46,6 +52,39 @@ impl StructurePiecePlacer {
     /// Vanilla template-piece placement flags: `UPDATE_CLIENTS`.
     pub(crate) const TEMPLATE_UPDATE_FLAGS: UpdateFlags = UpdateFlags::UPDATE_CLIENTS;
 
+    /// Creates a structure mob when Steel implements its concrete type.
+    ///
+    /// Unsupported types are skipped instead of leaving behaviorless entities in generated
+    /// chunks.
+    fn create_mob(
```
carabistouflette, line 59: This helper has three outcomes (no factory, factory but not a Mob, actual mob) and none of them are covered by tests yet. Could we add a small test module here? e.g. `VILLAGER` → None (unregistered), `ITEM` → None (registered but `ItemEntity` isn't a Mob), `PIG` → Some with the persistence flag set. Cheap to write now and it pins the contract before structure spawns start depending on it.

steel-core/src/worldgen/structure/piece_placer/mod.rs
```diff
@@ -46,6 +52,39 @@ impl StructurePiecePlacer {
...
+        let Some(mob) = entity.as_mob() else {
+            log::warn!(
+                "Skipping structure mob spawn for {} at {position:?}: registered implementation is not a mob",
+                entity_type.key
+            );
+            return None;
+        };
+
+        if persistence_required {
+            mob.set_persistence_required();
```
carabistouflette, line 81: Not for this PR, but noting for later: vanilla calls `finalizeSpawn(..., EntitySpawnReason.STRUCTURE, null)` right after setPersistenceRequired/snapTo for all these structure mobs (witch, cat, elder, drowned, mansion). It's a no-op today since none of them are implemented, but when they land this is the spot that needs it, maybe worth a one-line comment so it doesn't get rediscovered.

steel-core/src/chunk_saver/storage/tests/entities.rs
```diff
@@ -484,7 +484,7 @@ fn unimplemented_block_entities_preserve_nbt_through_proto_save_load() {
     let mut nbt = NbtCompound::new();
     nbt.insert("LootTable", "minecraft:chests/simple_dungeon");
     nbt.insert("LootTableSeed", 42_i64);
-    let entity = BLOCK_ENTITIES.create_and_load_owned_or_raw(
+    let entity = BLOCK_ENTITIES.create_and_load_owned_or_unimplemented(
```
carabistouflette, line 521: The block-entity side has `unimplemented_block_entities_preserve_nbt_through_proto_save_load`, but there's no counterpart for the new entity behavior: save a chunk with an unregistered entity type → load → re-save → entity is absent. Since this is the "we now silently drop saved entities" contract, I'd like it pinned by a test so the behavior is deliberate, not incidental.

carabistouflette (commented):
One nuance worth spelling out in the notes: because `spawn_swamp_hut_mob` (and the vanilla-matching flag order generally) marks `spawned = true` before the spawn attempt, any hut/monument/ruin generated while a mob type is unimplemented will never get that mob, even after it's implemented. The skip isn't retry-later, it's permanent per generated structure.

steel-core/src/entity/registry.rs
kdcokenny: i'd say it's useful but out of scope for this pr/issue

steel-core/src/worldgen/structure/piece_placer/mod.rs
kdcokenny: added coverage for the unsupported path with witch

steel-core/src/worldgen/structure/piece_placer/mod.rs
kdcokenny: do you mind making an issue for this?

steel-core/src/chunk_saver/storage/tests/entities.rs
kdcokenny: added

steel-core/src/worldgen/structure/piece_placer/mod.rs
carabistouflette: done

steel-core/src/server/tests.rs
```diff
@@ -730,7 +730,7 @@ fn domain_restore_jobs_follow_same_session_player_replacement() {
         let pearl_uuid = [7; 16];
         let root = PersistentRootVehicle {
             attach: root_uuid,
-            entity: test_persistent_entity(&vanilla_entities::MINECART, root_uuid),
+            entity: test_persistent_entity(&vanilla_entities::PIG, root_uuid),
```
coco875, line 737: why change the entity type ?

coco875 (changes requested):

steel-core/src/server/tests.rs
kdcokenny: In order to remove `RawEntity`. minecart is unimplemented so it relied on that fallback. switched it to pig bc this test needs an entity that can actually load to test the session replacement behavior

coco875 (approved):

Joniii11 (approved):
lgtm 🌵

CuzImClicks (approved):

PR #558, opened 2026-08-29
feat: implement interaction entity

Laptop59:
## Type of change

- [x] Entity implementation

## Description

Implements the [interaction entity](https://minecraft.wiki/w/Interaction) from Vanilla. This is an invisible, invincible, interactable entity whose bounding box can be customized, and keeps track of the latest interaction (right-click) and attack (left-click). Used by map and data pack makers.

## How this was tested

1. Join the server
2. Summon the entity with `/summon minecraft:interaction`

## Screenshots / logs

With hitboxes enabled (`F3`+`B`):
interactions

## Checklist

- [x] Self-reviewed the diff

## Classes modified:

- Interaction

### DIFF
--- a/steel-core/src/entity/entities/objects/technical/interaction.rs
+++ b/steel-core/src/entity/entities/objects/technical/interaction.rs
@@ -23,6 +23,9 @@ const DEFAULT_WIDTH: f32 = 1.0;
 const DEFAULT_HEIGHT: f32 = 1.0;
 const DEFAULT_RESPONSE: bool = false;
 
+const DEFAULT_DIMENSIONS: EntityDimensions =
+    InteractionEntity::dimensions_for_width_height(DEFAULT_WIDTH, DEFAULT_HEIGHT);
+
 const TAG_WIDTH: &str = "width";
 const TAG_HEIGHT: &str = "height";
 const TAG_ATTACK: &str = "attack";
@@ -56,7 +59,7 @@ impl InteractionEntity {
     #[must_use]
     pub fn new(entity_type: EntityTypeRef, id: i32, position: DVec3, world: Weak<World>) -> Self {
         Self {
-            base: EntityBase::new(id, position, entity_type.dimensions, world),
+            base: EntityBase::new(id, position, DEFAULT_DIMENSIONS, world),
             entity_type,
             entity_data: SyncMutex::new(InteractionEntityData::new()),
             attack: SyncMutex::new(None),
@@ -74,7 +77,7 @@ impl InteractionEntity {
         world: Weak<World>,
     ) -> Self {
         Self {
-            base: EntityBase::with_uuid(id, uuid, position, entity_type.dimensions, world),
+            base: EntityBase::with_uuid(id, uuid, position, DEFAULT_DIMENSIONS, world),
             entity_type,
             entity_data: SyncMutex::new(InteractionEntityData::new()),
             attack: SyncMutex::new(None),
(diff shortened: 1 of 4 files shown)


steel-core/src/entity/entities/objects/technical/interaction.rs
```diff
@@ -0,0 +1,525 @@
...
+                nbt.insert(TAG_INTERACTION, interaction);
+            }
+        }
+    }
+
+    fn load_additional(&self, nbt: BorrowedNbtCompoundView<'_, '_>) {
+        let dimensions_changed = {
+            let mut guard = self.entity_data.lock();
+
+            guard
```
kdcokenny, line 239: i tested this with `{width:2,height:3,response:1}`. vanilla loads that as `(2,3,true)`, but these helpers only accept exact `Float` or `Byte` tags so this silently becomes `(1,1,false)`. `compound.long` has the same problem for `timestamp` below. can we accept any numeric tag like vanilla does?

steel-registry/src/entity_type.rs
```diff
@@ -4,6 +4,8 @@ use steel_utils::{DowncastType, DowncastTypeKey, Identifier};
 
 use crate::{RegistryTags, blocks::behavior::PushReaction};
 
+pub const DEFAULT_EYE_HEIGHT_FACTOR: f32 = 0.85;
```
kdcokenny: was making this public on purpose? `with_default_eye_height` is the only thing using it and vanilla keeps the factor private, so this looks like it could stay private too

steel-core/src/entity/entities/objects/technical/interaction.rs
```diff
@@ -0,0 +1,525 @@
...
+    pub fn last_attack(&self) -> Option<PlayerAction> {
+        *self.attack.lock()
+    }
+
+    /// Provides the latest interaction (right-click) on this entity, if it exists.
+    pub fn last_interaction(&self) -> Option<PlayerAction> {
+        *self.interaction.lock()
+    }
+
+    const fn dimensions_for_width_height(width: f32, height: f32) -> EntityDimensions {
```
kdcokenny: what's the point of wrapping this?

kdcokenny (changes requested):

steel-core/src/entity/entities/objects/technical/interaction.rs
Laptop59: > i tested this with `{width:2,height:3,response:1}`. vanilla loads that as `(2,3,true)`, but these helpers only accept exact `Float` or `Byte` tags so this silently becomes `(1,1,false)`. `compound.long` has the same problem for `timestamp` below. can we accept any numeric tag like vanilla does?

Honestly I think other entities do the same as well; I would believe it needs a refactor across usages.

kdcokenny: > > i tested this with `{width:2,height:3,response:1}`. vanilla loads that as `(2,3,true)`, but these helpers only accept exact `Float` or `Byte` tags so this silently becomes `(1,1,false)`. `compound.long` has the same problem for `timestamp` below. can we accept any numeric tag like vanilla does?
> 
> Honestly I think other entities do the same as well; I would believe it needs a refactor across usages.

interesting, def worth checking out.

steel-registry/src/entity_type.rs
Laptop59: Yeah, I forgot why I did that.

steel-core/src/entity/entities/objects/technical/interaction.rs
Laptop59: > what's the point of wrapping this?

Yeah, there was no point.

steel-core/src/entity/entities/objects/technical/interaction.rs
kdcokenny: def to https://github.com/Steel-Foundation/SteelMC/issues/559

kdcokenny (approved):

coco875 (approved):

JunkyDeveloper (approved):

PR #560, opened 2026-08-30
feat: implement first_tick

AnthonyDuong1:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [x] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description
This PR adds the first_tick flag to entites. This flag is initialized to true when a new entity is created, and once it has been ticked (through entity_base_tick), it will be set to false.

This is required for the first tick of an entity in contact with lava. It is also needed in refresh_dimensions.

## How this was tested
I added a test to make sure an entity that is in lava but before it has been ticked returns false for is_in_lava(). Then it gets ticked and returns true for is_in_lava().

cargo test, cargo fmt, and cargo clippy all passes.

## Checklist

- [X] Self-reviewed the diff


steel-core/src/entity/tests/fall_and_fluids.rs
```diff
@@ -75,6 +75,21 @@ fn stop_fall_flying_toggles_shared_state_back_to_false() {
...
+fn lava_contact_is_ignored_until_after_first_tick() {
+    let entity = TypedTestEntity::new(1, &vanilla_entities::PIG);
+    entity
+        .base()
+        .set_fluid_contact(EntityFluidContact::from_parts(0.0, 0.25, false, false));
+
+    assert!(entity.is_first_tick());
+    assert!(!entity.is_in_lava());
+
+    entity.set_first_tick(false);
```
kdcokenny: this test doesnt actually tick the entity as far as i can tell.. it just sets `first_tick` to false manually. plz fix

kdcokenny (changes requested):

steel-core/src/entity/tests/fall_and_fluids.rs
AnthonyDuong1: i changed it, hopefully its okay now. also should i try to add more tests?

kdcokenny: ya since this also changes refresh_dimensions, can you add a test that position fudging is skipped while first_tick is true but works after the first tick?

kdcokenny (approved):

steel-core/src/entity/tests/fall_and_fluids.rs
AnthonyDuong1: i added one using pigs but i think there are other mobs that are better but not implemented like slimes and fishes where they can have a bigger size than their default

Joniii11 (approved):
lgtm

coco875 (dismissed):

steel-core/src/entity/entity/mod.rs
```diff
@@ -2090,7 +2101,7 @@ pub trait Entity: EntityEventSource + ErasedType + Send + Sync + 'static {
 
     /// Returns true if this entity is currently touching lava.
     fn is_in_lava(&self) -> bool {
-        self.fluid_contact().lava_height() > 0.0
+        !self.is_first_tick() && self.fluid_contact().lava_height() > 0.0
```
chocodev11: huh can we put this in a pub so we only need to lock 1 time? or am i deadass

steel-core/src/entity/entity/mod.rs
```diff
@@ -1684,6 +1685,16 @@ pub trait Entity: EntityEventSource + ErasedType + Send + Sync + 'static {
         self.base().tick_count()
     }
 
+    /// Returns whether this entity has not completed its first tick.
+    fn is_first_tick(&self) -> bool {
```
CuzImClicks: pub

steel-core/src/entity/entity/mod.rs
```diff
@@ -1684,6 +1685,16 @@ pub trait Entity: EntityEventSource + ErasedType + Send + Sync + 'static {
...
+        self.base().is_first_tick()
+    }
+
+    /// Sets whether this entity has not completed its first tick.
+    fn set_first_tick(&self, first_tick: bool) {
```
CuzImClicks: pub(crate)

steel-core/src/entity/entity/mod.rs
AnthonyDuong1: ja

CuzImClicks (approved):

coco875 (approved):

PR #565, opened 2026-08-30
Rework recipe

coco875:
## Type of change

- [x] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Rework Recipe the goal is to make recipe easily extendible for both vanilla to include cooking, smelting, etc. And for future Plugin with plugin_type_keeps_custom_data_typed_through_matching_and_key_lookup. To test how will this system work I have implemented/finish to implement, Campfire, Furnace, BlastFurnace, Smoker block.

### Performance
lithium have more performance improvement on them, but didn't want to include it in the same pr, so I have done the minimal who are just cache the latest recipe, I think it cover most use case. There is also the mod FastSuite that optimise it even more but want to make simple system for a first iteration.

## How this was tested

Open the game, try to cook, try to craft. Try signal with comparator. Try exclusive recipe on each kind. Try not compatible one.

## Checklist

- [x] Self-reviewed the diff

## Classes modified:

CampfireBlock, FurnaceBlock, BlastFurnaceBlock, SmokerBlock

## Additional notes

I need to redac some doc but want a first feedback


steel-core/src/inventory/fuel_values.rs
```diff
@@ -0,0 +1,182 @@
...
+        assert!(burn_duration >= 0, "fuel duration cannot be negative");
+        self.burn_durations[item.id()] = burn_duration;
+    }
+}
+
+/// Vanilla 26.2 fuel values, built from the generated item registry and tags.
+pub static VANILLA_FUEL_VALUES: LazyLock<FuelValues> = LazyLock::new(|| {
+    const BASE: i32 = 200;
+
+    FuelValues::builder()
```
CuzImClicks, line 87: all of these values should be extracted or made by a tag

steel-core/src/inventory/menu/kinds/furnace_menu.rs
```diff
@@ -0,0 +1,254 @@
...
+    fn on_tick(
+        &mut self,
+        behavior: &mut MenuBehavior,
+        guard: &mut ContainerLockGuard,
+        _player: &Player,
+    ) {
+        self.update_data(behavior, guard);
+    }
+
+    fn quick_move(
```
CuzImClicks, line 125: could this also be achieved with MenuBuilder::route

steel-core/src/inventory/fuel_values.rs
```diff
@@ -0,0 +1,182 @@
...
+use steel_registry::items::{Item, ItemRef};
+use steel_registry::vanilla_item_tags::ItemTag;
+use steel_registry::vanilla_items;
+use steel_registry::{REGISTRY, RegistryEntry as _, RegistryExt as _, TaggedRegistryExt as _};
+use steel_utils::Identifier;
+
+/// Fuel durations indexed by item registry ID.
+///
+/// The builder is public so a future server registry assembly phase can apply
+/// plugin additions before publishing its process-wide values.
```
CuzImClicks: claude yappery

steel-core/src/block_entity/entities/campfire.rs
```diff
@@ -0,0 +1,293 @@
...
+use crate::player::Player;
+use crate::world::World;
+use crate::world::game_event::GameEventContext;
+
+/// Number of independently cooking item positions on a campfire.
+pub const CAMPFIRE_SLOTS: usize = 4;
+const BURN_COOL_SPEED: i32 = 2;
+const LIT: &BoolProperty = &BlockStateProperties::LIT;
+
+struct CampfireCookingState {
```
CuzImClicks: make this pub and add immutable accessors like .cooking_times() so plugins can read the state

steel-core/src/block_entity/entities/abstract_furnace.rs
```diff
@@ -0,0 +1,581 @@
...
+    const fn fuel_duration(self, vanilla_duration: i32) -> i32 {
+        match self {
+            Self::Furnace => vanilla_duration,
+            Self::BlastFurnace | Self::Smoker => vanilla_duration / 2,
+        }
+    }
+}
+
+/// Independently lockable furnace inventory and progress data.
+pub(crate) struct FurnaceContainer {
```
CuzImClicks: why pub crate?

steel-core/src/block_entity/entities/abstract_furnace.rs
```diff
@@ -0,0 +1,581 @@
...
+impl AbstractFurnaceBlockEntity {
+    fn new(
+        block_entity_type: BlockEntityTypeRef,
+        kind: FurnaceKind,
+        level: Weak<World>,
+        pos: BlockPos,
+        state: BlockStateId,
+    ) -> Self {
+        let base = Arc::new(BlockEntityBase::new(block_entity_type, level, pos, state));
+        let container = Arc::new(SyncMutex::new(FurnaceContainer::new(kind)));
```
CuzImClicks: you can use .into_shared()

CuzImClicks (changes requested):
i just did a small read. thanks for finally addressing recipes! :D

steel-core/src/block_entity/entities/abstract_furnace.rs
coco875: thanks

steel-core/src/block_entity/entities/campfire.rs
coco875: a bit to much projection but done

steel-core/src/inventory/fuel_values.rs
coco875: remove

steel-core/src/block_entity/entities/abstract_furnace.rs
coco875: I have simplify to be pub (without restriction)

steel-core/src/block_entity/entities/abstract_furnace.rs
coco875: I also see that it's missing in barrel an potentially other container

steel-core/src/inventory/fuel_values.rs
coco875: I didn't take attention to it and I have double check and it's hardcoded... I will update the extractor then (will still check 26.3 just in case it get datapack driven) https://mcsrc.dev/2/26.2/net/minecraft/world/level/block/entity/FuelValues#L43

coco875: image
26.3 change it... Can I just put a todo, so like that we can fix it when we update to 26.3 with the new system ?

steel-core/src/behavior/items/shovel.rs
Laptop59: A bit out of scope but maybe implement this?

steel-core/src/block_entity/entities/abstract_furnace.rs
Laptop59: I can't find any code for awarding stats here, or triggering advancement criterion triggers. Should probably leave a TODO for those

steel-core/src/inventory/menu/kinds/furnace_menu.rs
coco875: the problem is that vanilla have the behavior of allowing placing a non valid can_smelt item in the first slot, so with route from what I understand the move are the same filter as can_place in it, so it make a difference from vanilla behavior

Image

steel-core/src/entity/entities/mobs/passive/sheep/mod.rs
Laptop59, line 333: I know you didn't make these functions, but it looks like some other mobs like cats and wolves also use these color functions. Maybe move them to `DyeColor`.

steel-core/src/inventory/slots/furnace_slots.rs
```diff
@@ -0,0 +1,148 @@
...
+            .get_typed_mut::<FurnaceContainer>(container_id)
+            .map(FurnaceContainer::take_recipes_used);
+        self.base.set_changed(guard);
+        if let Some(recipes) = recipes {
+            guard.run_unlocked(|| {
+                pop_furnace_experience(&player.get_world(), player.position(), recipes);
+            });
+        }
+        None
+    }
```
Laptop59, line 148: I can't find any code for awarding stats here, unlocking recipes, or triggering advancement criterion triggers. Should probably leave a TODO for those.

steel-core/src/block_entity/registry.rs
```diff
@@ -15,13 +15,7 @@ use steel_registry::{REGISTRY, RegistryEntry, RegistryExt};
...
 
 use super::SharedBlockEntity;
-use super::entities::{
-    BarrelBlockEntity, BeehiveBlockEntity, BlastFurnaceBlockEntity, BrushableBlockEntity,
-    CampfireBlockEntity, ChiseledBookShelfBlockEntity, ComparatorBlockEntity,
-    DaylightDetectorBlockEntity, EndGatewayBlockEntity, EndPortalBlockEntity, FurnaceBlockEntity,
-    JukeboxBlockEntity, PistonMovingBlockEntity, PotentSulfurBlockEntity, RawBlockEntity,
-    SignBlockEntity, SmokerBlockEntity,
-};
+use super::entities::*;
```
coco875, line 25: a bit anoying so I have simplify it

steel-registry/src/data_components/registry/patch_persistence.rs
```diff
@@ -1,11 +1,26 @@
...
+    ) -> i32 {
+        let entry = hash_entry(key.to_string().compute_hash(), value.compute_hash());
+        let mut hasher = ComponentHasher::new();
+        hasher.start_map();
+        hasher.put_raw_bytes(&entry.key_bytes);
+        hasher.put_raw_bytes(&entry.value_bytes);
+        hasher.end_map();
+        hasher.finish()
+    }
+
```
Laptop59, line 23: This solution feels a bit hacky for me, because this hash probably won't match with a real hash of the same template (but not in the build step). Would a OnceLock be better for hashing purposes in `ItemStackTemplate`?

steel-core/src/behavior/items/shovel.rs
```diff
@@ -87,7 +87,6 @@ impl ItemBehavior for ShovelItem {
                 0,
                 None,
             );
-            // TODO: CampfireBlock::dowse() — eject cooking items
```
Laptop59, line 90: Why did you remove the TODO? It looks like CampfireBlock::dowse also dispatches a game event in addition to spawning particles if client-side.

steel-core/src/behavior/items/shovel.rs
Laptop59, line 96: A bit out of scope but maybe implement this?

steel-registry/src/data_components/registry/patch_persistence.rs
coco875: the function below (compute_persistent_hash) also do the same 
```rust
let mut hasher = ComponentHasher::new();
        hasher.start_map();
        for entry in &entries {
            hasher.put_raw_bytes(&entry.key_bytes);
            hasher.put_raw_bytes(&entry.value_bytes);
        }
        hasher.end_map();
```
so it's just a mimic of it. Not really dive into it to know how hacky it is

steel-core/src/behavior/items/shovel.rs
coco875: particle are client side, so only game event mater, I will add it then ? or at least change the todo ?

Laptop59: > particle are client side, so only game event mater, I will add it then ? or at least change the todo ?

Yeah, just implement the `dowse` method which will call the game event.

steel-core/src/inventory/slots/furnace_slots.rs
coco875: done

steel-core/src/behavior/items/shovel.rs
coco875: there is alreadt the event here https://github.com/Steel-Foundation/SteelMC/pull/565/changes/BASE..fb565e34757727aaba96ce6c5d3b8e15c68c18c0#diff-c79ab0767bb8d53763626db13fdbf1a8e7ac437d8853e105fc924bfec6dba1fdR97-R101 no ?

Laptop59: Yes, there is already an event (I did not see that), but I would still prefer putting that in the `dowse` function and removing that outside

Laptop59: I guess it doesn't really matter that much.

coco875: I can do that but don't see the advantage

image

steel-core/src/behavior/items/shovel.rs
coco875: prefer let it to somone else and avoid increase the pr size (who are already at 5k)

steel-core/src/behavior/items/shovel.rs
Laptop59: > I can do that but don't see the advantage
> image

I just noticed that this `dowse` function is also used in `AbstractThrownPotion` and in the `CampfireBlock` itself, so it might be worth it

steel-core/src/entity/entities/mobs/passive/sheep/mod.rs
coco875: hmm if it re-use later I am tempted to do it, but I also don't want to increase even more change, maybe I can just add a todo ?

Laptop59: > hmm if it re-use later I am tempted to do it, but I also don't want to increase even more change, maybe I can just add a todo ?

That's fine.

steel-registry/src/recipe/crafting/matching.rs
```diff
@@ -0,0 +1,405 @@
...
+                &recipe.map
+            } else {
+                &recipe.material
+            };
+            if !ingredient.test(input.get(x, y)) {
+                return false;
+            }
+        }
+    }
+    let Some(data) = input.map_data(4) else {
```
kdcokenny: where does map_data get filled for actual crafting? From what I can see, CraftingInput::new sets every entry to None, and the live path through CraftingContainer::as_positioned_input never calls with_map_data. I couldnt find any other caller either, so this recipe looks like it can never match even for a valid scalable map. vanilla reads the saved map data from the level here, so I think that data needs to be wired into the live recipe input before this can work..

kdcokenny (dismissed):

steel-core/src/inventory/fuel_values.rs
CuzImClicks: yes

steel-core/src/behavior/blocks/building/campfire_block.rs
```diff
@@ -94,6 +108,46 @@ impl BlockBehavior for CampfireBlock {
...
+                    &vanilla_recipe_types::CAMPFIRE_COOKING,
+                    &SingleItemRecipeInput::new(stack.clone()),
+                )
+                .is_some()
+        });
+        if !accepted {
+            return InteractionResult::TryEmptyHandInteraction;
+        }
+
+        let Some(block_entity) = world.get_block_entity(pos) else {
```
CuzImClicks, line 134:
```java
if (level.getBlockEntity(pos) instanceof CampfireBlockEntity campfire) {
    ItemStack itemInHand = player.getItemInHand(hand);
    if (level.recipeAccess().propertySet(RecipePropertySet.CAMPFIRE_INPUT).test(itemInHand)) { ... }
}
return InteractionResult.TRY_WITH_EMPTY_HAND;
```

try with empty hand here instead of consume

steel-registry/src/recipe/crafting/matching.rs
```diff
@@ -0,0 +1,541 @@
...
+        }
+    }
+    has_target && has_dye
+}
+
+/// Matches the fixed 3×3 map-scaling layout and its saved-map restrictions.
+///
+/// The source map's [`super::MapRecipeData`] must be attached to the center
+/// input slot so exploration maps and maps at the maximum scale are rejected.
+fn map_extending(recipe: &MapExtendingRecipe, input: &CraftingInput) -> bool {
```
CuzImClicks, line 358: add small todo for the pr that implements maps to finish implementing this

CuzImClicks (dismissed):

CuzImClicks (approved):

teknostom (approved):
LGTM

PR #570, opened 2026-08-31
Add default_methods macro

4lve:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

### Problem
In Rust when we override a method there is no way to call that original code anymore. Lets say you have a `hurt()` default entity implementation. But now you want to make this new entity that is immune to projectile damage. In Java you would just override the method, return if projectile and otherwise just call super.hurt() to apply the normal damage. But in Rust this breaks down.

If you were to call self.hurt() or Entity::hurt(self) you would end up calling your own overridden method again, which quickly becomes an infinite loop that would stall the server.

### Previous Approach
There are a few ways to avoid this recursion, but they don't look that clean:

- Copy the default implementation into the override. This duplicates logic and the implementations may drift apart if the default one gets changed.
- Extract the default implementation into a separate helper such as `default_hurt`. Steel has used this approach until now, but it adds an extra public method to the main trait and could get as messy if someone accidentally overrides the `default_hurt`

### Solution
This PR introduces the `#[default_methods]` macro. Methods marked with `#[default_method]` have their original implementations exposed through a generated trait named
`<Trait>Defaults`.

For example (Dummy code, won't work in steel):
```rust
#[default_methods]
trait Entity {
  #[default_method]
  fn hurt(&self, source: &DamageSource, amount: f32) -> bool {
    true
  }
}

impl Entity for ProjectileImmuneEntity {
  fn hurt(&self, source: &DamageSource, amount: f32) -> bool {
    if source.is_projectile() {
        return false;
    }
    
    EntityDefaults::hurt(self, source, amount)
  }
}
```

## How this was tested
I didn't have time to test this in game but behavior should remain unchanged

## Checklist

- [x] Self-reviewed the diff


steel-macros/src/default_methods.rs
```diff
@@ -0,0 +1,353 @@
...
+        if reference.lifetime.is_none() {
+            reference.lifetime = Some(self.0.clone());
+        }
+        visit_mut::visit_type_reference_mut(self, reference);
+    }
+
+    fn visit_type_fn_ptr_mut(&mut self, _function: &mut syn::TypeFnPtr) {}
+}
+
+fn replace_self(tokens: TokenStream, replacement: &Ident) -> TokenStream {
```
kdcokenny, line 229: this rewrites every token named self, including tokens inside macros and nested impls. stringify!(self) returns __default_methods_receiver, and a nested fn value(&self) doesnt compile anymore.
could we make this syntax aware so it only rewrites uses of this methods receiver?

steel-macros/src/default_methods.rs
```diff
@@ -0,0 +1,353 @@
...
+
+impl VisitMut for ElidedOutputLifetimes<'_> {
+    fn visit_type_reference_mut(&mut self, reference: &mut TypeReference) {
+        if reference.lifetime.is_none() {
+            reference.lifetime = Some(self.0.clone());
+        }
+        visit_mut::visit_type_reference_mut(self, reference);
+    }
+
+    fn visit_type_fn_ptr_mut(&mut self, _function: &mut syn::TypeFnPtr) {}
```
kdcokenny, line 226: the &strs inside impl Fn(&str) -> &str have their own lifetime. this visitor ties both to &self, so a valid default method stops compiling with Fn is not general enough.
could we leave references inside Fn, FnMut, and FnOnce alone like function pointers?

steel-macros/src/default_methods.rs
```diff
@@ -0,0 +1,353 @@
...
+use syn::{
+    Attribute, FnArg, GenericParam, Ident, ItemTrait, Lifetime, Meta, Pat, ReceiverKind,
+    ReturnType, TraitItem, TraitItemFn, Type, TypeReference, parse_quote,
+};
+
+pub fn expand(original: &mut ItemTrait) -> syn::Result<TokenStream> {
+    let trait_name = &original.ident;
+    let defaults_name = format_ident!("{}Defaults", trait_name);
+    let visibility = &original.vis;
+    let conditional_attributes: Vec<_> = original
```
kdcokenny, line 14: as far as I can tell, this only copies cfg attrs, so lint attrs on the source trait dont apply to the generated trait. for example, #[allow(missing_docs)] still makes the generated method fail under deny(missing_docs).

steel-macros/src/default_methods.rs
```diff
@@ -0,0 +1,353 @@
...
+        }
+    }
+
+    method.sig.inputs[receiver_index] =
+        parse_quote!(#(#attributes)* #binding_mutability #receiver_name: #receiver_type);
+
+    let body = method
+        .default
+        .take()
+        .expect("only provided methods are copied");
```
kdcokenny, line 159: just double checking that this is intentional bc codex brought up that this [violates agents.md](https://github.com/Steel-Foundation/SteelMC/blob/master/AGENTS.md#:~:text=.clone()%20to%20appease%20borrow%20checker%2C%20.unwrap()/.expect()%20in%20production%2C%20unsafe%20other%20than%20the%20keyed%20DowncastType%20exception%20below)

kdcokenny (changes requested):

steel-core/src/entity/mob/tests.rs
```diff
@@ -790,16 +790,16 @@ fn looting_collects_nearby_item_into_main_hand() {
...
+    // loop against regressing behind the effective-AI gate.
     init_vanilla_registry();
     init_behaviors();
     let world = fresh_test_world("mob_looting_no_ai");
     insert_ready_full_chunk(&world, ChunkPos::new(0, 0));
 
-    let mob = Arc::new(PigEntity::new(
-        &vanilla_entities::PIG,
+    let mob = Arc::new(EndermiteEntity::new(
+        &vanilla_entities::ENDERMITE,
```
coco875, line 802: why changing this ?

coco875 (changes requested):

PR #577, opened 2026-08-31
fix: Match Vanilla chunk ticket timing and ownership

### LINKED ISSUE #566: Player simulation tickets can stay active after changing worlds
`ChunkMap::remove_player` queues the player ticket removal through the async scheduler. if an epoch is already running, the old world can keep entity ticking until a later commit.

vanilla tracks `PLAYER_LOADING` and `PLAYER_SIMULATION` separately, then applies simulation updates before entities tick.

i found this while testing #255. a fishing rod pearl stasis chamber around x 8000 triggers by itself after entering the nether because the overworld bobber gets another tick and despawns. the same setup can look fine near 0 0, which hides the issue.

when the last player leaves, their simulation contribution should be gone before the next entity tick. other valid simulation tickets should still keep the chunk ticking.

kdcokenny:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactor / code cleanup
- [x] Performance improvement
- [ ] Chore / tooling

## Description

Steel ran loading and simulation tickets through the same async scheduler. That could leave a player's simulation source active in the old world for another entity tick.

Ticket updates now run together in Vanilla's chunk source phase, after scheduled ticks and before random chunk and entity gameplay. Generation and save work stay async.

Tickets use registry backed Vanilla `TicketType` values and one canonical store. Player loading and simulation sources have separate ownership, timed behavior comes from each type, and equal chunk requests use refcounted leases so one handle cannot release another.

Loading and simulation share an incremental eight-neighbor tracker with separate level limits. Player movement updates the affected region without rebuilding propagation from every loading source. Gameplay lookup cache scopes end before holder lifecycle changes and restart for later phases, so cached misses cannot hide revived chunks.

## How this was tested

Regression coverage includes shared and moving player sources, receipt ordering, timed expiration around scheduled ticks, persisted portal restore, unload and revival races, generation save pins, and cached misses across holder revival. Deterministic source-operation sequences compare both propagation domains against a full recomputation oracle.

- 241 focused chunk tests passed.
- All 2,529 enabled `steel-core` unit tests passed; one existing test is ignored.
- `cargo fmt --all --check`, `typos`, and `cargo clippy -r --all-targets --all-features` passed.

The isolated loading benchmark moved one player among 50 separated players at view distance 10. Median propagation time fell from approximately 142 ms to 2.4 ms. The tracker modules were optimized and dependencies used development builds; these are comparative microbenchmark results, not release-server tick timings.

## Screenshots / logs

N/A

## Checklist

- [x] Self-reviewed the diff

## Classes modified:

N/A

## Additional notes

`PLAYER_LOADING` reaches the same final coverage as Vanilla. Vanilla's separate 4 slot admission dispatcher is tracked in #586 instead of adding more scheduling code here.

Closes #566


steel-core/src/chunk/chunk_ticket_storage.rs
```diff
@@ -0,0 +1,670 @@
...
+    tickets: Vec<PersistentChunkTicket>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+struct PersistentChunkTicket {
+    #[serde(rename = "type")]
+    ticket_type: Identifier,
+    chunk_x: i32,
+    chunk_z: i32,
+    level: u8,
```
darkstarworks, line 33: `level` is a required field, but worlds saved before this PR have no `level` key in `chunk_tickets.toml`. `load_or_default` only falls back to `Default` when the file is *absent*, so an existing save parses as `missing field 'level'` and becomes `InvalidData`, failing the whole world load I think

the fix would be to give `level` a `#[serde(default = ...)]` returning the portal ticket level.

steel-core/src/chunk/player_ticket_tracker.rs
```diff
@@ -0,0 +1,399 @@
...
+        );
+        assert!(
+            self.player_positions.insert(player_id, pos).is_none(),
+            "player position index retained a moved player"
+        );
+
+        if was_empty {
+            changes
+                .merge(storage.add_ticket(pos, player_simulation_ticket(self.simulation_distance)));
+            Self::for_each_square_position(pos, self.view_distance, |covered_pos| {
```
darkstarworks, line 90: a player crossing one chunk boundary goes through `add_player`, calling `remove_player_from_center` for the old centre and then increments the new one. Both walk a full `for_each_square_position`, so at view distance 10 that is 441 decrements + 441 increments per crossing, even though the two squares overlap by about 90% and the real delta is roughly 84 chunks.

because of the performance nature of the PR I think treating a move as a set difference instead of a remove-then-add is more optimised?

> `for_each_square_ring` at line 254 already establishes the "only touch the delta" pattern

Maybe falling back to the current full-square path when the centres are more than `2 * view_distance` apart (no overlap) keeps it simple (also [:194](https://github.com/Steel-Foundation/SteelMC/pull/577/changes#diff-b5277b7bb2f63526b9c8ec5217705052ba42f48f894acaef48a3c9a2de2b1836R194))

steel-core/src/chunk/chunk_map/mod.rs
```diff
@@ -605,21 +606,23 @@ impl ChunkMap {
     where
         F: FnOnce() -> R,
     {
-        let ticket = ChunkTicket::full_chunks(radius);
-
-        let ticket_revision = self.add_chunk_ticket(center, ticket);
+        let ticket_level = ChunkTicketLevel::for_full_chunk_radius(radius);
+        let Some(ticket_receipt) = self.acquire_chunk_request_leases(&[center], ticket_level)
```
darkstarworks: acquired here and released only by the explicit calls on lines 625 and 634. Two paths reach neither(?) the future being dropped at `sleep(...).await` on line 630 (task abort, instead of the `cancel_token` path that is handled), and `f()` panicking on line 633.
Because `CHUNK_REQUEST` is `NO_TIMEOUT`, a leaked lease pins that chunk and its radius for the lifetime of the world with no recovery path. kinda low risk, but permanent

I think holding `ChunkRequestHandle` from `request_chunks` rather than calling the lease API directly (like `player_spawn_finder` already uses) allows for `Drop` to do it freely

steel-core/src/chunk/chunk_ticket_storage.rs
kdcokenny: yeah you're right, old saves with ticket entries would fail to load here but we are in alpha and breaking changes are expected, esp for correctness. also, defaulting it to the portal level would silently accept missing levels for other ticket types

steel-core/src/chunk/player_ticket_tracker.rs
kdcokenny: just to clarify, this isn't a performance pr. it's to make raiden's fishing rod ender stasis chamber functional by aligning us with vanilla. I can make this optimization but I think it belongs in a followup

steel-core/src/chunk/chunk_ticket_storage.rs
```diff
@@ -0,0 +1,670 @@
...
+}
+
+/// A recoverable invalid value in persisted chunk ticket data.
+#[derive(Debug, Error, PartialEq, Eq)]
+pub(crate) enum ChunkTicketStorageLoadError {
+    #[error("unknown chunk ticket type `{0}`")]
+    UnknownTicketType(Identifier),
+    #[error("invalid chunk ticket level {level} for type `{ticket_type}`")]
+    InvalidTicketLevel { ticket_type: Identifier, level: u8 },
+}
```
darkstarworks: `from_persistent` returns `Err` on the first bad entry, and `World::new` converts it into an `io::Error` that fails the world load, so a single bad ticket is fatal rather than recoverable.

steel-core/src/chunk/chunk_map/mod.rs
kdcokenny: good catch! fixed

steel-core/src/chunk/chunk_ticket_storage.rs
kdcokenny: so for this vanilla logs the decoding error and uses partial data when available. i'll check its ticket list handling so we match that properly

kdcokenny: i'm thinking that this belongs in a separate pr considering that this pr wasn't meant to fix it and this issue predates this pr

coco875 (approved):

CuzImClicks (approved):

PR #578, opened 2026-09-01
fix: Create fresh players for death and End respawns

### LINKED ISSUE #573: Respawn reuses the same Player instance instead of creating a new one
found this while checking what #571 needs for vanilla parity.

vanilla creates a fresh `ServerPlayer` for death respawn and the End credits respawn. `PlayerList.respawn` moves the existing connection to it, copies state through `ServerPlayer.restoreFrom`, preserves the numeric entity id, replaces the old player in the server and world indexes, then returns the new player.

normal dimension changes are different. vanilla keeps the same `ServerPlayer` for those.

Steel currently resets the same `Arc<Player>` in `player/lifecycle/respawn.rs`. `player/lifecycle/spawn_sync.rs` already documents this difference.

this becomes observable once `DamageSource` keeps exact entity refs. a source created before death would start pointing at the reset post respawn player. vanilla keeps pointing at the retired player from before the respawn.

this blocks #571.

## desired behavior

- death respawns allocate a fresh `Player`
- the End credits respawn allocates a fresh `Player`
- normal dimension changes keep the same `Player`
- the new player keeps the same numeric entity id where vanilla does, but has a new `EntityInstanceId`
- the existing connection starts routing packets to the new player before later input is processed
- server and world player indexes replace the old player with the new one
- queued packets or jobs tied to the old incarnation cannot mutate the replacement
- state transfer matches every field and branch in the target vanilla `ServerPlayer.restoreFrom`
- the old player can stay alive while an exact historical ref exists, without becoming active again or creating an `Arc` cycle

the connection probably needs a stable session owner that can point at the current `Arc<Player>`, but that should land as part of this complete respawn path rather than as a separate unused foundation.

## verification

tests should prove that:

- death and End credits respawns produce a different allocation and instance id
- the numeric entity id still matches vanilla
- a normal dimension change keeps the same allocation
- the connection and every player index point only at the replacement after respawn
- the old player remains retired even if another owner still holds its `Arc`
- both `restoreFrom` branches copy and reset the same state as vanilla

kdcokenny:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Fixes #573.

Death and End credits respawns now create a fresh `Player`, like vanilla's `PlayerList.respawn`. The replacement keeps the same profile UUID, numeric entity ID, connection, packet order, chat state, and chunk sender. It gets a new `EntityGeneration`.

The respawn swap updates the world and server indexes together, restores every player field Steel currently has, marks the old player removed, and routes queued input and pending restore work to the new player. Normal dimension changes still reuse the same `Player`.

## How this was tested

Added focused tests for both respawn paths, player replacement, state restore, packet ordering, pending restores, and cross world ender pearls.

## Screenshots / logs

N/A

## Checklist

- [x] Self-reviewed the diff

## Classes modified:

- `ServerPlayer`
- `PlayerList`
- `ServerGamePacketListenerImpl`
- `PlayerChunkSender`
- `ThrownEnderpearl`
- `AttributeMap`

## Additional notes

Steel doesn't have every field vanilla copies in `ServerPlayer.restoreFrom` yet. This copies every field Steel currently has without adding fake placeholders.


steel-core/src/server/packet_processor.rs
```diff
@@ -966,7 +1207,7 @@ mod tests {
         }
 
         let state = queue.state.lock();
-        let lane = state.lanes.get(&key).expect("session lane should exist");
+        let lane = state.lanes.get(&1).expect("session lane should exist");
```
coco875, line 1210:
```suggestion
        let lane = state.lanes.get(1).expect("session lane should exist");
```

coco875 (dismissed):
I don't have seen problem but there is only the "duplicate"/very close functionality between uuid and entity generation that I don't like

steel-core/src/server/packet_processor.rs
kdcokenny: lanes is an FxHashMap, so the & in .get(&1) is needed here.

coco875 (approved):

CuzImClicks (approved):

PR #581, opened 2026-09-01
fix: drop invalid --fix arg from prek end-of-file-fixer hook

darkstarworks:
## Type of change

- [x] Chore / tooling

## Description

The prek `end-of-file-fixer` hook is configured with `args = ["--fix"]`, but that hook takes no arguments. On current prek (0.5.0) the unknown argument is rejected, which aborts the whole pre-commit run before any hook executes, so no local commit can go through:

```
error: Failed to run hook `end-of-file-fixer`
  caused by: error: unexpected argument '--fix' found
```

CI does not run prek, so this only affects contributors running the local pre-commit suite documented in `AGENTS.md`. `end-of-file-fixer` fixes files by default, so dropping the argument leaves behaviour unchanged. (`--fix` is a valid argument for `mixed-line-ending`, which is likely where it was copied from.)

## How this was tested

`prek run end-of-file-fixer --all-files` fails before the change with the error above and passes after it. Confirmed against the prek builtin-hooks docs, which list `end-of-file-fixer` supported arguments as "None".

## Screenshots / logs

Not applicable, a tooling config change.

## Checklist

- [x] Self-reviewed the diff


Joniii11 (approved):
looks good lol

JunkyDeveloper (approved):

CuzImClicks (approved):

PR #583, opened 2026-09-01
feat: add game_event_with_source_entity

AnthonyDuong1:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [X] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description
This PR adds game_event_with_source_entity so that game_event can take in any entity or None as the source. Previously, we only had 2 functions, one that uses the entity that called game_event as the source entity and another that takes in a player for the source entity. Entities/items like item frames, armor stands, minecarts, etc emits a game event when broken. They can be broken by mobs like creepers exploding and skeletons shooting it. This means those mobs are the source, but currently we don't have a way to set them as a game event source.

## How this was tested
I ran cargo fmt, cargo test, and cargo clippy and there were no warnings or fails.

## Checklist

- [X] Self-reviewed the diff

## Additional notes
I removed game_event_with_player since we can just use Some(player) with game_event_with_source_entity as the equivalent.


Joniii11 (approved):
lgtm

CuzImClicks (changes requested):
please fix these call sites aswell

steel-core/src/entity/living_entity.rs:2605

steel-core/src/entity/entity/mod.rs:3015

steel-core/src/entity/entity/mod.rs:3076

steel-core/src/entity/projectile/mod.rs:614

steel-core/src/entity/entities/mobs/passive/sheep/mod.rs:565

CuzImClicks (approved):

JunkyDeveloper (approved):

PR #443, opened 2026-08-22
feat: /fill command

GreenedDev:
BIG CREDITS TO 4LVE. the implementation is his: https://github.com/Steel-Foundation/SteelMC/commit/24068bd384eed1afb3a67ff21d8a80e03a4ff894
i just ported the unmerged commit to new steel and resolved conflicts.

## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [x] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Documentation
- [ ] Chore / tooling

## Description
implements fill command

## How this was tested
in-game

## Screenshots / logs
image

## Checklist

- [x] Code builds w/o errors or warnings
- [x] Self-reviewed the diff
- [ ] Docs updated (if applicable)
- [x] No leftover debug code / comments

PR #585, opened 2026-09-01
fix: /setblock failing after block_state parser change

Purdze:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description
#443 changed `BlockStateParser` to produce `BlockInput`, but `/setblock` still read the argument as `BlockPredicate`, so every invocation failed. Now uses `block_input()` + `BlockInput::place` like `/fill`, requires a loaded position like vanilla, shares `placement_flags` with `/fill`, and removes the unused `update_neighbour_on_block_set`. Includes an end-to-end regression test.

## How this was tested

new regression test runs the command through the real dispatcher and asserts
the resulting block states, e.g.:
`setblock 8 64 8 minecraft:stone` places stone
`setblock 8 64 8 minecraft:dirt keep` fails, stone kept
`setblock 8 65 8 minecraft:end_rod[facing=east]` resolves to vanilla state id 14637

## Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/steel-core/src/command/builtins/setblock.rs
+++ b/steel-core/src/command/builtins/setblock.rs
@@ -9,7 +9,6 @@ use super::super::{
     registration::CommandRegistration,
 };
 use super::execute::loaded_block_position;
-use crate::world::tick_scheduler::TickPriority;
 use steel_registry::blocks::block_state_ext::BlockStateExt;
 use steel_utils::Identifier;
 use steel_utils::translations::{COMMANDS_SETBLOCK_FAILED, COMMANDS_SETBLOCK_SUCCESS};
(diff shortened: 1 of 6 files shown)


coco875 (dismissed):

steel-core/src/command/builtins/setblock.rs
```diff
@@ -53,95 +49,36 @@ fn set_block(
...
+        level.update_neighbors_on_block_set(block_pos, old_state);
+        let placed_state = level.get_block_state(block_pos);
+        if placed_state.has_fluid() {
             level.schedule_fluid_tick(
                 block_pos,
-                block_state.get_fluid_state().fluid_id,
+                placed_state.get_fluid_state().fluid_id,
                 0,
                 TickPriority::ExtremelyHigh,
             );
```
CuzImClicks: delete this, since .place already calls on_place for liquid_block down the line which schedules a fluid tick already

CuzImClicks (changes requested):

steel-core/src/command/builtins/setblock.rs
Purdze: yep, done this already just hadnt committed it yet

CuzImClicks (approved):

coco875 (approved):

PR #587, opened 2026-09-01
ItemStack::consume

CuzImClicks:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

ItemStack::consume is a vanilla function that normally ensures that consumable goods are skipped when having infinite resources.

steel-registry cannot depend on steel-core so our consume function only takes bool, but this pushes users that want to implement something that consumes into handling the case of infinite resources. 

the code output is mostly identical because it just expands to what most call sites used before

## How this was tested

threw some enderpearls and cargo test and cargo clippy

## Checklist

- [x] Self-reviewed the diff


steel-registry/src/item_stack.rs
```diff
@@ -133,6 +133,26 @@ impl ItemStack {
         self.count -= amount;
     }
 
+    /// Decreases the count by the given amount, unless the user has infinite
+    /// materials.
+    ///
+    /// See: `Player::has_infinite_materials`
+    #[inline]
+    pub const fn consume(&mut self, amount: i32, has_infinite_materials: bool) {
```
coco875, line 141: there is a lot of consume(1, ...) I think it would make the code a bit more clean if we add a function that consume 1 by default like consume_single(...)

coco875 (changes requested):
only one suggestion

steel-registry/src/item_stack.rs
chocodev11: i feel like it is overkill

Laptop59: I know this is opinionated, but maybe `consume_one` would be a better name and more understandable? It's a good suggestion though (but yeah 1 is pretty self-explanatory)

JunkyDeveloper (dismissed):

JunkyDeveloper (approved):

coco875 (approved):

PR #607, opened 2026-09-05
add handle_first_tick_bubble_column

AnthonyDuong1:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [X] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description
This PR adds handle_first_tick_bubble_column in ThrowableProjectile. This allows throwable projectiles to have the correct initial velocity on the first tick.

## How this was tested
I ran cargo fmt, cargo test, and cargo clippy with no errors. I created a test that tests the position of the snowball on the first tick and on the second tick. The first tick should apply the bubble column effect at the very beginning which gives the snowball a positive initial velocity, causing the position to increase vertically. The second tick shouldn't apply the bubble column effect at the beginning, and on a snowball with 0 velocity, it'll have a negative vertical velocity, causing it to move down.

## Checklist

- [X] Self-reviewed the diff


steel-core/src/entity/projectile/throwable.rs
```diff
@@ -71,4 +74,114 @@ pub trait ThrowableProjectile: Projectile {
             self.hit_target_or_deflect_self(&result);
         }
     }
+
+    /// Vanilla `ThrowableProjectile.handleFirstTickBubbleColumn`.
```
CuzImClicks: useless comment since this function is literally ThrowableProjectile::handle_first_tick_bubble_column

CuzImClicks (dismissed):
looks good but just remove the comment

steel-core/src/entity/projectile/throwable.rs
AnthonyDuong1: removed ty

CuzImClicks (dismissed):

steel-core/src/entity/projectile/throwable.rs
```diff
@@ -71,4 +74,114 @@ pub trait ThrowableProjectile: Projectile {
...
+
+        let mut ignored_effects = InsideBlockEffectCollector::new();
+
+        for pos in BlockPos::between_closed(min, max) {
+            let state = world.get_block_state(pos);
+            if state.get_block() != &vanilla_blocks::BUBBLE_COLUMN {
+                continue;
+            }
+
+            BLOCK_BEHAVIORS
```
teknostom, line 108: This falls through to the entity handler. It should be the projectile trait, projectile does not clamp, entity does. (from what i understand)

teknostom (changes requested):

steel-core/src/entity/projectile/throwable.rs
AnthonyDuong1: yea just checked and projectile does override and directly adds the values without clamping. i'll fix that and add a test for it

AnthonyDuong1: also saw ender pearl specifically calls Entity's handler, so i added that and made a test for it

CuzImClicks (approved):
i added particles for you

teknostom (approved):
LGTM

PR #609, opened 2026-09-05
fix: Match vanilla behavior on item use and release

### LINKED ISSUE #529: Active item use can overwrite changes to the held stack
found this while looking through #411, but the bug was already there before that PR.

`tick_active_item_use` only checks if the hand still has the same item type. it then runs `on_use_tick` with its local stack and writes that stack back into the hand no matter what happened during the callback.

if `on_use_tick` changes the held stack, Steel can finish using the old stack and overwrite the new one. vanilla uses the held stack while ticking and checks the full stack again before `finishUsingItem`.

we should match that without bringing back the empty hand bug that #411 fixes. a test where `on_use_tick` replaces the held stack should cover it.

NandoBF:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Fixes #529 by better mimicking vanilla behaviour on functions `tick_active_item_use` and `release_using_item`.

Previously the held item would be replaced with an empty stack and the `on_use_tick` would be run on the local item stack. Now, the held item remains unchanged and `on_use_tick` is run on a clone of the held item. After that it is checked if the item still matches the item that was being held, if it does overwrite to apply item changes, if it doesn't it returns early. This check is also done after `finish_using_item`.

Also changed `release_using_item` to use the same pattern so that it matches vanilla since it had the same bug.

## How this was tested

Ran cargo fmt, cargo clippy, cargo check and typos. There aren't any tests that cover this bug and I also didn't make one since the necessary behaviour to replicate this is in PR #411, which isn't merged.

## Checklist

- [x] Self-reviewed the diff

## Additional notes

Although the bug was present in `release_using_item` I don't think any vanilla item would be able to trigger it, since the only item that changes to something different than an empty hand on release would be an elytra (which isn't used on hand). However, plugins/mods could trigger that bug.


steel-core/src/player/mod.rs
```diff
@@ -399,7 +399,15 @@ impl Player {
             self,
             active.remaining_ticks(),
         );
-        self.inventory.lock().set_item_in_hand(hand, item);
+
+        {
+            let mut inventory = self.inventory.lock();
+            let current_hand = inventory.get_item_in_hand_mut(hand);
+            if current_hand.item() == active.item() {
+                *current_hand = item;
```
kdcokenny, line 407: same thing here in `release_using_item`, if `release_using` changed the stack in the hand, checking only `current_hand.item() == active.item()` will overwrite those changes.

steel-core/src/player/mod.rs
```diff
@@ -419,28 +427,45 @@ impl Player {
...
+            let inventory = self.inventory.lock();
+            inventory.get_item_in_hand(hand).clone()
         };
         let world = self.get_world();
         let behavior = ITEM_BEHAVIORS.get_behavior(item.item());
         behavior.on_use_tick(&world, self, &mut item, active.remaining_ticks());
 
         if self.active_item_use_hand() != Some(hand) {
-            self.inventory.lock().set_item_in_hand(hand, item);
             return;
```
kdcokenny: what happens here if `on_use_tick` damages the item (like a brush) and stops item use? returning early here drops those changes without writing them back to the inventory.

steel-core/src/player/mod.rs
```diff
@@ -419,28 +427,45 @@ impl Player {
...
         }
+
+        {
+            let mut inventory = self.inventory.lock();
+            let current_hand = inventory.get_item_in_hand_mut(hand);
+            if current_hand.item() == active.item() {
+                *current_hand = item.clone();
```
kdcokenny: this writes `item` back as long as the item id matches, which overwrites any changes made to the held stack during `on_use_tick`.

and because it writes `item` back here, `hand_item` down on line 455 is always going to equal `item`. so `if item == hand_item` will basically always be true and never hit the `release_using_item` branch.

vanilla checks if the hand still matches the stack from before `onUseTick`. if it changed during the tick, it releases instead of finishing and doesnt overwrite the hand.

steel-core/src/player/mod.rs
```diff
@@ -419,28 +427,45 @@ impl Player {
...
         };
         if active.remaining_ticks() <= 0 {
-            item = behavior.finish_using(&mut item, &world, self);
-            self.stop_using_item();
+            let hand_item = self.inventory.lock().get_item_in_hand(hand).clone();
+            if item == hand_item {
+                let result = behavior.finish_using(&mut item, &world, self);
+                let mut inventory = self.inventory.lock();
+                let current_hand = inventory.get_item_in_hand_mut(hand);
+                if current_hand.item() == active.item() {
```
kdcokenny: in vanilla `completeUsingItem`, it also checks `!this.useItem.isEmpty()` before calling `finishUsingItem`.

also why check `current_hand.item() == active.item()` after `finish_using`? `finish_using` is supposed to change the item type for things like soup or honey bottles turning into empty bowls/bottles.

kdcokenny (changes requested):
also we def do need a test for this. i dont think you need #411 for this as long as you test the tick_active_item_use fn directly in player/tests.rs

steel-core/src/player/mod.rs
NandoBF: Isn't it possible for `finish_using` itself to modify the item? Without the check it would open up the same bug `on_use_tick` had, overwriting what `finish_using` did with the returned result.
I believe in vanilla it happens with a stack with more than 1 item of honey bottles, for example. With 2 honey bottles it would return the empty glass and modify the inventory to decrease the amount of items.
At least, I believe this can happen but there is a high chance I am wrong.

steel-core/src/player/mod.rs
```diff
@@ -414,39 +421,66 @@ impl Player {
...
             return;
         }
+
         let Some(active) = self.living_base.decrement_active_item_use() else {
-            self.inventory.lock().set_item_in_hand(hand, item);
             return;
         };
+
+        // TODO: Vanilla checks !useOnRelease here.
+        // implemented in PR #469
```
JunkyDeveloper: Didn't do only quick todo check, so no review
means this should be merged after 469?

NandoBF: To keep it completely in line with vanilla behavior, yes.

NandoBF: Also, do you mean that I shouldn't leave TODO comments?

steel-core/src/behavior/mod.rs
```diff
@@ -229,6 +229,8 @@ pub fn init_behaviors() {
     ITEM_BEHAVIORS.0.get_or_init(|| {
         let mut item_behaviors = ItemBehaviorRegistry::new();
         register_item_behaviors(&mut item_behaviors);
+        #[cfg(test)]
+        test_item_use::register_test_behavior(&mut item_behaviors);
```
coco875: feel like an hacky solution for me

coco875 (changes requested):

CuzImClicks (dismissed):

steel-core/src/player/mod.rs
```diff
@@ -445,39 +453,59 @@ impl Player {
...
+            let mut inventory = self.inventory.lock();
+            let current_stack = inventory.get_item_in_hand_mut(hand);
+            if *current_stack == original_stack {
+                *current_stack = item.clone();
+            }
+        }
+
         if self.active_item_use_hand() != Some(hand) {
-            self.inventory.lock().set_item_in_hand(hand, item);
+            self.stop_using_item(); // can't guarantee the item clears the flag
```
kdcokenny: if the callback releases main hand use and starts offhand use, this cancels the new session too. could we only do cleanup when no use is active and test that hand switch?

steel-core/src/player/mod.rs
```diff
@@ -445,39 +453,59 @@ impl Player {
...
+            let pre_finish = self.inventory.lock().get_item_in_hand(hand).clone();
+            if item == pre_finish {
+                let stack_before_finish = item.clone();
+                let outcome = behavior.finish_using(&mut item, &world, self);
+                self.apply_item_use_cooldown(&stack_before_finish);
+                let mut inventory = self.inventory.lock();
+                let current_hand = inventory.get_item_in_hand_mut(hand);
+                match outcome {
+                    FinishUseResult::Replaced(result) => *current_hand = result,
+                    FinishUseResult::InPlace if *current_hand == pre_finish => *current_hand = item,
```
kdcokenny: an equal clone still passes this check, so a callback that installs 2 bowls and shrinks its working copy to 1 overwrites the replacement. could we track replacements instead of comparing contents and test an equal clone? tick and release have the same issue.

steel-core/src/player/mod.rs
```diff
@@ -445,39 +453,59 @@ impl Player {
...
         };
+
         let world = self.get_world();
         behavior.on_use_tick(&world, self, &mut item, active.remaining_ticks());
 
+        {
+            let mut inventory = self.inventory.lock();
+            let current_stack = inventory.get_item_in_hand_mut(hand);
+            if *current_stack == original_stack {
+                *current_stack = item.clone();
```
kdcokenny: these direct writes skip `set_changed()`, including finish and release. could we use a tracked setter or `mutate_item_in_hand` and check the counter in the cookie test?

kdcokenny (changes requested):

NandoBF:
Added a `hand_write_generation` to `PlayerInventory` that tracks whether the hand slot was written during hooks. Now the clone only applies when the hand wasn't changed.
Changed tests to address the reviews.

PR #610, opened 2026-09-05
fix(chunk): let ticket revival win over an in-flight save preparation

### LINKED ISSUE #561: SIGABRT during startup pregen: ChunkGenerationTask::new expects distance-manager-created holders (scheduling race)
## Summary

Startup pregen intermittently aborts with a panic in `ChunkGenerationTask::new`:

```
The chunkholder should be created by distance manager before the generation task is scheduled. This occurring means there is a bug in the distance manager or you called this yourself.
```

The `StaticCache2D::create` closure (`steel-core/src/chunk/chunk_generation_task.rs:157`) looks up every chunk within the task's worst-case dependency radius in `chunk_map.chunks` and `expect`s it to exist. During pregen window churn, a neighbor holder is occasionally missing when the task is scheduled, so the `expect` fires (the build uses `panic = "abort"`, so this becomes SIGABRT with a core dump).

Verified call chain from the core dump (thread name `chunk-worker`, tokio blocking task):

```
ChunkMap::prepare_scheduling_epoch          chunk_map/mod.rs:1148 (closure)
 -> ChunkHolder::schedule_chunk_generation_task_b
   -> ChunkHolder::reschedule_chunk_task_b
     -> ChunkMap::schedule_generation_task_b   generation_readiness.rs:20
       -> ChunkGenerationTask::new             (target_status = Carvers at crash time)
         -> StaticCache2D::create closure      chunk_generation_task.rs:157  <- panic
```

`schedule_generation_task_b` is the only call site of `ChunkGenerationTask::new` in the tree.

## Frequency

2 aborts in ~75 pregen runs at `PREGEN_SIZE=65` (65x65 chunks, default `PREGEN_WINDOW_SIZE=32`): one on commit `38e27a3`, one on `bfd6882`. ~11 further runs at smaller pregen sizes had no abort. Machine: pinned 4 P-cores, Linux laptop. Not tied to a specific recent commit.

## Backtrace (gdb, profiling core)

```
#2  abort ()
#3  std::sys::pal::unix::abort_internal
#6  panic_abort::__rust_start_panic
#12 core::panicking::panic_fmt
#14 core::option::expect_failed
#16 steel_core::chunk::chunk_generation_task::{impl#1}::new::{closure#0} at steel-core/src/chunk/chunk_generation_task.rs:157
#17 steel_core::chunk::chunk_generation_task::{impl#0}::create::{closure#0} at steel-core/src/chunk/chunk_generation_task.rs:51
    (StaticCache2D::create -> iteration over worst_case_radius positions)
#30 ChunkMap::schedule_generation_task_b
#41 ChunkMap::prepare_scheduling_epoch
#42 ChunkMap::spawn_scheduling_epoch (tokio spawn_blocking)
```

## Secondary observation

No panic message appears in console output or the rotated log file for the aborted runs. With `panic = "abort"` the hook's stderr message seems to get lost (or is never emitted) before the abort. Worth checking whether panic output survives the abort path in logging config.

## Repro

Loop `PREGEN_SIZE=65 steel` runs in fresh dirs under sustained load; the abort reproduces roughly once every ~35-40 runs.

Found while benchmarking #479; both the PR build and the merge-base build crashed with identical panic site, thread, and signal, so it is unrelated to that PR.

Purdze:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Fixes #561 

fixes the fatal `chunk-worker` panic players hit, introduced by #307

`ChunkGenerationTask::new` needs a holder in `ChunkMap::chunks` for every position in its dependency radius. reviving an unloading holder used a CAS that only accepted `SAVE_LIFECYCLE_UNLOADING`, so it failed while a save snapshot held `SAVE_LIFECYCLE_PREPARING`. the revival got parked in `deferred_revivals`, leaving a position with a ticket level and no holder for a full epoch. neighbours revived in the same commit were scheduled against that hole

dying is a wide net for the race: `Player::tick_death` drops the whole player-view ticket 20 ticks after death, those chunks are freshly dirty from the item drops, so a save spawns for each one just before the ticket returns

the fix is vanilla's: revival always wins, and the in-flight save discovers it lost. `revive_from_unloading` is infallible now, `ChunkSavePreparationGuard::finish` hands the snapshot back only if the phase still owned the holder, and `deferred_revivals` is deleted. since revival can now win, `prepare_chunk_save` treats "promoted underneath the snapshot" as an abandon rather than a panic

## How this was tested

- `cargo test`: full workspace green, 0 failures
- `cargo clippy -r --all-targets --all-features` and `cargo fmt --all --check`: clean
- the regression test is falsified. `revival_during_save_preparation_keeps_the_generation_neighborhood_complete` pins a holder in `PREPARING`, revives it, then schedules a generation task over its neighbourhood. i temporarily reinstated the old deferral and it reproduced the exact panic at `chunk_generation_task.rs:157`. reverted, re-ran green
- three more tests cover the lifecycle and the new abandon path: `revival_wins_over_an_in_flight_save_preparation`, `an_uncontested_save_preparation_keeps_its_snapshot`, `chunk_save_abandons_a_proto_status_snapshot_of_a_promoted_chunk`

i died 20+ times without `keepInventory` in loaded terrain and no `chunk-worker` panic

## Checklist

- [x] Self-reviewed the diff

`ChunkHolder`, `ChunkGenerationTask`, `ChunkMap`, `ChunkStorage`

## Additional notes

two pre-existing things found along the way, both left alone

1. `Arc::strong_count` is the wrong unload gate. `process_unloads` skips the save and unload whenever `Arc::strong_count(holder) != 1`. that means "nobody anywhere holds an Arc", not "generation depends on this". the gameplay lookup cache, light workset and ticking snapshot all clone holder Arcs, so any of them lingering delays an unload for unrelated reasons. vanilla uses an explicit `generationRefCount` registered as a save dependency. steel already has that in `ChunkSaveDependency` / `active_save_dependencies`, `process_unloads` just doesn't consult it. worth its own PR. note `ChunkSavePreparationGuard`'s doc currently leans on the strong-count gate to rule out a stale guard releasing a second phase, so that changes with it
2. POI data is dropped before the save reads it. `begin_chunk_unload` calls `poi_storage.remove_chunk(pos)` before the save runs, so `pois_to_persistent` always writes an empty POI list on unload and loses `free_tickets`. vanilla's `ChunkMap.save` calls `poiManager.flush(pos)` first

### DIFF
--- a/steel-core/src/chunk/chunk_holder.rs
+++ b/steel-core/src/chunk/chunk_holder.rs
@@ -219,22 +219,34 @@ impl Drop for ChunkSaveDependency {
     }
 }
 
+/// Reservation of a holder's save-preparation phase.
+///
+/// It excludes a second preparation but not a ticket revival; [`Self::finish`] hands the snapshot
+/// back only if the phase still owned the holder. One guard exists per holder: it keeps a strong
+/// reference, and `ChunkMap::process_unloads` only saves a holder whose strong count is 1.
 pub(crate) struct ChunkSavePreparationGuard {
     holder: Arc<ChunkHolder>,
+    /// Whether the lifecycle was already handed back, so `Drop` does not repeat it.
+    released: bool,
+}
+
+impl ChunkSavePreparationGuard {
+    /// Ends the phase, returning `value` only if no revival reclaimed the holder meanwhile.
+    #[must_use = "a snapshot prepared while losing a revival race must be discarded"]
+    pub(crate) fn finish<T>(mut self, value: T) -> Option<T> {
+        self.released = true;
+        self.holder.end_save_preparation().then_some(value)
+    }
 }
 
 impl Drop for ChunkSavePreparationGuard {
     fn drop(&mut self) {
-        let result = self.holder.save_lifecycle.compare_exchange(
-            SAVE_LIFECYCLE_PREPARING,
-            SAVE_LIFECYCLE_UNLOADING,
-            Ordering::AcqRel,
-            Ordering::Acquire,
-        );
-        assert!(
-            result.is_ok(),
-            "chunk save preparation ended outside the preparing lifecycle"
-        );
+        if self.released {
+            return;
+        }
+
+        // The preparation bailed before a snapshot existed, so nothing escaped.
+        let _ = self.holder.end_save_preparation();
     }
 }
 
@@ -711,19 +723,37 @@ impl ChunkHolder {
             .ok()
             .map(|_| ChunkSavePreparationGuard {
                 holder: Arc::clone(self),
+                released: false,
             })
     }
 
-    /// Attempts to reactivate an unloading holder without waiting for save preparation.
-    pub(crate) fn try_revive_from_unloading(&self) -> bool {
+    /// Leaves the preparing lifecycle, reporting whether the phase still owned the holder.
+    fn end_save_preparation(&self) -> bool {
         self.save_lifecycle
             .compare_exchange(
+                SAVE_LIFECYCLE_PREPARING,
                 SAVE_LIFECYCLE_UNLOADING,
-                SAVE_LIFECYCLE_ACTIVE,
                 Ordering::AcqRel,
                 Ordering::Acquire,
             )
             .is_ok()
+    }
+
+    /// Reactivates an unloading holder even while a save snapshot is in flight, as vanilla's
+    /// `ChunkMap.updateChunkScheduling` does via `pendingUnloads`. It cannot fail: a ticketed
+    /// position without a holder in `ChunkMap::chunks` breaks `ChunkGenerationTask::new`. A single
+    /// `try_update` covers both source states, since a finishing preparation moves `PREPARING` back
+    /// to `UNLOADING` between sequential CAS attempts.
+    pub(crate) fn revive_from_unloading(&self) {
+        let previous =
+            self.save_lifecycle
+                .try_update(Ordering::AcqRel, Ordering::Acquire, |lifecycle| {
+                    (lifecycle != SAVE_LIFECYCLE_ACTIVE).then_some(SAVE_LIFECYCLE_ACTIVE)
+                });
+        debug_assert!(
+            previous.is_ok(),
+            "an already active chunk holder was revived from unloading"
+        );
     }
 
     /// Applies a step to the chunk.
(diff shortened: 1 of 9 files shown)


steel-core/src/chunk/chunk_holder.rs
```diff
@@ -164,6 +164,8 @@ pub struct ChunkHolder {
     /// Number of save dependencies that have not completed yet.
     active_save_dependencies: AtomicUsize,
     /// Coordinates unloading revival with the short immutable save-preparation phase.
+    ///
+    /// Revival always wins; see [`ChunkHolder::revive_from_unloading`].
```
coco875: it's useful ?

steel-core/src/chunk/chunk_holder.rs
```diff
@@ -219,22 +221,38 @@ impl Drop for ChunkSaveDependency {
...
 }
 
+/// Reservation of a holder's save-preparation phase.
+///
+/// The phase excludes a second preparation but deliberately does not exclude a ticket revival. End
+/// it with [`Self::finish`]: the snapshot is moved into the guard and only comes back out when the
+/// phase still owned the holder, so a preparation that lost the revival race cannot reach disk.
+///
+/// Only one guard can exist per holder at a time: it keeps a strong reference, and
+/// `ChunkMap::process_unloads` only spawns a save for a holder whose strong count is 1.
```
coco875: Can it be shorter ?

steel-core/src/chunk/chunk_holder.rs
```diff
@@ -711,21 +729,46 @@ impl ChunkHolder {
...
+    /// Vanilla revives unconditionally: `ChunkMap.updateChunkScheduling` pulls the holder straight
+    /// back out of `pendingUnloads` and lets the pending unload find out that it lost. A concurrent
+    /// preparation learns that from [`ChunkSavePreparationGuard::finish`] and throws its snapshot
+    /// away. Revival cannot be allowed to fail: a position that holds a ticket level but has no
+    /// entry in `ChunkMap::chunks` is a hole in the dependency grid that
+    /// `ChunkGenerationTask::new` cannot resolve.
+    ///
+    /// A single `try_update` is required rather than one CAS per source state: a preparation ending
+    /// concurrently moves `PREPARING` back to `UNLOADING`, so sequential attempts could both miss
+    /// and spuriously fail.
```
coco875: I don't think that compare to vanilla are useful

steel-core/src/chunk_saver/storage/tests/chunks_sections.rs
```diff
@@ -15,6 +15,27 @@ fn chunk_save_rejects_full_status_for_proto_data() {
...
+        single_empty_section(),
+        ChunkPos::new(0, 0),
+        0,
+        16,
+        Weak::new(),
+    );
+    let _ = chunk.promote_to_full();
+
+    // A revival can promote a chunk after its status was captured, so the stale proto status is
+    // an abandoned snapshot rather than corruption.
```
coco875: does the comment are useful ?

steel-core/src/chunk_saver/storage/chunk.rs
```diff
@@ -263,11 +263,18 @@ impl ChunkStorage {
     }
 
     /// Converts a runtime section to persistent format.
+    ///
+    /// Sections still in worldgen `Building` mode are finalized under the same write guard that
+    /// serializes them: worldgen re-enters `Building` on every column write, so a chunk that
+    /// resumes generation mid-snapshot would defeat any earlier finalization pass.
```
coco875: it explain modification made no ? I don't think it's useful

steel-core/src/chunk_saver/storage/mod.rs
```diff
@@ -556,9 +556,14 @@ impl ChunkStorage {
     /// Call this during the holder's snapshot-preparation phase, then pass the result to
     /// `save_chunk_data` after ending that phase.
     ///
+    /// Returns `None` when there is nothing to write, and when the chunk was promoted underneath
+    /// the snapshot: `ChunkHolder::revive_from_unloading` lets a proto chunk resume generation
+    /// while this runs, which makes the `status` captured beforehand stale rather than corrupt.
+    /// Callers treat `None` as "leave the chunk dirty and save it later".
+    ///
```
coco875, line 558: the comment start being pretty long

steel-core/src/chunk_saver/storage/mod.rs
```diff
@@ -656,7 +657,13 @@ impl ChunkStorage {
             // Proto ticks are pending, so Vanilla ignores the current game
             // time when serializing their already-relative delays.
             let Some(snapshot) = chunk.scheduled_ticks.snapshot(0) else {
-                panic!("Proto chunk scheduled-tick container was finalized before saving");
+                // Promotion to Full closes the proto container.
```
coco875, line 655: I don't think that the comment are useful

coco875 (changes requested):
Did you check if it make a difference in performance ?

CuzImClicks (approved):

PR #618, opened 2026-09-06
fix: match vanilla friction and air drag in living entity travel

darkstarworks:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Three gaps between `LivingEntity`'s walking path and vanilla's.

* `getFrictionInfluencedSpeed` compensates for slippery ground by dividing by the cube of the block's friction. Above the default friction that cancels out the extra slide, but below it hands out a speed boost instead. Vanilla only applies the compensation past the default and returns the plain speed otherwise. No vanilla block is grippier than the default today, so nothing changes in practice, but the guard belongs here before one is.
* The `friction_modifier` and `air_drag_modifier` attributes were ignored, so anything that changes how an entity slides or how much the air drags on it had no effect. Both are now folded in the way vanilla does, through a shared `compute_modified_friction`.
* The bare `0.6`, `0.91` and `0.98` in that path are named while it is being touched.

## How this was tested

Unit tests for the modifier helper (a modifier of 1 leaves friction alone, 0 removes the slipperiness, a large value stays clamped) and for the grippy-ground guard.

Two behavioural tests walk a real pig: it moves at vanilla's 0.1376 blocks per tick on ordinary ground, and slides far further on ice once it stops walking.

`cargo fmt --all --check`, `cargo clippy -r --workspace`, `typos` and `cargo test --workspace` all pass.

## Screenshots / logs

Not applicable, this is server-side movement logic.

## Checklist

- [x] Self-reviewed the diff

LivingEntity

## Additional notes

* **Attribute lookup.** These two attributes are read with `get_value(...).unwrap_or(...)` rather than `required_value`. Extracted entity data only lists the attributes a type overrides, and no entity type declares either of these, so `required_value` panics on the first walking mob. A type that declares neither keeps the untouched value, so behaviour is unchanged for everything that exists today.
* **The speed scale keeps vanilla's inlined `0.216_000_02`** rather than recomputing `0.6³`, so the float arithmetic stays bit-identical to vanilla.


steel-core/src/entity/living_entity.rs
```diff
@@ -3104,3 +3163,166 @@ pub(crate) fn shearing_loot_items_with_rng<R: rand::Rng, E: LivingEntity + ?Size
...
+        // mob, so the top speed barely changes. What changes is how long it
+        // takes to stop, which is what makes ice feel like ice.
+        assert!(
+            on_ice.coasted > on_grass.coasted * 5.0,
+            "ice should keep the pig sliding, got {} against {}",
+            on_ice.coasted,
+            on_grass.coasted
+        );
+    }
+}
```
JunkyDeveloper: are these really needed? and maybe move it to another file because we hit 3k lines already here...

darkstarworks:
Good catch! It also matters sooner than it looks: with the friction modifier from this PR, a value of 2 (the sticky sulfur cube) turns 0.6 ground into 0.2. The old formula would've caused it to walk 27 times faster.
Added the branch plus a test for friction below 0.6.

steel-core/src/entity/living_entity.rs
```diff
@@ -2383,19 +2401,40 @@ pub trait LivingEntity: Entity {
...
+        let cubed = block_friction * block_friction * block_friction;
+        self.get_speed() * (DEFAULT_FRICTION_SPEED_SCALE / cubed)
+    }
+
+    /// The entity's air-drag modifier attribute.
+    fn air_drag_modifier(&self) -> f32 {
+        self.attributes()
+            .lock()
+            .get_value(vanilla_attributes::AIR_DRAG_MODIFIER)
+            .unwrap_or(NO_FRICTION_MODIFIER) as f32
```
CuzImClicks:
```suggestion
            .unwrap_or(vanilla_attributes::AIR_DRAG_MODIFIER.default_value) as f32
```

CuzImClicks (changes requested):

CuzImClicks (approved):

steel-core/src/entity/living_entity.rs
```diff
@@ -4,6 +4,24 @@ use steel_registry::{DyeColor, vanilla_custom_stats};
...
+const DEFAULT_BLOCK_FRICTION: f32 = 0.6;
+/// The scale that cancels default ground friction, kept as the inlined literal
+/// so the arithmetic matches.
+const DEFAULT_FRICTION_SPEED_SCALE: f32 = 0.216_000_02;
+/// Horizontal air drag out of fluid.
+const BASE_HORIZONTAL_AIR_DRAG: f32 = 0.91;
+/// Vertical air drag out of fluid (flying animals use the horizontal one).
+const BASE_VERTICAL_AIR_DRAG: f32 = 0.98;
+/// Identity modifier, used when an entity type does not declare the attribute.
+const NO_FRICTION_MODIFIER: f64 = 1.0;
```
coco875, line 17: by memory we use some of this const in other place can you check that ?

coco875 (changes requested):

steel-core/src/entity/living_entity.rs
darkstarworks: Good call, leash.rs had its own 0.91 (`BASE_HORIZONTAL_FRICTION`) for the same drag. The leash uses the shared const now, which also closes a small precision gap: vanilla's `angularFriction` returns `0.91F`, and the old f64 `0.91` wasn't quite that value. The 0.98s on items, falling blocks and XP orbs are their own drags in vanilla, so I left those.

JunkyDeveloper (changes requested):
I never submitted that Review

PR #620, opened 2026-09-06
feat: complete some todos

GreenedDev:
## Type of change

- [x] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## How this was tested
in-game

## Checklist

- [x] Self-reviewed the diff

potent_sulfur_block.rs
candle_cake_block.rs
spore_blossom_block.rs

### DIFF
--- a/steel-core/src/behavior/blocks/decoration/candle_cake_block.rs
+++ b/steel-core/src/behavior/blocks/decoration/candle_cake_block.rs
@@ -81,6 +81,7 @@ impl BlockBehavior for CandleCakeBlock {
             && state.get_value(LIT)
         {
             world.set_block(pos, state.set_value(LIT, false), UpdateFlags::UPDATE_ALL);
+            // TODO: particles!
             world.play_block_sound(
                 &sound_events::BLOCK_CANDLE_EXTINGUISH,
                 pos,
--- a/steel-core/src/behavior/blocks/decoration/candle_cake_block.rs
+++ b/steel-core/src/behavior/blocks/decoration/candle_cake_block.rs
@@ -28,7 +28,6 @@ use crate::{
 
 /// Behavior for Candle Cakes
 /// TODO:
-/// - [ ] animation ticks
 /// - [ ] onExplosion
 #[block_behavior]
 pub struct CandleCakeBlock {
@@ -82,7 +81,6 @@ impl BlockBehavior for CandleCakeBlock {
             && state.get_value(LIT)
         {
             world.set_block(pos, state.set_value(LIT, false), UpdateFlags::UPDATE_ALL);
-            // TODO: particles!
             world.play_block_sound(
                 &sound_events::BLOCK_CANDLE_EXTINGUISH,
                 pos,
--- a/steel-core/src/behavior/blocks/decoration/candle_cake_block.rs
+++ b/steel-core/src/behavior/blocks/decoration/candle_cake_block.rs
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+use glam::DVec3;
 use steel_macros::block_behavior;
 use steel_registry::{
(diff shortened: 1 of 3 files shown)


steel-core/src/behavior/blocks/decoration/candle_cake_block.rs
```diff
@@ -82,7 +81,6 @@ impl BlockBehavior for CandleCakeBlock {
             && state.get_value(LIT)
         {
             world.set_block(pos, state.set_value(LIT, false), UpdateFlags::UPDATE_ALL);
-            // TODO: particles!
```
JunkyDeveloper, line 85: where is the particles implemented?
I checked vanilla and it does it

GreenedDev: I visually compared vanilla and steel and extinguishing fire on a cake looks same

GreenedDev: maybe it's client sided somehow?

JunkyDeveloper: for your own the client plays sounds, particles. So did you tested with a second account?

steel-core/src/behavior/blocks/decoration/candle_cake_block.rs
```diff
@@ -28,14 +30,14 @@ use crate::{
...
 /// TODO:
-/// - [ ] animation ticks
 /// - [ ] onExplosion
 #[block_behavior]
 pub struct CandleCakeBlock {
     block: BlockRef,
 }
 
 const LIT: &BoolProperty = &BlockStateProperties::LIT;
+const PARTICLE_OFFSETS: [DVec3; 1] = [DVec3::new(8.0 * 0.0625, 16.0 * 0.0625, 8.0 * 0.0625)];
```
JunkyDeveloper, line 40: is this not only 1 element in the array, so why a loop below?

JunkyDeveloper (changes requested):

steel-core/src/behavior/blocks/decoration/candle_cake_block.rs
GreenedDev: that's how vanilla does it. 
   private static final Iterable<Vec3> PARTICLE_OFFSETS = List.of(new Vec3(8.0, 16.0, 8.0).scale(0.0625));

CuzImClicks (approved):

steel-core/src/behavior/blocks/decoration/candle_cake_block.rs
JunkyDeveloper: Still doesn't really answer the question why creating a vec to loop over one element :D I would say hardcore that one thing is better with a comment.

Because our goal is not to reproduce shitty Mojang code 😂

CuzImClicks: the compiler does not care. since its a loop over a const its gonna get unrolled into the identical code as it would be without the loop anyways. idk why vanilla does it this way but for rust it doesnt really matter i think

GreenedDev: so what do i do?

JunkyDeveloper: ok if the compiler don't cares I'm fine and resolve this thread

JunkyDeveloper (approved):

PR #623, opened 2026-09-07
feat: implement the enchanting table

JosepBove:
## Type of change

- [x] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Implements the vanilla enchanting table end to end:

- `EnchantingTableBlock`: opens the menu on use, creates its block entity, counts bookshelf power over the vanilla 32-offset ring with the `enchantment_power_provider` / `enchantment_power_transmitter` tags, copies a placed item's `CUSTOM_NAME` onto the block entity, and is not pathfindable.
- `EnchantingTableBlockEntity`: persists `CustomName`; the display name feeds the menu title. The book animation is client-side, so there is no server ticker.
- `EnchantmentMenu` (`EnchantmentKind`): single-item slot, lapis-only slot, the ten vanilla data slots (three costs, seed, three enchantment clues, three level clues, sent as shorts with full precision kept server-side), seeded offers, level and lapis payment, book to enchanted book transmute, vanilla quick-move rules, and `stillValid` range checks. Both table slots return to the player on close.
- Offer selection math ported from `EnchantmentHelper` (`getEnchantmentCost`, `selectEnchantment`, `getAvailableEnchantmentResults`, `filterCompatibleEnchantments`) driven by `LegacyRandom`, with the same RNG call order as vanilla, plus a `WeightedRandom` port and a bit-exact Java `Math.round(float)` in `steel-utils`.
- `Enchantment::get_min_cost`, `get_max_cost`, `is_primary_item` and `EnchantmentCost::calculate` in `steel-registry`.
- Player `enchantmentSeed` stored beside the XP fields in `Experience`, persisted in the player save (data version 7, storage header 10), written as `XpSeed` in command NBT, and rerolled on load when zero like vanilla.
- Container button clicks: a `MenuKind::click_menu_button` hook and a real `handle_container_button_click` (previously a logging stub). Spectator clicks are ignored as in vanilla.

Vanilla reference: Minecraft 26.2 `EnchantmentMenu`, `EnchantingTableBlock`, `EnchantingTableBlockEntity`, `EnchantmentHelper`, `WeightedRandom`, `Player`.

Out of scope, left for follow-ups: `ENCHANTED_ITEM` advancement trigger (no advancement foundation yet), client particles, dropping a named table as a named item on break (loot `copy_components`, see #591), and wiring `enchant_with_levels` loot functions to the new `select_enchantment`.

## How this was tested

Target: Minecraft 26.2 (`0.15.3+mc26.2`), Windows 11.

Automated, all green on the branch head:
- `cargo test --workspace`
- `RUSTFLAGS="-Dwarnings" cargo clippy -r --all-targets --all-features`
- `cargo fmt --all --check`
- `typos`

New tests cover: weighted selection and Java rounding (including the `0.49999997` midpoint case); cost windows and primary vs supported items; offer determinism, pairwise compatibility, highest-level window scan, book bypass of the primary filter; seed persistence round trip and zero-seed reroll; button click routing by container id; the single-item slot cap through the real insert path; block entity name NBT round trip; block entity creation, bookshelf ring, transmitter blocking, placement name copy, and menu opening on use; and the menu itself through `Menu::clicked` / `Menu::click_menu_button`: offers with a full ring (third offer at 30), identical offers for the same seed, unenchantable items yielding nothing, rejection without lapis or levels, spending 3 lapis and 3 levels with a seed reroll, creative skipping lapis but still spending levels, book to enchanted book with stored enchantments, single-item quick move, validity by block and range, protocol short wrapping, and the absolute data-slot wire order.

In-game: a manual checklist for this feature (block and menu basics, bookshelf power, seeded offers, lapis and level payment in survival, creative, books, shift-click and closing, range and validity, custom name across a restart, per-player seed persistence, and enchantment effects applying in combat) is prepared and will be run against a 26.2 client; results will be posted in this PR before merge.

## Screenshots / logs

Particles of bookshelves
image

image

image

image

image

image

image

image

image

## Checklist

- [x] Self-reviewed the diff

- EnchantingTableBlock
- EnchantmentMenu
- EnchantingTableBlockEntity

## Additional notes

- The player save schema changed, so `PLAYER_DATA_VERSION` is 7 and `PLAYER_STORAGE_VERSION` is 10; existing player files are rejected at the header as the project does not ship migrations yet.
- Vanilla keeps `enchantmentSeed` as a separate `Player` field; Steel stores it inside `Experience` because vanilla saves it beside the XP fields and `onEnchantmentPerformed` mutates both. This is documented on the field.
- `ItemStack::with_count_and_patch` is used for the book transmute; it sanitizes patch entries equal to the new prototype's defaults, where vanilla's `transmuteCopy` forgets any component the new item defines. The two agree for anything reachable in play, and the difference is commented.
- Bookshelf world reads run through `ContainerLockGuard::run_unlocked` so no block state is read while the menu's containers are locked.


steel-core/src/behavior/blocks/container/enchanting_table_block.rs
```diff
@@ -0,0 +1,323 @@
...
+    /// the player-placement path here until an implicit-component hook exists
+    /// for the others.
+    fn set_placed_by(
+        &self,
+        _state: BlockStateId,
+        world: &Arc<World>,
+        pos: BlockPos,
+        source: &PlacementSource<'_>,
+    ) {
+        let Some(name) = source.with_item(|stack| stack.get(CUSTOM_NAME).cloned()) else {
```
Joniii11: in vanilla handles CUSTOM_NAME through blockentity implicit components so this only covers the placement path which we dont want we want ttocover every path lol

steel-core/src/behavior/blocks/container/enchanting_table_block.rs
```diff
@@ -0,0 +1,323 @@
...
+            .get_block();
+        let transmitter = world
+            .get_block_state(pos.offset(offset.x / 2, offset.y, offset.z / 2))
+            .get_block();
+        REGISTRY
+            .blocks
+            .is_in_tag(provider, &BlockTag::ENCHANTMENT_POWER_PROVIDER)
+            && REGISTRY
+                .blocks
+                .is_in_tag(transmitter, &BlockTag::ENCHANTMENT_POWER_TRANSMITTER)
```
Joniii11: so basically vanilla shortcircuits after the provider check ts eagerly reads the transmitter too doubling blockstate reads for none provider offsets and the real mcuses shortcircuit && 

```suggestion 
        let provider = world
            .get_block_state(pos.offset(offset.x, offset.y, offset.z))
            .get_block();
        if !REGISTRY
            .blocks
            .is_in_tag(provider, &BlockTag::ENCHANTMENT_POWER_PROVIDER)
        {
            return false;
        }

        let transmitter = world
            .get_block_state(pos.offset(offset.x / 2, offset.y, offset.z / 2))
            .get_block();
        REGISTRY
            .blocks
            .is_in_tag(transmitter, &BlockTag::ENCHANTMENT_POWER_TRANSMITTER)
```

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    /// Vanilla `slotsChanged`: rerolls the offers whenever a menu slot changes.
+    fn slots_changed(
+        &mut self,
+        behavior: &mut MenuBehavior,
+        guard: &mut ContainerLockGuard,
+        _player: &Player,
+    ) {
+        self.update_offers(behavior, guard);
```
Joniii11, line 267: minecraft only recalculates offers when enchantSlots changes but we call this after unrelated inventory clicks too so those clicks can reroll offers

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,401 @@
...
+            log::debug!(
+                "{} pressed invalid button id: {button_id}",
+                player.gameprofile.name
+            );
+            return false;
+        };
+        let enchantment_cost = offer as i32 + 1;
+        let has_infinite_materials = player.has_infinite_materials();
+
+        let mut guard = behavior.lock_all_containers();
```
Joniii11: ts also locks the player inventory although the button only needs enchant_slots update_offers then releases/relocks all guards so maybe could we lock only the enchanting container here for performance!!! speeed go brrrt

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,401 @@
...
+        let container_id = ContainerId::from_arc(&self.enchant_slots);
+        let (stack, mut currency) = {
+            let container = guard
+                .get(container_id)
+                .expect("enchant container not locked");
```
Joniii11: well i am not the biggest fan of that expect here could we remove it?

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,401 @@
...
+/// `enchantment_seed` is the opening player's vanilla `enchantmentSeed`.
+#[must_use]
+pub fn enchantment(
+    inventory: Shared<PlayerInventory>,
+    container_id: u8,
+    pos: BlockPos,
+    world: &Arc<World>,
+    enchantment_seed: i32,
+) -> Menu {
+    let enchant_slots = SimpleContainer::new(2).into_shared();
```
Joniii11: magic numver

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,401 @@
...
+    let enchant_slots_ref = ContainerRef::from(Arc::clone(&enchant_slots));
+
+    let mut builder = MenuBuilder::new(&vanilla_menu_types::ENCHANTMENT, container_id);
+    let item = builder.section_at(
+        enchant_slots_ref.clone(),
+        [0],
```
Joniii11: maybe also use a const for this? cause i see that the slot 0 is otem and used a bit more often like name it uh ITEM_SLOT

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,401 @@
...
+        SectionKind::custom(|container, index| {
+            Box::new(EnchantItemSlot::new(container.clone(), index))
+        }),
+    );
+    let lapis = builder.section_at(
+        enchant_slots_ref,
+        [1],
```
Joniii11: same here with item before so this would be then LAPIS_SLOT yk

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,401 @@
...
+        }
+        self.sync_data(behavior);
+    }
+}
+
+impl MenuKind for EnchantmentKind {
+    /// Vanilla `stillValid`: the table must still stand there and be in reach.
+    fn still_valid(&self, _behavior: &MenuBehavior, player: &Player) -> bool {
+        self.world.get_block_state(self.block_pos).get_block() == &vanilla_blocks::ENCHANTING_TABLE
+            && player.is_within_block_interaction_range_with_buffer(self.block_pos, 4.0)
```
Joniii11: magic number 4.0 lol

steel-utils/src/java.rs
```diff
@@ -149,9 +149,28 @@ fn floating_to_string(
     output
 }
 
+/// Rounds like Java's `Math.round(float)`: half-up at the exact midpoint,
+/// using the JDK's bit-level algorithm so floats just below `n + 0.5` do not
+/// round up through float addition. Saturates and maps NaN to 0 like Java.
+#[must_use]
+pub const fn round_f32(value: f32) -> i32 {
```
Joniii11, line 171: so many magic hexxes ore number idk what you wanna call it

steel-core/src/inventory/menu/kind.rs
```diff
@@ -33,6 +33,18 @@ pub trait MenuKind: ErasedType + Send + Sync {
     /// any result. No-op for kinds without a rename input.
     fn on_rename(&mut self, _behavior: &mut MenuBehavior, _name: String, _player: &Player) {}
 
+    /// Handles a client button press (vanilla `clickMenuButton`). Returns true
+    /// when the press changed the menu so the caller re-syncs it. No-op for
+    /// kinds without buttons.
```
Joniii11: true means the button was acepte/handled not necessarily tha it changed the menu ts impl can return true without mutating it so bit misleading ts

steel-core/src/inventory/menu/kinds/enchantment_menu/tests.rs
```diff
@@ -0,0 +1,323 @@
...
+    );
+    player.experience.lock().set_levels(30);
+
+    assert!(menu.click_menu_button(2, &player));
+
+    assert!(table_item(&menu).is_enchanted());
+    assert!(table_lapis(&menu).is_empty());
+    let experience = player.experience.lock();
+    assert_eq!(experience.level(), 27);
+    assert_ne!(experience.enchantment_seed(), SEED);
```
Joniii11: mc does not guarantee that a rerolled seed differs from the previous value so this is technically wrongish

steel-core/src/inventory/menu/kinds/enchantment_menu/tests.rs
```diff
@@ -0,0 +1,323 @@
...
+    drop(experience);
+    let kind = kind(&menu);
+    assert_ne!(kind.enchantment_seed, SEED);
```
Joniii11: mc does not guarantee that a rerolled seed differs from the previous value so this is technically wrongish x2

steel-core/src/player/tests.rs
```diff
@@ -1268,6 +1268,39 @@ fn persistent_player_data_restores_independent_experience_fields_and_score() {
...
+fn zero_saved_enchantment_seed_is_rerolled_on_load() {
+    init_vanilla_registry();
+    let world = fresh_test_world("player_data_enchantment_seed_reroll");
+    let player = TestPlayerBuilder::new(Arc::clone(&world), "SeedTester", 1).build();
+    assert_eq!(player.experience.lock().enchantment_seed(), 0);
+
+    let data = PersistentPlayerData::from_player(&player);
+    assert_eq!(data.enchantment_seed, 0);
+    data.apply_to_player_without_location(&player);
+    assert_ne!(player.experience.lock().enchantment_seed(), 0);
```
Joniii11: mc does not guarantee that a rerolled seed differs from the previous value so this is technically wrongish x3

steel-registry/src/enchantment/mod.rs
Joniii11: yes ik this is not part of your pr but would it be possible since you modified the file already fix this?: ts reparses static tag refs into owned Identifiers in the offer hot path causing many shortlived sring allocations per reroll can these stay borrowed/typed

```suggestion
fn parse_tag_ref(tag_ref: &'static str) -> Option<Identifier> {
    let without_hash = tag_ref.strip_prefix('#')?;
    Some(if let Some((ns, path)) = without_hash.split_once(':') {
        Identifier::new_static(ns, path)
    } else {
        Identifier::vanilla_static(without_hash)
    })
}
```

Joniii11 (changes requested):

steel-core/src/behavior/blocks/container/enchanting_table_block.rs
JosepBove: Done. Block entities now have apply_implicit_components and collect_implicit_components hooks with final apply_components_from_item_stack and collect_components, mirroring vanilla applyComponents. BlockItem::place calls it before setPlacedBy so every placement path names the table, and the set_placed_by override is gone.

steel-core/src/behavior/blocks/container/enchanting_table_block.rs
JosepBove: Done, the provider check returns early before the transmitter state is read.

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
JosepBove: Done. SimpleContainer now counts set_changed calls like vanilla timesChanged and the menu only recomputes when the table container's count moved. Added a test that removes the bookshelves mid session and checks the offers hold until lapis enters the table.

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
JosepBove: Done, the button handler locks only the table container now.

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
JosepBove: Removed both, they return false instead.

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
JosepBove: Named TABLE_SLOT_COUNT.

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
JosepBove: Added ITEM_SLOT and used it in the menu and the button handler.

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
JosepBove: Added LAPIS_SLOT.

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
JosepBove: Container::DEFAULT_DISTANCE_BUFFER already existed but was unused, so the enchantment, anvil and crafting menus and the block entity validity check all use it now.

steel-utils/src/java.rs
JosepBove: Replaced with FloatConsts values derived the same way the JDK derives them.

steel-core/src/inventory/menu/kind.rs
JosepBove: Reworded to say it returns whether the press was handled.

steel-core/src/inventory/menu/kinds/enchantment_menu/tests.rs
JosepBove: Agreed, nextInt can repeat. The test now asserts the player to menu seed hand off instead.

steel-core/src/inventory/menu/kinds/enchantment_menu/tests.rs
JosepBove: Same fix as above.

steel-core/src/player/tests.rs
JosepBove: The zero seed rule now takes its reroll source as a parameter so the test pins it deterministically, and a separate test covers the 4242 round trip.

steel-registry/src/enchantment/mod.rs
JosepBove: Done, went one step further: the build script emits static Identifier values for supported_items, primary_items and exclusive_set, so there is no parsing at runtime at all and parse_tag_ref is gone.

steel-core/src/behavior/blocks/container/enchanting_table_block.rs
```diff
@@ -0,0 +1,322 @@
...
+            while z <= 2 {
+                if x.abs() == 2 || z.abs() == 2 {
+                    offsets[count] = IVec3::new(x, y, z);
+                    count += 1;
+                }
+                z += 1;
+            }
+            x += 1;
+        }
+        y += 1;
```
coco875, line 48: why not for loop ?

steel-core/src/block_entity/entities/shulker_box.rs
```diff
@@ -381,10 +384,10 @@ impl BlockEntity for ShulkerBoxBlockEntity {
         nbt.insert("Items", NbtList::Compound(items));
     }
 
-    fn apply_components_from_item(&self, item: &ItemStack) {
-        let Some(contents) = item.get(CONTAINER) else {
-            return;
-        };
+    /// Vanilla `BaseContainerBlockEntity.applyImplicitComponents`, container part:
+    /// `CONTAINER` (or empty) is copied into the slots.
+    fn apply_implicit_components(&self, components: &mut ImplicitComponentGetter<'_>) {
```
coco875, line 389: I don't think precise the vanilla name a useful because it's very transparent

steel-core/src/block_entity/components.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub fn get_or_default<T: Component + DowncastType + Clone>(
+        &mut self,
+        component: DataComponentType<T>,
+        default: T,
+    ) -> T {
+        self.get(component).unwrap_or(default)
+    }
+
+    /// Vanilla `patch.forget(implicitComponents::contains).split().added()`:
+    /// the stack's set components that no implicit read claimed.
```
coco875, line 54: I think it can be simplified

steel-core/src/block_entity/components.rs
```diff
@@ -0,0 +1,67 @@
...
+};
+use steel_registry::item_stack::ItemStack;
+use steel_utils::{DowncastType, Identifier};
+
+/// Component view handed to [`super::BlockEntity::apply_implicit_components`].
+///
+/// Mirrors the recording `DataComponentGetter` inside vanilla
+/// `BlockEntity.applyComponents`: reads resolve against the placed stack's
+/// effective components, and every type read is remembered so it stays out
+/// of the block entity's stored component map.
```
coco875, line 20: I think it can be simplified

steel-core/src/block_entity/mod.rs
```diff
@@ -428,6 +438,38 @@ pub trait BlockEntity: ErasedType + Send + Sync {
         self.base().load_components(nbt);
     }
 
+    /// Reads the components this entity represents through its own fields.
+    ///
+    /// Vanilla `BlockEntity.applyImplicitComponents`: every component read from
+    /// `components` is consumed and kept out of the stored component map.
```
coco875, line 444: I don't think precise the vanilla name a useful because it's very transparent

steel-core/src/block_entity/mod.rs
```diff
@@ -428,6 +438,38 @@ pub trait BlockEntity: ErasedType + Send + Sync {
...
+    fn apply_implicit_components(&self, _components: &mut ImplicitComponentGetter<'_>) {}
+
+    /// Adds the components this entity represents through its own fields.
+    ///
+    /// Vanilla `BlockEntity.collectImplicitComponents`; the inverse of
+    /// [`Self::apply_implicit_components`].
```
coco875, line 450: I don't think precise the vanilla name a useful because it's very transparent

steel-core/src/block_entity/mod.rs
```diff
@@ -428,6 +438,38 @@ pub trait BlockEntity: ErasedType + Send + Sync {
...
+    fn collect_implicit_components(&self, _components: &mut DataComponentMap) {}
+
+    /// Applies the placed stack's components: implicit ones through
+    /// [`Self::apply_implicit_components`], the rest into the stored component map.
+    ///
+    /// Vanilla `BlockEntity.applyComponentsFromItemStack`; final, do not override.
```
coco875, line 456: I don't think precise the vanilla name a useful because it's very transparent

steel-core/src/block_entity/mod.rs
```diff
@@ -428,6 +438,38 @@ pub trait BlockEntity: ErasedType + Send + Sync {
...
+    fn apply_components_from_item_stack(&self, stack: &ItemStack) {
+        let mut getter = ImplicitComponentGetter::new(stack);
+        self.apply_implicit_components(&mut getter);
+        self.base()
+            .set_components(getter.remaining_stored_components());
+    }
+
+    /// Returns the stored components together with the implicit ones.
+    ///
+    /// Vanilla `BlockEntity.collectComponents`; final, do not override.
```
coco875, line 466: I don't think precise the vanilla name a useful because it's very transparent

steel-core/src/enchantment_helper/selection.rs
```diff
@@ -0,0 +1,276 @@
...
+    if !stack.has(ENCHANTABLE) {
+        return 0;
+    }
+    let bookcases = bookcases.min(15);
+    let selected =
+        random.next_i32_bounded(8) + 1 + (bookcases >> 1) + random.next_i32_bounded(bookcases + 1);
+    match slot {
+        0 => (selected / 3).max(1),
+        1 => selected * 2 / 3 + 1,
+        _ => selected.max(bookcases * 2),
```
coco875, line 42: I see multiple magic number, so number that have a particular and important meaning for the code to be understand

steel-core/src/enchantment_helper/selection.rs
```diff
@@ -0,0 +1,276 @@
...
+fn filter_compatible_enchantments(
+    candidates: &mut Vec<EnchantmentInstance>,
+    target: &EnchantmentInstance,
+) {
+    candidates
+        .retain(|candidate| Enchantment::are_compatible(target.enchantment, candidate.enchantment));
+}
+
+#[cfg(test)]
+mod tests {
```
coco875, line 130: also magic number in test

steel-core/src/inventory/container/mod.rs
```diff
@@ -17,7 +17,8 @@ use steel_registry::blocks::properties::Direction;
 use steel_registry::item_stack::ItemStack;
 use steel_utils::ErasedType;
 
-/// Default distance buffer for container interaction range checks.
+/// Vanilla `Container.DEFAULT_DISTANCE_BUFFER`: the reach slack for container
+/// and menu `stillValid` checks.
```
coco875, line 21: I don't think precise the vanilla name a useful because it's very transparent

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,431 @@
...
+        EnchantingTableBlock::BOOKSHELF_OFFSETS
+            .iter()
+            .filter(|offset| {
+                EnchantingTableBlock::is_valid_book_shelf(&self.world, self.block_pos, **offset)
+            })
+            .count() as i32
+    }
+
+    /// Vanilla `EnchantmentMenu.getEnchantmentList`: reseeds `random` with
+    /// `seed + slot` and selects the offer; books drop one result when several were rolled.
```
coco875, line 173: I don't think precise the vanilla name a useful because it's very transparent

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,431 @@
...
+            let clue = list[random.next_i32_bounded(list.len() as i32) as usize];
+            self.enchant_clue[offer] = clue.enchantment.id() as i32;
+            self.level_clue[offer] = clue.level as i32;
+        }
+        self.sync_data(behavior);
+    }
+}
+
+impl MenuKind for EnchantmentKind {
+    /// Vanilla `stillValid`: the table must still stand there and be in reach.
```
coco875, line 246: I don't think precise the vanilla name a useful because it's very transparent

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,431 @@
...
+        self.world.get_block_state(self.block_pos).get_block() == &vanilla_blocks::ENCHANTING_TABLE
+            && player.is_within_block_interaction_range_with_buffer(
+                self.block_pos,
+                f64::from(DEFAULT_DISTANCE_BUFFER),
+            )
+    }
+
+    /// Vanilla `slotsChanged`: rerolls the offers only when `enchantSlots`
+    /// changed. Vanilla reaches this through that container's `setChanged`
+    /// override, so player-inventory clicks leave the offers alone.
```
coco875, line 257: I don't think precise the vanilla name a useful because it's very transparent

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,431 @@
...
+        guard: &mut ContainerLockGuard,
+        _player: &Player,
+    ) {
+        if self.table_changes(guard) == Some(self.table_changes_seen) {
+            return;
+        }
+        self.update_offers(behavior, guard);
+    }
+
+    /// Vanilla `clickMenuButton`: applies offer `button_id` when the player can pay.
```
coco875, line 270: I don't think precise the vanilla name a useful because it's very transparent

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,431 @@
...
+            log::debug!(
+                "{} pressed invalid button id: {button_id}",
+                player.gameprofile.name
+            );
+            return false;
+        };
+        let enchantment_cost = offer as i32 + 1;
+        let has_infinite_materials = player.has_infinite_materials();
+
+        // Only the table container takes part; the player inventory stays unlocked.
```
coco875, line 290: I don't think the comment are useful

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,431 @@
...
+        if self.costs[offer] <= 0
+            || stack.is_empty()
+            || ((level < enchantment_cost || level < self.costs[offer]) && !has_infinite_materials)
+        {
+            return false;
+        }
+
+        // The seed here is discarded immediately: `enchantment_list` reseeds `random`
+        // with `seed + offer` before using it, mirroring vanilla's reuse of one
+        // long-lived `random` field rather than seeding fresh for the roll itself.
```
coco875, line 314: I don't think it's useful

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,431 @@
...
+            let mut experience = player.experience.lock();
+            experience.on_enchantment_performed(enchantment_cost, rand::random());
+            self.enchantment_seed = experience.enchantment_seed();
+        }
+
+        // Vanilla `transmuteCopy`: a book keeps its count and components as an enchanted book.
+        // `with_count_and_patch` approximates it via `sanitize_against`, which drops patch
+        // entries equal to the new prototype's default, where vanilla's `forget` drops any
+        // entry the new item defines a default for regardless of value. The two differ only
+        // for patches unreachable through gameplay; identical for anything reachable in play.
```
coco875, line 331: I don't think it's useful

steel-core/src/inventory/menu/kinds/enchantment_menu.rs
```diff
@@ -0,0 +1,431 @@
...
+            self.block_pos,
+            1.0,
+            rand::random::<f32>() * 0.1 + 0.9,
+            None,
+        );
+        true
+    }
+
+    /// Vanilla `quickMoveStack`: table slots empty into the inventory, lapis goes
+    /// to its slot, and anything else moves a single item into the empty item slot.
```
coco875, line 371: I don't think precise the vanilla name a useful because it's very transparent

steel-core/src/player/player_inventory/player_handlers.rs
```diff
@@ -186,20 +186,31 @@ impl Player {
...
-        // - Stonecutter recipe selection
-        // - Loom pattern selection
-        // - Lectern page turning
+        self.reset_last_action_time();
+        match self.take_open_menu_for_callback(Some(packet.container_id)) {
+            Ok(mut menu) => {
+                if self.game_mode() == GameType::Spectator {
+                    // Vanilla `handleContainerButtonClick` silently drops spectator
+                    // clicks; the resync only happens in `handleContainerClick`.
+                } else if !menu.still_valid(self) {
```
coco875, line 198: does it's a todo else it can be simplified the if

steel-registry/build/enchantments/mod.rs
```diff
@@ -582,3 +574,34 @@ pub(crate) fn build() -> TokenStream {
 
     stream
 }
+
+/// Turns a `#namespace:path` tag reference into a static `Identifier` expression.
+///
+/// Vanilla models these fields as `HolderSet`s that may also list entries
+/// directly; the extracted vanilla data only uses tags, so anything else is a
+/// build error rather than a silently wrong type.
```
coco875, line 582: I don't think that compare to vanilla are useful

steel-registry/src/enchantment/mod.rs
```diff
@@ -112,10 +118,31 @@ impl Enchantment {
 
     /// Checks if this enchantment can be applied to the given item via `supported_items` tag.
     pub fn can_enchant(&self, item: ItemRef) -> bool {
-        let Some(tag) = parse_tag_ref(self.supported_items) else {
+        REGISTRY.items.is_in_tag(item, &self.supported_items)
+    }
+
+    /// Vanilla `Enchantment.getMinCost`.
```
coco875, line 124: I don't think precise the vanilla name a useful because it's very transparent

steel-registry/src/enchantment/mod.rs
```diff
@@ -112,10 +118,31 @@ impl Enchantment {
...
+    #[must_use]
+    pub const fn get_min_cost(&self, level: i32) -> i32 {
+        self.min_cost.calculate(level)
+    }
+
+    /// Vanilla `Enchantment.getMaxCost`.
```
coco875, line 130: I don't think precise the vanilla name a useful because it's very transparent

steel-registry/src/enchantment/mod.rs
```diff
@@ -112,10 +118,31 @@ impl Enchantment {
...
+    #[must_use]
+    pub const fn get_max_cost(&self, level: i32) -> i32 {
+        self.max_cost.calculate(level)
+    }
+
+    /// Vanilla `Enchantment.isPrimaryItem`: supported, and inside `primary_items`
+    /// when that narrower tag is set.
```
coco875, line 137: I don't think precise the vanilla name a useful because it's very transparent

steel-utils/src/java.rs
```diff
@@ -149,9 +149,46 @@ fn floating_to_string(
     output
 }
 
+/// `jdk.internal.math.FloatConsts`, derived the way the JDK derives them.
+mod float_consts {
```
coco875, line 153: seems a bit weird that we don't need it until now, there is specific reason ?

coco875 (changes requested):
I have a done a first batch of comment on code quality, I still need to check if it's match with vanilla

PR #626, opened 2026-09-07
add vanilla tag command

SoullessGhosty:
## Type of change

- [x] Command implementation
- [x] New feature

## Description

This PR adds the vanilla `/tag` command to SteelMC.

The command supports adding, removing, and listing entity scoreboard tags using SteelMC's existing entity tag API and command framework.

Supported commands:

- `/tag <targets> add <name>`
- `/tag <targets> remove <name>`
- `/tag <targets> list`

The implementation supports both single and multiple entity targets and provides appropriate vanilla-style success and failure feedback.

It also handles cases where:
- An entity already has the specified tag.
- A tag cannot be added.
- An entity does not have the specified tag.
- An entity has no tags.
- Multiple entities are targeted.

A command-tree test was added to verify that the `/tag` command has the expected vanilla argument structure and argument types.

## How this was tested

The command was tested using:

/tag @s add test
/tag @s list
/tag @s remove test


steel-core/src/command/builtins/tag.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+fn list_tags(context: &SteelCommandContext<CommandSource>) -> Result<i32, CommandSyntaxError> {
+    let targets = context.entities("targets")?;
+
+    if let [target] = targets.as_slice() {
+        let tags = target.tags();
+
+        if tags.is_empty() {
+            return Err(CommandSyntaxError::dynamic(
```
Joniii11: vanilla returns success with result 0 here not an error alos /tag ... list uses broadcastToAdmins=false

steel-core/src/command/builtins/tag.rs
```diff
@@ -0,0 +1,241 @@
...
+pub(super) fn registration() -> CommandRegistration<CommandSource> {
+    CommandRegistration::new(Identifier::vanilla_static("tag"), |_| command())
+}
+
+fn command() -> CommandNodeBuilder<CommandSource, SteelCommandRuntime> {
+    literal("tag").then(
+        argument("targets", SteelArgumentType::entities())
+            .then(literal("add").then(argument("name", ArgumentType::word()).executes(add_tag)))
+            .then(
+                literal("remove").then(argument("name", ArgumentType::word()).executes(remove_tag)),
```
Joniii11: in game adds server side suggestions here (minecraft:ask_server) for the selected entity tags which we dont do here

steel-core/src/command/builtins/tag.rs
```diff
@@ -0,0 +1,241 @@
...
+            .message([
+                TextComponent::plain(tag.to_owned()),
+                TextComponent::plain(target.plain_text_name()),
+            ])
+            .component()
+    } else {
+        translations::COMMANDS_TAG_ADD_SUCCESS_MULTIPLE
+            .message([
+                TextComponent::plain(tag.to_owned()),
+                TextComponent::plain(changed.to_string()),
```
Joniii11: in vanilla shows targets.size() in this message while only the command result uses the changed count

steel-core/src/command/builtins/tag.rs
```diff
@@ -0,0 +1,241 @@
...
+        return Err(CommandSyntaxError::dynamic(TextComponent::from(
+            &translations::COMMANDS_TAG_ADD_FAILED,
+        )));
+    }
+
+    let message = if let [target] = targets.as_slice() {
+        translations::COMMANDS_TAG_ADD_SUCCESS_SINGLE
+            .message([
+                TextComponent::plain(tag.to_owned()),
+                TextComponent::plain(target.plain_text_name()),
```
Joniii11: in vanilla they use the rich entity display component here not a flattened plaintext name

steel-core/src/command/builtins/tag.rs
```diff
@@ -0,0 +1,241 @@
...
+            ])
+            .component()
+    } else {
+        translations::COMMANDS_TAG_REMOVE_SUCCESS_MULTIPLE
+            .message([
+                TextComponent::plain(tag.to_owned()),
+                TextComponent::plain(changed.to_string()),
```
Joniii11: in vanilla shows targets.size() in this message while only the command result uses the changed count

steel-core/src/command/builtins/tag.rs
```diff
@@ -0,0 +1,241 @@
...
+                    .message([TextComponent::plain(target.plain_text_name())])
+                    .component(),
+            ));
+        }
+
+        let message = translations::COMMANDS_TAG_LIST_SINGLE_SUCCESS
+            .message([
+                TextComponent::plain(target.plain_text_name()),
+                TextComponent::plain(tags.len().to_string()),
+                TextComponent::plain(tags.join(", ")),
```
Joniii11: vanilla uses ComponentUtils.formatList() here which keeps the tags styled instead of joining them into one plain text component

Joniii11 (changes requested):

steel-core/src/command/builtins/tag.rs
```diff
@@ -0,0 +1,232 @@
...
+                .collect::<BTreeSet<_>>();
+
+            for tag in tags {
+                if tag.to_lowercase().starts_with(&prefix) {
+                    builder.suggest(tag);
+                }
+            }
+        },
+    )
+}
```
Laptop59: 1. `add` doesn't suggest tags.
2. This is not how you should suggest preexisting tags; instead you should use `argument_with_suggestions` instead of `argument` to add the things you want to suggest.

steel-core/src/command/builtins/tag.rs
```diff
@@ -0,0 +1,232 @@
...
+    let changed = targets
+        .iter()
+        .filter(|target| target.remove_tag(tag))
+        .count();
+
+    if changed == 0 {
+        return Err(CommandSyntaxError::dynamic(TextComponent::from(
+            &translations::COMMANDS_TAG_REMOVE_FAILED,
+        )));
+    }
```
Laptop59: Do something similar to what I said in my review in `add_tag` here.

steel-core/src/command/builtins/tag.rs
```diff
@@ -0,0 +1,232 @@
...
+    for (index, tag) in tags.iter().enumerate() {
+        if index > 0 {
+            list = list.add_child(TextComponent::plain(", "));
+        }
+
+        list = list.add_child(TextComponent::plain(tag.to_owned()));
+    }
+
+    list
+}
```
Laptop59: It would be cool to add Vanilla's text component joining utilities to some utils but I guess for this PR it's fine and it can be done in another PR because other commands just do it their own way

steel-core/src/command.zip
Laptop59, line 1: Untrack this file from git or delete it. Anyways, if you want to allow others to look at your source code, you can just refer them to your GitHub branches in your fork.

steel-core/src/entity.zip
Laptop59, line 1: Same thing here for this file.

steel-core/src/command/builtins/tag.rs
```diff
@@ -0,0 +1,232 @@
...
+    let changed = targets
+        .iter()
+        .filter(|target| target.add_tag(tag.to_owned()))
+        .count();
+
+    if changed == 0 {
+        return Err(CommandSyntaxError::dynamic(TextComponent::from(
+            &translations::COMMANDS_TAG_ADD_FAILED,
+        )));
+    }
```
Laptop59: Maybe instead of counting and then checking `== 0`, you can use something like
```rs
let changed = targets
    .iter()
    .any(|target| target.add_tag(tag.to_owned()));
```
then
```rs
if !changed {
    return Err(CommandSyntaxError::dynamic(TextComponent::from(
        &translations::COMMANDS_TAG_ADD_FAILED,
    )));
}
```

Laptop59 (changes requested):
Here is my review:

steel-core/src/command/builtins/tag.rs
SomeYellowGuy: Actually, it's better to make a new type of argument and parser here instead of using `argument_with_suggestions`. You can refer to other parsers like `TimelineParser` and other parsers that use `REGISTRY` to make your own implementation. You could also refer to https://steelmc.dev/development/commands/02-arguments/.

SoullessGhosty: Should be Done

steel-core/src/command.zip
SoullessGhosty: Done

steel-core/src/command/builtins/tag.rs
SoullessGhosty: Should be Done i think?

steel-core/src/command/execution/argument.rs
```diff
@@ -762,6 +771,85 @@ unit_argument_parser!(
     protocol(ProtocolArgumentType::EntityAnchor, None)
 );
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct WordParser;
+
+impl_downcast_type!(WordParser, "steel:command/parser/word");
+
+impl SteelArgumentParser for WordParser {
```
SomeYellowGuy: I don't think there's a need for another word argument. There is already a built-in Brigadier one. Use `ArgumentType::word()`?

SomeYellowGuy (changes requested):

steel-core/src/command/execution/argument.rs
SoullessGhosty: Changed.

SomeYellowGuy (approved):
Looks good, but still not sure why the degree-related diffs are there

steel-core/src/command/builtins/tag.rs
SoullessGhosty: Done

steel-core/src/command/builtins/tag.rs
SoullessGhosty: > vanilla uses ComponentUtils.formatList() here which keeps the tags styled instead of joining them into one plain text component

done

steel-core/src/command/builtins/teleport.rs
```diff
@@ -415,6 +414,15 @@ fn wrap_rotation((yaw, pitch): (f32, f32)) -> (f32, f32) {
...
 }
 
+fn wrap_degrees(value: f32) -> f32 {
+    let wrapped = value.rem_euclid(360.0);
+    if wrapped >= 180.0 {
+        wrapped - 360.0
+    } else {
+        wrapped
+    }
+}
```
coco875, line 424: why remove the import and duplicate the function here ?

steel-core/src/command/execution/selector/parser.rs
```diff
@@ -703,6 +703,17 @@ fn parse_range<T: Copy, E>(
...
+pub(super) fn wrap_degrees(value: f32) -> f32 {
+    let mut value = value % 360.0;
+    if value >= 180.0 {
+        value -= 360.0;
+    }
+    if value < -180.0 {
+        value += 360.0;
+    }
+    value
+}
```
coco875, line 715: steel_math::wrap_degrees already have it, look like you duplicate a bit

steel-core/src/command/execution/coordinates.rs
```diff
@@ -134,9 +136,10 @@ impl LocalCoordinates {
     }
 
     fn position(self, anchor: DVec3, (yaw, pitch): (f32, f32)) -> DVec3 {
-        let y_rotation = (yaw + DEGREE_90) * DEG_TO_RAD;
-        let x_rotation = -pitch * DEG_TO_RAD;
-        let x_up_rotation = (-pitch + DEGREE_90) * DEG_TO_RAD;
+        let radians_per_degree = PI / 180.0;
+        let y_rotation = (yaw + 90.0) * radians_per_degree;
+        let x_rotation = -pitch * radians_per_degree;
+        let x_up_rotation = (-pitch + 90.0) * radians_per_degree;
```
coco875, line 142: why revert it ???

steel-core/src/command/execution/source.rs
```diff
@@ -858,8 +872,16 @@ fn admin_broadcast_source_name(
...
+    yaw = yaw.rem_euclid(360.0);
+    if yaw >= 180.0 {
+        yaw -= 360.0;
+    }
+    pitch = pitch.rem_euclid(360.0);
+    if pitch >= 180.0 {
+        pitch -= 360.0;
+    }
+    (yaw, pitch)
 }
```
coco875, line 885: why revert it ?

coco875 (changes requested):
seems like that some merge conflict gone wrong

PR #629, opened 2026-09-07
feat(entity): add the Ender Dragon

Purdze:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [x] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

the ender dragon .. it flies, it can be hit, and it dies

- the entity and its 8 parts .. head, neck, body, 3 tails, 2 wings, each with its own
  hitbox, positioned every tick off the flight history
- flight history .. 64-entry ring of `(y, yRot)` the parts trail behind, `get(delay)`
  indexes by mask not modulo, java's negative-int `&` wraps and a rust `%` wouldn't
- phase machine .. all 11 phases exist, 3 are real (hovering, holding pattern, dying)
  and the rest are trait defaults so an unimplemented one degrades to "stops steering"
  instead of a panic
- flight graph .. the 24 nodes on 3 rings and a dragon-local a* over vanilla's
  adjacency table
- per-part damage .. head takes full, everything else `dmg/4 + min(dmg, 1)`, and it
  needs a player or `ALWAYS_HURTS_ENDER_DRAGONS`
- contact sweeps .. wings knock you back, head and neck deal 10
- `checkWalls` .. carves through terrain with `mobGriefing` on, respects
  `DRAGON_IMMUNE` and `DRAGON_TRANSPARENT`
- death .. killing blow sets health back to 1 and enters dying, it flies to the podium
  and only then drops to 0, which is what starts the 200-tick sequence and the xp

things worth knowing

- `Mth.sin`/`Mth.cos` are a 65536-entry table, not `f32::sin`.. used
  `steel_math::trig` throughout, it's load-bearing for the node ring since node 0's
  angle is exactly `-2π` and table error decides whether `floor` gives 60 or 59
- vanilla does the part geometry in `float` then widens on assignment, doing it in
  `f64` lands on a different table entry.. matched the widening points
- a phase can call `set_phase` from inside its own tick, so every phase method takes
  `&self` and no lock is ever held across a phase call
- didn't reuse `NodeStore`/`NodeHeap`.. the dragon addresses nodes by index while
  those are keyed by coordinate hash, and `create_hash` packs `y & 0xff` so a player
  at y=329 would collide with a graph node at y=73

not in this pr .. no `EnderDragonFight`, so no boss bar, no arena lifecycle, no
crystal respawn. the dragon runs vanilla's legal `dragonFight == null` path and is
spawnable with `/summon`. the other 8 phases and the fight follow separately

a summoned dragon stays in hovering, `HOLDING_PATTERN` is only entered from
`EnderDragonFight` or from phases that are still placeholders.. vanilla does the same
thing, so the holding pattern isn't reachable in a real session until the fight lands

## How this was tested

- `cargo test --workspace` .. all pass, 3449 total
- 51 new tests .. node positions against the ring formulas and pairwise distinct, a
  path for every start/end pair, `find_path(i, i)` is one node, heap pop order is
  non-decreasing in `f` and takes the right child on a tie, part offsets against a
  filled flight history, `head_y_offset` is -1 while sitting, `check_walls` breaks a
  plain block with the rule on and reports a wall hit with it off, never touches
  bedrock, ignores fire, a killing blow leaves it at 1 hp in dying, a perched dragon
  skips the death flight, 200 ticks remove it, `DragonDeathTime` round-trips mid-animation
- a phase that calls `set_phase` from inside its own tick completes without
  deadlocking, that's the one that would catch the re-entrancy bug
- `cargo clippy --workspace --all-targets --all-features` .. zero warnings
- `cargo fmt --all --check` .. clean

in-game: not properly tested yet.. it spawns and hovers, i probed the holding pattern
by ticking one 100 times and it flies out to the radius-40 ring, but the damage
numbers, wing knockback and the death flight all still want a real look

## Checklist

- [x] Self-reviewed the diff

- `EnderDragon`
- `EnderDragonPart`
- `DragonFlightHistory`
- `EnderDragonPhaseManager`
- `DragonHoverPhase`
- `DragonHoldingPatternPhase`
- `DragonDeathPhase`

## Additional notes

stacked on #625 and #628, please merge those first..

the fight, the remaining phases and the crystal chain follow after


steel-core/src/entity/entities/mobs/bosses/ender_dragon/phases/holding_pattern.rs
```diff
@@ -0,0 +1,218 @@
...
+        let path_is_done = self.path_is_done();
+
+        if path_is_done == Some(true) && Self::try_hand_off(dragon, world) {
+            return;
+        }
+
+        // Vanilla's `currentPath == null || currentPath.isDone()`.
+        if path_is_done.is_none_or(|done| done) {
+            self.pick_new_path(dragon, world);
+        }
```
Laptop59: Comment feels a bit unnecessary because the `is_none_or` feels self-explanatory

steel-core/src/entity/entities/mobs/bosses/ender_dragon/phases/hover.rs
```diff
@@ -0,0 +1,66 @@
...
+
+impl DragonPhaseInstance for DragonHoverPhase {
+    fn phase(&self) -> EnderDragonPhase {
+        EnderDragonPhase::Hovering
+    }
+
+    /// Vanilla reports hovering as sitting.
+    ///
+    /// This is not just bookkeeping: it selects the perched wing-beat rate and the
+    /// lowered head offset, so a hovering dragon animates like a perched one.
```
Laptop59: Why is there a doc comment in a trait implementation?

steel-core/src/entity/entities/mobs/bosses/ender_dragon/tests/death.rs
```diff
@@ -0,0 +1,265 @@
...
+
+fn experience_orbs_near(world: &Arc<World>, position: DVec3) -> usize {
+    let area = WorldAabb::new(
+        position.x - 64.0,
+        position.y - 64.0,
+        position.z - 64.0,
+        position.x + 64.0,
+        position.y + 64.0,
+        position.z + 64.0,
+    );
```
Laptop59: `64.0` looks like a magic constant

steel-core/src/entity/entities/mobs/bosses/ender_dragon/tests/flight_graph.rs
```diff
@@ -0,0 +1,186 @@
...
+
+#[test]
+fn a_path_exists_between_every_pair_of_nodes() {
+    let graph = test_graph();
+
+    for start in 0..NODE_COUNT {
+        for end in 0..NODE_COUNT {
+            let path = graph
+                .find_path(start, end, None, WITH_CRYSTALS)
+                .unwrap_or_else(|| panic!("no path from node {start} to node {end}"));
```
Laptop59: `expect` would be more well-suited here

steel-core/src/entity/entities/mobs/bosses/ender_dragon/tests/flight_graph.rs
```diff
@@ -0,0 +1,186 @@
...
+            let path = graph
+                .find_path(start, end, None, NO_FIGHT)
+                .unwrap_or_else(|| panic!("no restricted path from {start} to {end}"));
+
+            for index in 0..path.node_count() {
+                let node = path.node(index).expect("index is within the path");
+                let radius = f64::from(node.x).hypot(f64::from(node.z));
+                assert!(
+                    radius < 50.0,
+                    "the restricted path from {start} to {end} used a radius-60 node"
```
Laptop59: `radius-60` sounds like a radioactive element, which is kinda funny. Maybe just say `outer ring node` instead?

steel-core/src/entity/entities/mobs/bosses/ender_dragon/tests/flight_graph.rs
```diff
@@ -0,0 +1,186 @@
...
+
+#[test]
+fn a_restricted_path_never_routes_through_the_outer_ring() {
+    let graph = test_graph();
+
+    for start in OUTER_RING_NODES..NODE_COUNT {
+        for end in OUTER_RING_NODES..NODE_COUNT {
+            let path = graph
+                .find_path(start, end, None, NO_FIGHT)
+                .unwrap_or_else(|| panic!("no restricted path from {start} to {end}"));
```
Laptop59: Again, `expect`

steel-core/src/entity/entities/mobs/bosses/ender_dragon/flight_graph.rs
```diff
@@ -0,0 +1,497 @@
...
+}
+
+impl DragonFlightGraph {
+    /// Samples the terrain and lays out the three rings.
+    ///
+    /// The rings are centered on the **world origin**, not on the dragon's fight
+    /// origin. Every other position the dragon works with is fight-origin-relative, so
+    /// this one is worth flagging rather than "fixing".
+    #[must_use]
+    pub fn build(world: &World) -> Self {
```
darkstarworks: > Unloaded chunks silently bake a bad flight graph

`build` samples the heightmap for all 24 nodes and the graph is then cached for the entity's lifetime via `get_or_insert_with`, matching vanilla's `nodes[0] == null` guard. The difference is underneath: vanilla's `Level.getHeightmapPos` blocking-loads the chunk, while Steel's `level_height_at` falls back to `get_min_y()` on a miss, so a graph built while the origin chunks are cold bakes in a flat y=73 ring permanently. 

Same fallback also hits `podium_target` and the egg position in `try_hand_off`.

steel-core/src/entity/entities/mobs/bosses/ender_dragon/flight_graph.rs
```diff
@@ -0,0 +1,497 @@
...
+                } else {
+                    (20.0_f32, 5, ring_angle(FRAC_PI_4, index - INNER_RING_START))
+                };
+
+                // `trig` is a port of `Mth.cos`/`Mth.sin`, a 65536-entry lookup table
+                // rather than `f32::cos`. That distinction decides `floor` outright at
+                // each ring's axis points, where the product lands on an integer.
+                let x = (radius * trig::cos(f64::from(angle))).floor() as i32;
+                let z = (radius * trig::sin(f64::from(angle))).floor() as i32;
+                let surface = world.heightmap_pos(
```
darkstarworks: [`flight_graph.rs:254`](https://github.com/Steel-Foundation/SteelMC/pull/629#discussion_r3959900957)

steel-core/src/entity/entities/mobs/bosses/ender_dragon/mod.rs
```diff
@@ -0,0 +1,1269 @@
...
+        self.update_dirty_mob_effect_entity_data();
+    }
+
+    /// Mirrors vanilla `EnderDragon.isFlapping`.
+    ///
+    /// Detects the point in the beat where the wings sweep down, which is what makes
+    /// the shared move path emit the flap game event.
+    fn is_flapping(&self) -> bool {
+        let flap = (*self.flap_time.lock() * TAU).cos();
+        let previous = (*self.o_flap_time.lock() * TAU).cos();
```
darkstarworks: `is_flapping` bypasses the trig table because `Mth.cos` here is the 65536-entry table, and the rest of this PR is careful to route through `steel_math::trig` for exactly that reason. `(flap_time * TAU).cos()` is `f32::cos`, so near the `-0.3` threshold the FLAP event can land a tick either side of vanilla

steel-core/src/entity/entities/mobs/bosses/ender_dragon/mod.rs
```diff
@@ -0,0 +1,1269 @@
...
+    ///
+    /// A full replacement for the shared living step, as vanilla's is: the dragon
+    /// steers from its phase rather than from movement input, and it shoves entities
+    /// with its wings instead of the usual entity pushing.
+    fn dragon_ai_step(&self, world: &Arc<World>) -> Option<MoveResult> {
+        self.process_flapping_movement();
+
+        // Everything below is inside vanilla's `else`; a dying dragon is moved by
+        // `tick_death` instead.
+        if self.is_dead_or_dying() {
```
darkstarworks, line 449: Vanilla sets `oFlapTime = flapTime` at the top of `aiStep`, above the `isDeadOrDying` split, so it runs every tick. Here it sits in `tick_flap_time`, below the early return, so it stops updating once the dragon is dying. Is `ai_step` still reached on the first dying tick? If so this could emit one stray flap (also on :814)

steel-core/src/entity/entities/mobs/bosses/ender_dragon/mod.rs
```diff
@@ -0,0 +1,1269 @@
...
+    /// not break, which is what stalls it; blocks that *were* broken are reported to
+    /// clients as particles instead.
+    fn check_walls(world: &Arc<World>, box_: WorldAabb) -> bool {
+        let min = BlockPos::containing(box_.min_x(), box_.min_y(), box_.min_z());
+        let max = BlockPos::containing(box_.max_x(), box_.max_y(), box_.max_z());
+
+        // Vanilla re-reads the game rule for every block. That takes a level-data read
+        // lock, and the body box alone spans dozens of blocks each tick, so the
+        // loop-invariant read is hoisted; the result is identical either way.
+        let griefing = world.get_game_rule(&MOB_GRIEFING);
```
darkstarworks, line 810: When `mob_griefing` is off, `destroyed` can never be set and the particle event is gated on it, so once `hit_wall` is true the rest of the scan cannot change the result or consume any RNG. An early return would save most of ~216 block reads per tick on servers running the rule off

steel-core/src/entity/entities/mobs/bosses/ender_dragon/mod.rs
```diff
@@ -0,0 +1,1269 @@
...
+                    f64::from((cc1 * 1.5 + cc * distance) * cc_tilt),
+                ),
+            );
+        }
+    }
+
+    /// Moves one hitbox to an offset from the dragon. Mirrors vanilla `tickPart`.
+    fn tick_part(&self, index: usize, offset: DVec3) {
+        self.sub_entities[index].set_part_position(self.position() + offset);
+    }
```
darkstarworks: `tick_part` calls `self.position()` on each of its 8 invocations, and `flight_history` is locked five times a tick (once here, again in `head_y_offset`, then once per tail iteration). 
Hoisting both would take a fair bit of lock traffic out of the hot path (also :580)

steel-core/src/entity/entities/mobs/bosses/ender_dragon/mod.rs
```diff
@@ -0,0 +1,1269 @@
...
+        self.with_flight_graph(world, |graph| {
+            graph.find_path(start, end, final_node, crystals)
+        })
+    }
+
+    /// Runs `action` against the flight graph, building it if this is the first use.
+    ///
+    /// `action` runs under the graph lock, so it must stay confined to the graph. The
+    /// two callers above satisfy that: a search reads node positions and nothing else.
+    fn with_flight_graph<R>(
```
darkstarworks, line 419: maybe lock chain under `with_flight_graph`.
`pick_new_path` holds the phase `state` lock across `find_path`, which takes the graph lock and, on first use, runs `build` with its 24 heightmap lookups. 
The doc says the action must stay confined to the graph, which holds for the searches, but `build` itself reaches into the world under that lock. ([holding_pattern.rs:150](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-c9b3a9478c71c44e6b7ca3c84decf75b321f93d07fd7d716c07fa3ad3dfe2140R150) & [holding_pattern.rs:168](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-c9b3a9478c71c44e6b7ca3c84decf75b321f93d07fd7d716c07fa3ad3dfe2140R168))

steel-core/src/entity/entities/mobs/bosses/ender_dragon/flight_graph.rs
```diff
@@ -0,0 +1,497 @@
...
+                Node::new(x, MINIMUM_NODE_Y.max(surface.y() + y_adjustment), z)
+            }),
+        }
+    }
+
+    /// Returns a node's position.
+    #[must_use]
+    pub const fn node(&self, index: usize) -> &Node {
+        &self.nodes[index]
+    }
```
darkstarworks, line 307: `node()` and `find_path` are `pub` and index `self.nodes` directly. Every in-tree caller is provably in range, but if third-party phases ever land (per the `Custom(u16)` note on the enum) these would want `get()` and an `Option` (also [:327](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-3046f26c4074bc723799341eec8e91a01f14fea470111d576021902916dbd37aR327))

steel-core/src/entity/entities/mobs/bosses/ender_dragon/mod.rs
```diff
@@ -0,0 +1,1269 @@
...
+
+        let (sample_5, sample_10) = {
+            let history = self.flight_history.lock();
+            (history.get(5), history.get(10))
+        };
+
+        // Every expression below is `f32` up to the point vanilla widens it, because
+        // the trig lookup indexes off the narrowed value. Doing the arithmetic in `f64`
+        // and narrowing afterwards would land on a different table entry.
+        let tilt = ((sample_5.y - sample_10.y) as f32 * 10.0).to_radians();
```
darkstarworks, line 550: The semantic constants are named well throughout, but the steering and part-geometry formulas keep vanilla's bare literals (part reach `4.5`/`6.5`/`5.5`/`1.5`, `dy * 0.01`, `turn * 0.1`, the `0.8 + 0.15` slide, `-0.3`, `0.7 / 40.0`).
Only flagging since it usually comes up, though a dense physics kernel is arguably more readable with them inline
(mod.rs:512, [521](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-f4461d32b18ff52d2aa8e6fb6afefcdc547772de85295a7ee297fea224e26d95R521), [525](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-f4461d32b18ff52d2aa8e6fb6afefcdc547772de85295a7ee297fea224e26d95R525), [537](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-f4461d32b18ff52d2aa8e6fb6afefcdc547772de85295a7ee297fea224e26d95R537), [541](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-f4461d32b18ff52d2aa8e6fb6afefcdc547772de85295a7ee297fea224e26d95R541), [591](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-f4461d32b18ff52d2aa8e6fb6afefcdc547772de85295a7ee297fea224e26d95R591), [818](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-f4461d32b18ff52d2aa8e6fb6afefcdc547772de85295a7ee297fea224e26d95R818), [873](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-f4461d32b18ff52d2aa8e6fb6afefcdc547772de85295a7ee297fea224e26d95R873), [884](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-f4461d32b18ff52d2aa8e6fb6afefcdc547772de85295a7ee297fea224e26d95R884), [897](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-f4461d32b18ff52d2aa8e6fb6afefcdc547772de85295a7ee297fea224e26d95R897), [900](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-f4461d32b18ff52d2aa8e6fb6afefcdc547772de85295a7ee297fea224e26d95R900), [915](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-f4461d32b18ff52d2aa8e6fb6afefcdc547772de85295a7ee297fea224e26d95R915), and [phases/mod.rs:131](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-2172fb6553c75c36b4502f149b640b4d4e3b80c0c0b7b72bedbd363a6e5744ccR131), [138](https://github.com/Steel-Foundation/SteelMC/pull/629/changes#diff-2172fb6553c75c36b4502f149b640b4d4e3b80c0c0b7b72bedbd363a6e5744ccR138)

steel-core/src/entity/entities/mobs/bosses/ender_dragon/phases/holding_pattern.rs
```diff
@@ -0,0 +1,218 @@
...
+            return;
+        };
+        state.target_location = Some(target);
+    }
+
+    /// Rolls for leaving the circle, and returns whether the dragon switched phase.
+    ///
+    /// Vanilla runs this only once a path has been walked to its end, so the dragon
+    /// commits to a full hop between nodes before it can peel off.
+    fn try_hand_off(dragon: &EnderDragonEntity, world: &Arc<World>) -> bool {
```
darkstarworks, line 97: `HOLDING_PATTERN` is only entered from `EnderDragonFight`, so `try_hand_off` is both unreachable and untested today. It is the branchiest code in the phase file and could be driven directly by a test cheaply now

steel-core/src/entity/entities/mobs/bosses/ender_dragon/mod.rs
Purdze: `pick_new_path` calls `find_closest_node` at `:148`, above the state lock at `:150`, and that already builds the graph through the same `with_flight_graph` so `build` never runs under the phase lock, only tests call these otherwise

PR #633, opened 2026-09-08
feat: give the fox its eat-berries goal

darkstarworks:
## Type of change

- [x] Entity implementation

## Description

Implements the vanilla fox `FoxEatBerriesGoal`.

A fox spotting a ripe sweet berry bush, or glow berries vine, walks over and checks it out for a couple of seconds, then takes what it finds.

* Adds `FoxEatBerriesGoal` at priority 10, composing the shared `MoveToBlockGoal` with the fox's own overrides: an accepted distance of `2.0`, a fresh path every 100 try-ticks, and a target test of a sweet berry bush at age 2 or better, or any block carrying glow berries
* Waits 40 ticks at the bush before interacting, with `0.05` per-tick `FOX_SNIFF` on the way over
* Picks one berry into the mouth if not already carrying something, dropping the rest beside the bush. A fully grown bush gives 1 extra berry. Bush is left at age 1 so it grows back rather than being destroyed
* Picking glow berries goes through `CaveVinesBlock::use_block`, which both cave vine blocks already share, so the drops, the sound and the emptied vine come from one place. The glow berry check moved onto the block as `CaveVinesBlock::has_glow_berries`, shared by the block and the goal
* Pick gated on `mobGriefing`, matching vanilla
* Exports `MoveToBlockGoal` from the shared goal module. (was unused, already written and tested.)

One vanilla line is deliberately not carried over: `pickSweetBerries` computes picked `BlockState` and drops the result, then sets the block properly a few lines later. Reproducing that would only add a dead assignment.

## How this was tested

Through `can_use`, `start` and `tick`, so the block search, the wait and the effect are all covered rather than just the picking:

* Which blocks a fox will walk to: Age 1 bush is ignored, ages 2 and 3 are not. Bare vine is ignored and a berried one is not. Block without berries is never a target
* A fox waiting out the vanilla pause before it takes anything, keeping one, dropping the rest, and leaving the bush at its picked age
* A fox already carrying something, dropping the whole picking instead of swapping it out
* With `mobGriefing` off, the bush untouched, nothing dropped, and the fox still waiting at it rather than giving up, matching vanilla
* A vine left bare after a fox strips its glow berries

The berries sit five blocks from the fox on a stone floor, with the fox's navigation and movement running alongside the goal, so the walk and the path updates are covered as well.

'The sniff on the way over' is a sound, so no server-side test.

`cargo fmt --all --check`, `RUSTFLAGS="-Dwarnings" cargo clippy -r --all-targets --all-features`, `typos` and `RUSTFLAGS="-Dwarnings" cargo test` all pass.

## Screenshots / logs

Not applicable. Server-side goal logic is covered by unit tests above.

## Classes modified:

Fox

## Additional notes

A few things to check during review:

* **Builds on #513.** This needs the fox entity core, and there is no `fox-entity` branch here to target, so it is opened against `master` and the diff shown includes #513's commits. This PR's own change is the commit after #513's. A rebase drops the overlap once #513 merges.
* **The goal only ever takes one helping.** `onReachedTarget` runs on every tick once the wait is over, but picking leaves the bush below the age the goal searches for and strips a vine of its berries, so the shared goal stops on its next check. That is vanilla's own arrangement rather than anything added here.
* **With `mobGriefing` off a fox will stand at the bush for a while.** The shared goal counts down against its stay ticks rather than ending, so the fox waits there for up to a minute or three before wandering off. Vanilla behaves the same way, and the test pins it, but it is the kind of thing that looks like a bug in game if you are not expecting it.
* **Overlaps with #604.** That PR adds a test pinning the fox's whole goal-priority list, and this PR adds an entry at priority 10, so whichever merges second needs one line of that list updated. Flagging it so it is not a surprise.
* **Search ranges live as constants rather than constructor arguments.** Vanilla passes `12` and `1` at the registration site; only the fox uses this goal, so they sit beside the other tuning values instead, the way the turtle's goals do.


steel-core/src/entity/entities/mobs/passive/fox/goals.rs
```diff
@@ -0,0 +1,624 @@
...
+        let age = state.get_value(SWEET_BERRY_AGE);
+        let mut count =
+            rand::random_range(BERRIES_PER_PICK) + i32::from(age == SWEET_BERRY_MAX_AGE);
+
+        let mut mouth_is_empty = false;
+        fox.with_equipment_slot(EquipmentSlot::MainHand, &mut |item_stack| {
+            mouth_is_empty = item_stack.is_empty();
+        });
+        if mouth_is_empty {
+            fox.living_base().equipment().lock().set(
```
Joniii11, line 555: vanilla calls setitemslot here which also calls onequipitem raw equipment set skips this whole hook

steel-core/src/entity/entities/mobs/passive/fox/goals.rs
```diff
@@ -0,0 +1,624 @@
...
+pub(super) const BERRY_WAIT_TICKS: i32 = 40;
+const BERRY_SNIFF_CHANCE: f32 = 0.05;
+const BERRIES_PER_PICK: RangeInclusive<i32> = 1..=2;
+const SWEET_BERRY_RIPE_AGE: u8 = 2;
+const SWEET_BERRY_MAX_AGE: u8 = 3;
+const SWEET_BERRY_PICKED_AGE: u8 = 1;
+const SWEET_BERRY_AGE: &IntProperty = &BlockStateProperties::AGE_3;
+const BERRIES: &BoolProperty = &BlockStateProperties::BERRIES;
+
+fn has_glow_berries(state: BlockStateId) -> bool {
```
Joniii11, line 53: why duplicate this here cause mc uses cavevines.hasglowberries in both places imo this should live on cavevinesblock too

steel-core/src/entity/entities/mobs/passive/fox/tests.rs
```diff
@@ -0,0 +1,857 @@
...
+        if !goal.can_continue_to_use(fox.as_ref()) {
+            return Some(tick);
+        }
+        goal.tick(fox.as_ref());
+    }
+    None
+}
+
+const BERRY_GOAL_TICKS: u32 = 200;
+const BERRY_BLOCK_POS: BlockPos = BlockPos::new(8, 64, 8);
```
Joniii11, line 727: this makes the move target exactly the fox position so the tests never really test walking/repath maybe put berries few blocks away

Joniii11 (changes requested):
else is good

steel-core/src/entity/entities/mobs/passive/fox/goals.rs
darkstarworks: Right, Steel had no `setItemSlot` hook at all, so I added `set_item_slot` / `on_equip_item` in its own PR: #695. The berry pickup and the fox's other mouth-item writes switch over once it lands.

steel-core/src/entity/entities/mobs/passive/fox/goals.rs
darkstarworks: done, it's `CaveVinesBlock::has_glow_berries` now and the block's own use path reads it too

steel-core/src/entity/entities/mobs/passive/fox/tests.rs
darkstarworks: done, the bush sits 5 blocks away on a stone floor and the fox walks over through its navigation

PR #634, opened 2026-09-08
fix: measure the waypoint distance from the mob, not the surface

darkstarworks:
## Type of change

- [x] Bug fix

## Description

A mob in shallow water can spin in place until its path times out.

Ground navigation keeps both the mob's actual position and a temporary position. Vanilla snaps the temporary Y to the surface for corner cutting and stuck detection, but uses the mob's own position when checking distance to the current waypoint:

```java
double xDistance = Math.abs(this.mob.getX() - (currentNodePos.getX() + 0.5));
double yDistance = Math.abs(this.mob.getY() - currentNodePos.getY());
double zDistance = Math.abs(this.mob.getZ() - (currentNodePos.getZ() + 0.5));
```

Steel used the snapped position for this check too. In shallow water, the snapped Y can be exactly one block above the waypoint, causing the strict vertical check to fail.

The waypoint therefore never retired. The mob kept steering toward the node beneath its feet, repeatedly reversing direction and spinning until the path timed out.

The tick context now carries both positions, and the waypoint check uses the mob's own position like vanilla. Away from water they are effectively identical.

Every mob in the tree registers a float goal, which is what arms the water branch, so the chicken, cow, endermite, pig and sheep can all do it.

## How this was tested

New unit test `standing_in_water_still_retires_the_waypoint_underfoot` in
`steel-core/src/entity/ai/navigation.rs`, verified red before the fix and green after. Reverted, it fails with:

```
left:  DVec3(0.5, 64.0, 0.5)   <- still steering at the node underfoot
right: DVec3(1.5, 64.0, 0.5)
```

In game: leave a pig, cow or sheep near a single block of water or a waterlogged block. Before the fix it spins after entering the block; after, it walks out normally.

## Checklist

- [x] Self-reviewed the diff

- PathNavigation
- GroundPathNavigation

## Additional notes

While tracing this, I noticed `shouldTargetNextNodeInDirection` is missing vanilla's `canMoveDirectly` early return. It is inert for now because the base method returns false and the navigation classes that override it, the amphibious, water-bound and flying ones, do not exist yet. Left alone to avoid widening this PR; worth a TODO when water navigation lands.


CuzImClicks (dismissed):

Joniii11 (approved):
lgtm good job x3 🌵

steel-core/src/entity/ai/navigation.rs
```diff
@@ -28,6 +28,9 @@ pub struct NavigationPathRequest<'a> {
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct NavigationTickContext {
+    /// The mob position with its Y snapped to the surface it navigates along.
+    pub temp_mob_position: DVec3,
```
coco875: hmm maybe ground_mob_position ?

coco875 (changes requested):

steel-core/src/entity/ai/navigation.rs
darkstarworks: done

kdcokenny (approved):

CuzImClicks (approved):

coco875 (approved):

PR #635, opened 2026-09-08
fix: check line of sight for every mob, not only ones that pathfind

darkstarworks:
## Type of change

- [x] Bug fix

## Description

Vanilla's `TargetingConditions` reaches the sensing cache through `Mob`:

```java
if (this.checkLineOfSight && targeter instanceof Mob mob && !mob.getSensing().hasLineOfSight(target)) {
    return false;
}
```

Steel asked for a `PathfinderMob` instead, which is narrower. A `Mob` outside that set skipped the sight check and could target through walls.

Every mob in the tree currently pathfinds, so nothing changes yet. Vanilla's flying mobs are `Mob` without being `PathfinderMob`, the ghast and the phantom among them, so the bug would appear once one of those can target.

The sensing cache already lives on `Mob`, so the lookup moves there unchanged and the condition uses `Mob`.

## How this was tested

New unit test `sight_check_applies_to_a_mob_that_does_not_pathfind` in
`steel-core/src/entity/ai/targeting.rs`. It uses a non-`PathfinderMob`, asserts that fact, and verifies the sight check still applies. Verified red before the fix and green after.

## Checklist

- [x] Self-reviewed the diff

- TargetingConditions
- Mob
- PathfinderMob

## Additional notes

This is latent, so there is no in-game repro. The test covers the mob shape that does not exist in the tree yet and locks in the correct behaviour before it arrives.


Joniii11 (approved):
lgtm good job 💃

steel-core/src/entity/ai/targeting.rs
```diff
@@ -125,3 +125,106 @@ impl Default for TargetingConditions {
...
+
+    impl HoveringTestMob {
+        fn new(id: i32) -> Self {
+            init_vanilla_registry();
+            Self {
+                base: EntityBase::new(
+                    id,
+                    DVec3::ZERO,
+                    vanilla_entities::PIG.dimensions,
+                    Weak::new(),
```
CuzImClicks:
```rust
    fn has_line_of_sight_with(
        &self,
        target: &dyn Entity,
        block_shape: ClipBlockShape,
        fluid: ClipFluid,
        target_eye_y: f64,
    ) -> bool {
        let Some(world) = self.level() else {
            return false;
        };
```

test always succeeds because both entities are constructed with level = Weak::new() and then .level() returns None

CuzImClicks (changes requested):

steel-core/src/entity/ai/targeting.rs
darkstarworks: Good catch. Both mobs are in the test world now; it checks a clear view first, then puts stone between them, and fails again if the old pathfinder-only gate comes back.

CuzImClicks (approved):

coco875 (approved):

CuzImClicks:
you had three prs that could be merged and now every time i merge one, the others are out of date again and i have to wait for you to merge again

PR #636, opened 2026-09-08
fix: keep any gliding entity's velocity synced, not only players

darkstarworks:
## Type of change

- [x] Bug fix

## Description

Vanilla's `ServerEntity` forces a velocity update for a gliding entity, and the
check covers every living entity rather than players alone:

```java
if (this.entity.needsSync || this.trackDelta || this.entity instanceof LivingEntity livingEntity && livingEntity.isFallFlying()) {
```

Steel had `forces_fall_flying_velocity_sync` implemented on the player, leaving
the entity trait default at `false`. A mob wearing elytra therefore stopped
sending velocity updates while gliding. Its position still updated normally,
so clients had to guess its motion until it landed.

The check now lives on the entity trait and matches vanilla's `LivingEntity`
check. The player override is removed because the shared implementation covers it.

On reachability: nothing structural stops a mob gliding. `try_to_start_fall_flying`
sits on the living entity trait here rather than on `Player` as it does in vanilla,
so any living entity can be asked to start. A glide then only continues while
`canGlide` finds an equipped glider, which an elytra on a mob satisfies: the item
sets no allowed entities, so it can go in any mob's chest slot.

Vanilla also hands the flag to a non-player two other ways, through the
`FallFlying` tag on load and through `ConversionType`, which copies it onto the
converted mob.

What is missing here is a caller. Nothing in the tree asks a mob to start gliding,
and Steel writes the `FallFlying` tag without reading it back, so the path is not
reachable yet. Once it is, the missing velocity sync would affect them.

## How this was tested

New unit test `gliding_non_player_living_entity_forces_velocity_sync` in
`steel-core/src/entity/tests/equipment_and_freezing.rs`, which glides a
non-player living entity and asserts its velocity is synced. Verified red
before the fix and green after.

## Checklist

- [x] Self-reviewed the diff

- ServerEntity
- Entity
- Player

## Additional notes

Found while diffing the entity/tracker boundary against the decompiled source.
It is latent rather than currently player-reproducible, so it can be held until
gliding mobs become reachable.

Separately, `FallFlying` is written in `save_living` but never read back, so a
gliding entity stops gliding after save/load. Vanilla restores it in
`readAdditionalSaveData`. That is a separate change and intentionally not part
of this diff.


CuzImClicks (approved):

Joniii11 (approved):
lgtm good job x2 👯

coco875 (approved):

PR #645, opened 2026-09-08
Disconnect clients when Login/Config packet decode fails

### LINKED ISSUE #527: Decode failures log a warning but leave the client connected instead of disconnecting
Follow-up from PR #394 (review comment https://github.com/Steel-Foundation/SteelMC/pull/394#discussion_r3855571325) and #386.

## Summary

When a client packet fails to decode, Steel logs a warning and keeps the connection alive. The client stays connected but is stuck, while vanilla disconnects it as an invalid-packet connection failure.

## Vanilla behavior

- `ServerboundSelectKnownPacks` caps its list via `ByteBufCodecs.list(64)` (`net/minecraft/network/protocol/configuration/ServerboundSelectKnownPacks.java`); more than 64 entries throws during decode.
- Any unhandled pipeline exception (including codec/decode failures) reaches `Connection.exceptionCaught` (`net/minecraft/network/Connection.java`), which sends `ClientboundDisconnectPacket` with a generic reason and closes the channel. Fail-closed: no packet can hang a handler mid-configuration.

## Steel behavior today

`steel-login/src/tcp_client.rs` → `start_incoming_packet_task`: every decoded-packet error from `process_packet` (all `?`-propagated `read_packet` errors in `handle_login`/`handle_config`) hits:

```rust
Err(err) => {
    log::warn!("Failed to get packet from client {id}: {err}");
}
```

and the read loop continues. In configuration state this leaves e.g. an oversized `SSelectKnownPacks` sender connected-but-stuck instead of kicked.

There is already a precedent path for the desired outcome: unexpected packet sequence errors go through `reject_unexpected_packet`, which kicks the client with `MULTIPLAYER_DISCONNECT_INVALID_PACKET`.

## Desired behavior

Route packet decode failures through an invalid-packet disconnect path (send disconnect + close) in Login/Config states, matching vanilla's fail-closed semantics. With that landed, PR #394's `bound = 64` enforcement produces a proper kick for oversized packs; currently it only produces a warn-and-hang for such clients.

## Repro

From PR #394's test setup (azalea fork sending 65 packs, `MODE=wrong`): expected — client is disconnected; actual — warning logged, connection remains until closed externally.

Never-lab:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Follow-up to #527 / the known-packs bound work.

Today if `process_packet` returns `PacketError` (decode blow-up, unexpected packet id in that state, etc.) we only `log::warn!` and keep reading. Client stays connected but stuck. Vanilla tears the connection down from `Connection.exceptionCaught`.

This kicks with `multiplayer.disconnect.invalid_packet` and breaks the incoming loop, same idea as `reject_unexpected_packet` for sequence mistakes.

Wrong packet *order* was already handled. This covers the Err path from reading the packet itself.

Closes #527

## How this was tested

- `cargo check -p steel-login`
- `cargo test -p steel-login --lib` (20 passed)
- Re-read `start_incoming_packet_task` / `handle_login` / `handle_config`: sequence failures still go through `reject_unexpected_packet`; this only changes the `PacketError` arm.

No live azalea repro on my side this time (the oversized known-packs case from #394). Happy to run that if you want it before merge.

## Screenshots / logs

N/A

## Checklist

- [x] Self-reviewed the diff

N/A (login TCP path only)

## Additional notes

Replaces closed #642 (same branch, template + clearer description).

@JunkyDeveloper on packages: yes — this is not about datapacks/resources being “installed”. Known-pack *negotiation that decodes fine* is unchanged. What hits this path is `PacketError` from the codec / unexpected id (e.g. `SSelectKnownPacks` with >64 entries failing `read_packet`). Corrupt or oversize framing ends up here; a normal client with a valid packs list does not.


steel-login/src/tcp_client.rs
```diff
@@ -796,6 +797,15 @@ impl JavaTcpClient {
         ConnectionAction::none()
     }
 
+    /// Kick + close when `process_packet` returns `PacketError` (bad decode / unexpected id).
+    pub(crate) async fn reject_packet_decode_error(&self, error: &PacketError) {
+        log::warn!("Failed to get packet from client {}: {error}", self.id);
+        self.kick(TextComponent::translated(
```
teknostom: Do the same as disconnect_slow_login, put it in timout.

teknostom (changes requested):

steel-login/src/tcp_client.rs
Never-lab: Done — `reject_packet_decode_error` now uses the same `timeout(SLOW_LOGIN_DISCONNECT_FLUSH_TIMEOUT, kick)` + best-effort `close()` pattern as `disconnect_slow_login`.

teknostom (dismissed):
LGTM, extracted the logic to one function instead of replication.

steel-login/src/tcp_client.rs
```diff
@@ -796,6 +804,16 @@ impl JavaTcpClient {
         ConnectionAction::none()
     }
 
+    /// Kick + close when `process_packet` returns `PacketError` (bad decode / unexpected id).
+    pub(crate) async fn reject_packet_decode_error(&self, error: &PacketError) {
+        log::warn!("Failed to get packet from client {}: {error}", self.id);
+        self.kick_with_flush_timeout(
+            TextComponent::translated(translations::MULTIPLAYER_DISCONNECT_INVALID_PACKET.msg()),
```
CuzImClicks, line 811:
```suggestion
            TextComponent::translated(
                translations::DISCONNECT_GENERIC_REASON
                    .message([format!("Internal Exception: {error}")]),
            ),
```

on packet decode failure this is sent:

```java
Component reason = Component.translatable("disconnect.genericReason", "Internal Exception: " + cause);
```

CuzImClicks (changes requested):

steel-login/src/tcp_client.rs
Never-lab: Updated — decode failures now kick with `disconnect.genericReason` / `Internal Exception: {error}`, matching `Connection.exceptionCaught`.

CuzImClicks (approved):

banocean (approved):

coco875 (approved):
for future pr for response I prefer if you use your own word instaed of using ai to respond

PR #649, opened 2026-09-09
feat: implement the bottle o' enchanting

JosepBove:
## Type of change

- [ ] Block implementation
- [x] Item implementation
- [ ] Command implementation
- [x] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Adds the Bottle o' Enchanting, item and projectile, matching vanilla 26.2:

- `ExperienceBottleItem`: plays the throw sound (volume 0.5, the usual `0.4 / (rand * 0.4 + 0.8)` pitch jitter), spawns the bottle from the player's eye 20 degrees above their look direction with power 0.7 and uncertainty 1.0, awards the item used stat and consumes one bottle unless the player has infinite materials.
- `ThrownExperienceBottleEntity`: sits on the existing `Projectile -> ThrowableProjectile -> ThrowableItemProjectile` stack. Gravity is 0.07 so it drops faster than a snowball, it deals no damage, and on impact it relays level event 2002 with the water potion colour, awards `3 + nextInt(5) + nextInt(5)` experience and discards itself. Owner, item stack and NBT round trip come from the shared traits.
- `ExperienceOrbEntity`: the orbs from a bottle need vanilla's `awardWithDirection`, so the orb now has the `(pos, roughDirection, value)` constructor path. Random spawn motion gets flipped to face the rough direction, the orb is nudged half its size along it, and `unstuckIfPossible` pushes it to the nearest free spot if it would start inside a block. Plain `award` still goes through the same code with a zero direction, so existing callers are unchanged.
- `spawn_throwable_item_projectile` takes a pitch offset (vanilla's `z` argument to `shootFromRotation`). Egg, snowball and ender pearl pass 0.0 and keep their behaviour.

On a block hit the orbs fly away from the face that was hit; on an entity hit they fly back along the bottle's flight path, same as vanilla.

Vanilla reference: Minecraft 26.2 `ExperienceBottleItem`, `ThrownExperienceBottle`, `ExperienceOrb`, `Projectile`.

Out of scope: dispenser support (`asProjectile` / `createDispenseConfig`) since there is no dispenser block yet, and the splash particles themselves, which the client renders from the level event.

## How this was tested

Target: Minecraft 26.2 (`0.15.3+mc26.2`), Windows 11.

Automated, all green on the branch head:
- `cargo test` for the whole workspace
- release clippy over all targets and features with `RUSTFLAGS=-Dwarnings`
- `cargo fmt` check
- `typos`

New tests cover: orb velocity flipped toward the rough direction, the half-size position nudge (and no nudge for a zero direction), `award_with_direction` splitting 11 XP into 7 + 3 + 1 orbs, an orb spawned inside stone getting unstuck, the pitch offset on the spawn helper producing the expected initial velocity, the bottle's gravity / default item / non-attackable defaults, the XP roll range, and both hit paths (block face and entity) awarding orbs biased the right way and discarding the bottle.

## Checklist

- [x] Self-reviewed the diff

- ExperienceBottleItem
- ThrownExperienceBottle

## Additional notes

- Vanilla's `Vec3.normalize()` returns zero for anything shorter than 1.0E-4 rather than only for an exact zero vector, so the orb nudge checks `length_squared() < 1.0e-8` before normalising. Commented at the call site.
- Vanilla's `Level.noCollision(AABB)` also checks entity collisions, so the orb unstuck check reuses the tick's `is_aabb_colliding` helper instead of a block-only test.
- Experience roll and orb spawn motion use the runtime RNG like the other throwables; none of it is seeded or save-visible.


steel-core/src/entity/entities/objects/items/experience_orb.rs
```diff
@@ -195,15 +223,62 @@ impl ExperienceOrbEntity {
...
+        };
+        self.base
+            .set_position_local(self.position() + nudge_direction * (size * 0.5));
         self.base.set_rotation((yaw, 0.0));
         self.base.set_velocity(velocity);
+
+        if let Some(world) = self.level()
+            // Vanilla `Level.noCollision(AABB)` is `noCollision(null, box)`, which
+            // also checks entity collisions, so reuse the tick's helper for it.
+            && self.is_aabb_colliding(&world, self.base.bounding_box())
```
Joniii11, line 256: mc uses a sourcekess noCollision(null, box) here ts entityaware check changes collision semantics to what we do

steel-core/src/entity/entities/objects/projectiles/experience_bottle.rs
```diff
@@ -0,0 +1,340 @@
...
+    fn projectile_owner_uuid(&self) -> Option<uuid::Uuid> {
+        self.owner_uuid()
+    }
+
+    fn projectile_owner(&self) -> Option<SharedEntity> {
+        self.get_owner()
+    }
+
+    fn attackable(&self) -> bool {
+        false
```
Joniii11, line 127: funny or  ot but in vanilla this is true so we can attack it

steel-core/src/entity/entities/objects/projectiles/experience_bottle.rs
```diff
@@ -0,0 +1,340 @@
...
+    fn base(&self) -> &EntityBase {
+        &self.base
+    }
+
+    fn entity_type(&self) -> EntityTypeRef {
+        self.entity_type
+    }
+
+    fn tick(&self) {
+        self.throwable_projectile_tick();
```
Joniii11, line 98: minecraft applies the first tick bubblecolumn effect before any gravity or inertia the shared tick still skips it so yeah bad

steel-core/src/entity/entities/objects/projectiles/experience_bottle.rs
```diff
@@ -0,0 +1,340 @@
...
+        false
+    }
+
+    fn synced_data(&self) -> Option<&dyn EntitySyncedData> {
+        Some(&self.entity_data)
+    }
+
+    fn save_additional(&self, nbt: &mut NbtCompound) {
+        self.save_projectile(nbt);
+        self.save_throwable_item(nbt);
```
Joniii11, line 136: so basically mc always saves the throwable item ncluding empty one but now we just reload an omitted empty stack as the default item

steel-core/src/behavior/items/experience_bottle.rs
```diff
@@ -0,0 +1,83 @@
...
+        world.play_sound_at(
+            &sound_events::ENTITY_EXPERIENCE_BOTTLE_THROW,
+            SoundSource::Neutral,
+            player.position(),
+            THROW_SOUND_VOLUME,
+            pitch,
+            None,
+        );
+
+        let mut thrown_item = context.inv.with_item(|item| item.clone());
```
Joniii11, line 55: would it be possibld to avoid tve whole cloning here?

steel-core/src/entity/entities/objects/items/experience_orb.rs
```diff
@@ -195,15 +223,62 @@ impl ExperienceOrbEntity {
...
+    /// `unstuckIfPossible` when the orb would start inside a block.
+    fn initialize_spawn_movement(&self, rough_direction: DVec3) {
         let yaw = rand::random_range(0.0..DEGREE_360);
-        let velocity = DVec3::new(
+        let mut velocity = DVec3::new(
             f64::from(rand::random_range(-0.1f32..0.1)) * 2.0,
             f64::from(rand::random_range(0.0f32..0.2)) * 2.0,
             f64::from(rand::random_range(-0.1f32..0.1)) * 2.0,
         );
+        if rough_direction.length_squared() > 0.0 && rough_direction.dot(velocity) < 0.0 {
```
Joniii11, line 236: can we just compute the length_squared once?

steel-core/src/entity/entities/objects/items/experience_orb.rs
```diff
@@ -195,15 +223,62 @@ impl ExperienceOrbEntity {
...
+            velocity = -velocity;
+        }
+
+        let size = self.base.bounding_box().size();
+        // Vanilla `Vec3.normalize()` returns the zero vector below length 1.0E-4
+        // rather than only for an exactly-zero input.
+        let nudge_direction = if rough_direction.length_squared() < 1.0e-8 {
```
Joniii11, line 243: can we just compute the length_squared once?

x2

Joniii11, line 243: magic the number!!!

steel-core/src/entity/entities/objects/items/experience_orb.rs
```diff
@@ -195,15 +223,62 @@ impl ExperienceOrbEntity {
...
+            DVec3::ZERO
+        } else {
+            rough_direction.normalize()
+        };
+        self.base
+            .set_position_local(self.position() + nudge_direction * (size * 0.5));
```
Joniii11, line 249: magic 0.5

steel-core/src/entity/entities/objects/items/experience_orb.rs
```diff
@@ -195,15 +223,62 @@ impl ExperienceOrbEntity {
...
+        }
+    }
+
+    /// Vanilla `ExperienceOrb.unstuckIfPossible`: moves the orb to the closest
+    /// free spot within `max_distance` of its center.
+    fn unstuck_if_possible(&self, world: &Arc<World>, max_distance: f64) {
+        let dimensions = self.base.dimensions();
+        let width = f64::from(dimensions.width);
+        let height = f64::from(dimensions.height);
+        let center = self.position() + DVec3::new(0.0, height / 2.0, 0.0);
```
Joniii11, line 268: also magic

steel-core/src/entity/entities/objects/items/experience_orb.rs
```diff
@@ -195,15 +223,62 @@ impl ExperienceOrbEntity {
...
+            max_distance,
+            max_distance,
+            max_distance,
+        )];
+        let provider = WorldCollisionProvider::for_entity(world, self.as_entity_event_source());
+        if let Some(free_center) =
+            provider.find_free_position(&allowed_centers, center, width, height, width)
+        {
+            self.base
+                .set_position_local(free_center - DVec3::new(0.0, height / 2.0, 0.0));
```
Joniii11, line 280: magic yeah ik but whar tf does 2.0 refer to

steel-core/src/entity/entities/objects/items/experience_orb.rs
```diff
@@ -195,15 +223,62 @@ impl ExperienceOrbEntity {
...
             f64::from(rand::random_range(-0.1f32..0.1)) * 2.0,
             f64::from(rand::random_range(0.0f32..0.2)) * 2.0,
             f64::from(rand::random_range(-0.1f32..0.1)) * 2.0,
         );
+        if rough_direction.length_squared() > 0.0 && rough_direction.dot(velocity) < 0.0 {
+            velocity = -velocity;
+        }
+
+        let size = self.base.bounding_box().size();
+        // Vanilla `Vec3.normalize()` returns the zero vector below length 1.0E-4
```
Joniii11, line 241: well the comment is misleading it says 1.0e-4 but below is e-8

steel-core/src/behavior/items/egg.rs
```diff
@@ -54,6 +54,7 @@ impl ItemBehavior for EggItem {
             world,
             player,
             &mut thrown_item,
+            0.0,
```
Joniii11, line 57: well yeah magic but couldnt we just use like 1 default constant cause you chanfed that in multiple files

Joniii11 (changes requested):

steel-core/src/entity/entities/objects/items/experience_orb.rs
```diff
@@ -111,19 +126,32 @@ impl ExperienceOrbEntity {
         }
     }
 
-    /// Spawns vanilla experience orbs for an XP amount.
-    pub fn award(world: &Arc<World>, position: DVec3, mut amount: i32) {
+    /// Vanilla `ExperienceOrb.award`: spawns orbs for an XP amount with no spawn bias.
```
coco875, line 129: precise the vanilla name aren't useful like it's very transparent

steel-core/src/entity/entities/objects/items/experience_orb.rs
```diff
@@ -111,19 +126,32 @@ impl ExperienceOrbEntity {
...
+    pub fn award(world: &Arc<World>, position: DVec3, amount: i32) {
+        Self::award_with_direction(world, position, DVec3::ZERO, amount);
+    }
+
+    /// Vanilla `ExperienceOrb.awardWithDirection`: splits `amount` into vanilla
+    /// orb values, merging into nearby matching orbs first, and biases each new
+    /// orb's spawn motion along `rough_direction`.
```
coco875, line 136: precise the vanilla name aren't useful like it's very transparent

steel-core/src/entity/entities/objects/items/experience_orb.rs
```diff
@@ -195,15 +223,62 @@ impl ExperienceOrbEntity {
         self.state.lock().health
     }
 
-    fn initialize_spawn_movement(&self) {
+    /// Vanilla spawn motion: random yaw, random velocity flipped to face
+    /// `rough_direction`, the position nudged half the orb's size along it, then
+    /// `unstuckIfPossible` when the orb would start inside a block.
```
coco875, line 228: why precise it's vanilla ? if it's just the behavior of the function curently

steel-core/src/entity/entities/objects/projectiles/experience_bottle.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    fn entity_type(&self) -> EntityTypeRef {
+        self.entity_type
+    }
+
+    fn tick(&self) {
+        self.throwable_projectile_tick();
+    }
+
+    /// Vanilla `ThrownExperienceBottle.getDefaultGravity` (0.07, heavier than other thrown items).
```
coco875, line 101: precise the vanilla name aren't useful like it's very transparent

steel-core/src/entity/entities/objects/projectiles/experience_bottle.rs
```diff
@@ -0,0 +1,340 @@
...
+    /// biased away from the hit surface, then discard. No entity damage.
+    fn on_hit(&self, hit: &ProjectileHit) {
+        self.projectile_on_hit(hit);
+
+        let Some(world) = self.level() else {
+            return;
+        };
+
+        // VANILLA CLIENT-LOCAL: level event 2002 renders the splash particles on
+        // clients; the server only relays it with the water potion color.
```
coco875, line 160: the comment aren't useful

steel-core/src/entity/entities/objects/projectiles/experience_bottle.rs
```diff
@@ -0,0 +1,340 @@
...
+    }
+}
+
+impl Projectile for ThrownExperienceBottleEntity {
+    fn projectile_base(&self) -> &ProjectileBase {
+        &self.projectile_base
+    }
+
+    /// Vanilla `ThrownExperienceBottle.onHit`: splash event, experience orbs
+    /// biased away from the hit surface, then discard. No entity damage.
```
coco875, line 151: precise the vanilla name aren't useful like it's very transparent

steel-core/src/entity/projectile/mod.rs
```diff
@@ -664,15 +664,18 @@ pub trait Projectile: Entity + ProjectileEventSource {
...
-/// `shootFromRotation`, add it to the world, then run `applyOnProjectileSpawned`
-/// (the `minecraft:projectile_spawned` enchantment effects). `create` receives
-/// the spawn position and must return the concrete projectile type. Returns
-/// `None` when the world rejects the projectile so the caller can fail the use.
+/// `shootFromRotation` (adding `pitch_offset` degrees to the player's pitch,
+/// vanilla's `z` argument), add it to the world, then run
+/// `applyOnProjectileSpawned` (the `minecraft:projectile_spawned` enchantment
+/// effects). `create` receives the spawn position and must return the
+/// concrete projectile type. Returns `None` when the world rejects the
+/// projectile so the caller can fail the use.
```
coco875, line 672: I think the comment can be shorter

coco875 (changes requested):

PR #650, opened 2026-09-09
feat(block): Added grass and mycelium spreading

ezraclintoc:
## Type of change

- [x] Block implementation

## How this was tested

prek run
and going ingame and testing to see if grass and mycelium spread.
and seeing if placing a block above grass and mycelium made them fade away

## Checklist

- [x] Self-reviewed the diff

Created the SpreadingSnowyBlock file and class to match with java, but it could easily be renamed to Spreading if needed.

The current code makes grass and mycelium spreads. And grass and mycelium will fade into dirt if they they have a block above them.


steel-core/src/behavior/blocks/vegetation/spreading_snowy_block.rs
```diff
@@ -0,0 +1,79 @@
...
+                self.base_block.default_state(),
+                UpdateFlags::UPDATE_ALL,
+            );
+        } else if world.max_local_raw_brightness(pos.above(), world.sky_darkening()) >= 9 {
+            let default_block_state = self.own_block.default_state();
+
+            for _ in 0..4 {
+                // This copies the java code, but maybe we should change this to have a range of -1..2
+                let test_pos = pos.offset(
+                    rand::random_range(0..3) - 1,
```
ezraclintoc: is there any reason not to do -1..2?

steel-core/src/behavior/blocks/vegetation/spreading_snowy_block.rs
```diff
@@ -0,0 +1,78 @@
...
+        if above_state.get_block() == &vanilla_blocks::SNOW
+            && above_state.get_value(&BlockStateProperties::LAYERS) == 1
+        {
+            return true;
+        }
+        if above_state.get_fluid_state().is_full() {
+            return false;
+        }
+
+        let light_dampening_top_face = above_state.get_light_dampening();
```
CuzImClicks:
```suggestion
        let light_dampening_top_face = get_light_block_into(
            state,
            above_state,
            Direction::Up,
            above_state.get_light_dampening(),
        );
```

dont forget to fix imports

steel-core/src/behavior/blocks/vegetation/spreading_snowy_block.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+    fn can_propagate(state: BlockStateId, level: &Arc<World>, pos: BlockPos) -> bool {
+        Self::can_stay_alive(state, level, pos)
+            && !level
+                .get_block_state(pos.above())
+                .get_fluid_state()
+                .fluid_id
+                .has_tag(&FluidTag::WATER)
+    }
+    pub(super) fn random_tick(&self, state: BlockStateId, world: &Arc<World>, pos: BlockPos) {
```
CuzImClicks:
```suggestion
    pub fn random_tick(
        own: BlockRef,
        base: BlockRef,
        state: BlockStateId,
        world: &Arc<World>,
        pos: BlockPos,
    ) {
```

own_block and base_block are only used here in this function, we dont actually need to make SpreadingSnowyBlock have any attributes at all. it becomes a zero sized struct that is only used to stop the methods from being free floating (instead of "random_tick" you refer to it through SpreadingSnowyBlock::random_tick)

this allows us to get rid of the attributes on GrassBlock and Mycelium block. i also want you to write some docs for the methods and use pub instead of pub crate. this goes for all functions in this struct. (dont forget to export SpreadingSnowyBlock in the mod.rs through pub use ... so that they actually become public).

CuzImClicks (changes requested):
good work, this will get merged in no time i believe, for those who want to play grass vs mycelium join the server and then paste this into the console:

```
gamemode creative @a
tp @a 0 250 0
gamerule max_block_modifications 500000
gamerule random_tick_speed 20
gamerule advance_time false
gamerule spawn_mobs false
time set noon
fill -101 200 -101 101 201 101 minecraft:stone_bricks
fill -100 200 -100 100 201 100 minecraft:air
fill -100 200 -100 100 200 100 minecraft:dirt
fill -100 200 -100 -90 200 100 minecraft:mycelium
fill 90 200 -100 100 200 100 minecraft:grass_block
tp @a 0 201 0
```

then you can run /tick sprint 3d to watch it

steel-core/src/behavior/blocks/vegetation/spreading_snowy_block.rs
```diff
@@ -0,0 +1,81 @@
...
+use std::sync::Arc;
+use steel_registry::blocks::properties::BlockStateProperties;
+use steel_registry::entity_data::Direction;
+use steel_registry::vanilla_blocks;
+use steel_registry::{blocks::block_state_ext::BlockStateExt, vanilla_fluid_tags::FluidTag};
+use steel_utils::types::UpdateFlags;
+use steel_utils::{BlockPos, BlockStateId};
+
+/// A structure implementing the spreading of grass blocks and its variants
+pub struct SpreadingSnowyBlock {}
```
coco875: I think that SpreadingGrassBlock can be better or a name that are easier to understand that it's for mycelium, grass, etc

steel-core/src/behavior/blocks/vegetation/spreading_snowy_block.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl SpreadingSnowyBlock {
```
coco875: you can remove the wrapper in a struct/impl, it's useless

steel-core/src/behavior/blocks/vegetation/spreading_snowy_block.rs
```diff
@@ -0,0 +1,81 @@
...
+            return false;
+        }
+
+        let light_dampening_top_face = get_light_block_into(
+            state,
+            above_state,
+            Direction::Up,
+            above_state.get_light_dampening(),
+        );
+        light_dampening_top_face < 15
```
coco875: we have a const for MAX_LIGHT

coco875 (changes requested):

steel-core/src/behavior/blocks/vegetation/spreading_snowy_block.rs
ezraclintoc: ok, will fix

I just picked SpreadingSnowyBlock because that is the java/vanilla name

steel-core/src/behavior/blocks/vegetation/spreading_snowy_block.rs
ezraclintoc: will fix

steel-core/src/behavior/blocks/vegetation/spreading_snowy_block.rs
ezraclintoc: ok, will fix

CuzImClicks: actually i requested the zero sized struct because i dont like free floating functions. this gives them a dedicated space to live in yk

ezraclintoc: it was there for just a little big of organization, and for exporting it, making a public random_tick function could be confusing

CuzImClicks: i just think this `SpreadingSnowyBlock::random_tick` is cleaner than spreading_random_tick

coco875: like we didn't find a conclusion just let it like that and fix other comment, I will approve and I let it open just in case but if a second maintener approve then it's a problem that we can alway fix it later

Joniii11: i like clicks solution way better

CuzImClicks (approved):

coco875 (approved):

PR #651, opened 2026-09-09
feat(core): make server startup embeddable

Purdze:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactor / code cleanup
- [ ] Performance improvement
- [x] Chore / tooling

## Description

Steel was built standalone-only, so `steel-core` carries a few assumptions that are
correct for a dedicated server and wrong for a library. This lifts five of them so the
crate can also be driven in-process, without changing how the standalone server behaves.

- **`init_globals` is now idempotent.** It returned
  `Err("global registry has already been initialized")` on a second call, because
  `init_vanilla_registry` only reports `published` the first time. `Server::new` called
  it, so constructing a second server in one process failed. The idempotent
  `init_globals_once` next to it was gated to tests and benches; the two are now one
  function guarded by a `Once`, matching vanilla `Bootstrap.bootStrap`, which is likewise
  idempotent behind a flag and carries no suffix. The rename is a separate commit.

- **`save_path` may now be absolute.** It went through `validate_relative_path`, so it
  resolved against the process working directory. That is right for a server that owns
  its directory, but a caller that cannot chdir per world needs to point at one. The new
  `validate_save_path` accepts absolute paths and still rejects `.` and `..`;
  `validate_relative_path` keeps its meaning for `storage.config.path`, which is joined
  under the save root. Vanilla `LevelStorageSource` takes an arbitrary `Path baseDir`.

- **Offline startup no longer waits on the Mojang services keys.** `Server::new` awaited
  the initial key fetch unconditionally, five seconds of connect timeout on a machine
  with no route. With online mode off secure chat can never be *enforced*, since
  `enforces_secure_chat` requires it, so blocking startup on the fetch buys nothing. The
  keys are still read offline by `handle_chat_session_update`, which has no online-mode
  gate, so the background refresh still starts exactly as before; only the await is gated.
  Vanilla never blocks construction on this either, and never gates the fetch on online
  mode: `Main.java:126` builds `Services` before the server exists, and
  `handleChatSessionUpdate` reads the validator ungated.

- **`RuntimeConfig` has a `Default`.** It has 21 public fields and no constructor, so
  every field added upstream broke every caller. The defaults reproduce a freshly
  packaged server, leaving the optional service endpoints and server links unset. This
  also collapses two byte-identical test helpers, and the spam-threshold and
  neighbor-update literals now live beside `RuntimeConfig` so the `steel` crate's serde
  defaults reference them instead of repeating them.

- **The root manifest no longer declares `cargo-features`.** Its only consumer is a
  commented-out profile line. Stable cargo refuses to load a manifest declaring a
  nightly-only cargo feature, so the key blocked non-nightly workspaces from taking Steel
  as a path or git dependency. Note this is not a blocker on the pinned nightly, where
  the key is simply inert.

## How this was tested

`cargo fmt --all --check`, `cargo clippy -r --all-targets --all-features` under
`-Dwarnings`, `typos .`, and `cargo test --workspace` all pass. Each new behavior has a
test, and each was checked against the pre-fix code rather than only observed passing:

- `a_second_server_bootstraps_in_the_same_process` builds two servers through
  `Server::new` in one process. Reinstating the pre-fix bootstrap makes it fail with the
  registry error.
- `offline_startup_does_not_wait_for_the_services_key_fetch` points an offline server at
  an unroutable endpoint (RFC 5737 documentation space, so the connect stalls rather than
  being refused) and asserts startup finishes inside the connect timeout. Removing the
  offline gate makes it fail at five seconds.
- `offline_startup_still_requests_the_services_keys` points an offline server at a local
  listener and asserts the request arrives, so skipping the wait cannot become skipping
  the fetch. Gating `ServiceKeyStore::start` on online mode makes it fail.
- `packaged_config_matches_the_runtime_config_defaults` compares `RuntimeConfig::default()`
  against the packaged `config.toml` field by field. The destructure is exhaustive, so a
  new field stops it compiling until someone decides what the packaged config says.
- `resolves_an_absolute_save_path`, `rejects_traversal_in_a_save_path`,
  `rejects_an_empty_save_path` and `world_storage_paths_stay_relative_to_the_save_root`
  cover the path change from both sides.

The three server-startup tests run on a thread with `DEBUG_STACK_SIZE`, the same stack
`main` gives Steel, since Windows debug builds otherwise overflow in the generated
density functions.

## Checklist

- [x] Self-reviewed the diff


steel-core/src/server/mod.rs
CuzImClicks, line 565:
```suggestion
        let service_keys_ready = config
            .online_mode
            .then(|| service_keys.start(cancel_token.clone()));
```

only launch when config mode is actually true since you say yourself it can never consult the keys anyway

steel-core/src/server/tests.rs
```diff
@@ -3322,3 +3322,104 @@ fn setblock_command_places_blocks_and_keep_mode_skips_occupied_positions() {
...
+fn offline_startup_does_not_wait_for_the_services_key_fetch() {
+    with_server_runtime(|runtime| {
+        runtime.block_on(async {
+            let save_root = test_storage_root("offline-service-keys");
+
+            let started = SystemTime::now();
+            let server = offline_server(&save_root)
+                .await
+                .expect("an offline server should start");
+            let elapsed = started.elapsed().expect("startup should be monotonic");
```
CuzImClicks:
```suggestion
            let started = Instant::now();
            let server = offline_server(&save_root)
                .await
                .expect("an offline server should start");
            let elapsed = started.elapsed();
```

just use instant

CuzImClicks (changes requested):

CuzImClicks (dismissed):

steel-core/src/server/mod.rs
```diff
@@ -550,13 +550,15 @@ impl Server {
...
+        init_globals();
         log::info!(
             "SteelMC is not affiliated with Mojang or Microsoft. Use is subject to the Minecraft EULA: https://aka.ms/MinecraftEULA"
         );
 
         // Authlib starts this fetch alongside server initialization and waits on first use.
-        // Steel completes the same initial attempt before opening its listener.
+        // Steel completes the same initial attempt before opening its listener. Offline
+        // servers skip the wait but still fetch: `enforces_secure_chat` needs online mode,
+        // but `handle_chat_session_update` reads the keys ungated, as vanilla does.
```
coco875: the adding to the comment just complet what the line 703 do, it's necessary ?

steel-core/src/server/tests.rs
```diff
@@ -3322,3 +3323,147 @@ fn setblock_command_places_blocks_and_keep_mode_skips_occupied_positions() {
...
+default = true
+
+[[domains.minecraft.worlds]]
+name = "overworld"
+generator = "minecraft:flat"
+default = true
+"#,
+        save_root.display()
+    ))
+    .expect("worlds config should parse");
```
coco875, line 3352: why not define WorldsConfig directly instead of parsing it from a toml ?

steel-core/src/server/tests.rs
```diff
@@ -3322,3 +3323,147 @@ fn setblock_command_places_blocks_and_keep_mode_skips_occupied_positions() {
...
+        CancellationToken::new(),
+        config,
+        worlds_config,
+        PermissionGroupManager::transient(PermissionGroupsConfig::default())
+            .expect("default permission groups should resolve"),
+    )
+    .await
+}
+
+/// Stops a started server and drops its save directory, best effort.
```
coco875: the comment are necessary ?

steel-core/src/config.rs
```diff
@@ -92,6 +98,36 @@ pub struct RuntimeConfig {
     pub chunk_encoding_threads: Option<usize>,
 }
 
+impl Default for RuntimeConfig {
+    /// The settings a freshly packaged server ships, leaving the optional service
+    /// endpoints and server links unset.
+    fn default() -> Self {
```
coco875, line 104: hmm why not define more const for all field ?

coco875: (at least max_player, view_distance, simulation_distance) even if it's not public

steel-core/src/config.rs
```diff
@@ -352,7 +388,8 @@ impl WorldsConfig {
         generator_registry: &WorldGeneratorRegistry,
         storage_registry: &WorldStorageRegistry,
     ) -> Result<ResolvedWorldsConfig, String> {
-        validate_relative_path(&self.save_path, "save_path")?;
+        // An embedded server cannot chdir per world, so its save root may be absolute.
```
coco875: the comment are necessary ?

steel-core/src/config.rs
```diff
@@ -707,29 +744,39 @@ fn validate_world_name(name: &str, domain: &str) -> Result<(), String> {
     Ok(())
 }
 
-/// Validates a config path as a relative path under a Steel-owned root.
-pub fn validate_relative_path(path: &str, field: &str) -> Result<(), String> {
+/// Validates a config path, rejecting `.` and `..` components.
+///
+/// Absolute paths are accepted. Use [`validate_relative_path`] for paths that have to stay
+/// under a Steel-owned root.
```
coco875, line 751: pretty sure it can be shorten

steel/src/config/tests.rs
```diff
@@ -53,6 +50,84 @@ fn packaged_configs_parse() {
     assert!(DEFAULT_GROUPS.starts_with(GROUPS_CONFIG_HEADER));
 }
 
+/// `RuntimeConfig::default()` claims to be what a freshly packaged server ships, so the two
+/// are checked against each other rather than kept in step by hand. The destructure is
+/// exhaustive on purpose: a new field stops this compiling until someone decides what the
+/// packaged config says about it.
```
coco875: the comment are needed ?

steel/src/config/tests.rs
```diff
@@ -53,6 +50,84 @@ fn packaged_configs_parse() {
...
+    assert_eq!(
+        chat_spam_threshold_seconds,
+        defaults.chat_spam_threshold_seconds
+    );
+    assert_eq!(
+        command_spam_threshold_seconds,
+        defaults.command_spam_threshold_seconds
+    );
+
+    // `CompressionInfo` has no `PartialEq`, so its two fields stand in for the whole.
```
coco875: I don't think that the comment are necessary

steel/src/config/tests.rs
```diff
@@ -53,6 +50,84 @@ fn packaged_configs_parse() {
...
+        .expect("default config enables compression");
+    assert_eq!(
+        packaged_compression.threshold,
+        default_compression.threshold
+    );
+    assert_eq!(packaged_compression.level, default_compression.level);
+
+    // The remaining four diverge by design, so only the packaged side is pinned. Links are
+    // an embedder's to choose, and a thread count of zero means the same automatic default
+    // that `None` does.
```
coco875: I don't think that the comment are necessary

Cargo.toml
```diff
@@ -161,6 +160,7 @@ debug = true
 
 [profile.dev]
 debug = 0
+# Uncommenting this needs `cargo-features = ["codegen-backend"]` back at the top of the file.
```
coco875: the comment can be remove ?

coco875 (changes requested):

steel-core/src/server/mod.rs
Purdze: not quite the same thing, 703 says the wait is skipped offline, the comment says why the fetch still runs, which is the opposite direction. clicks read the old wording and opened a suggestion to drop the fetch entirely earlier in this pr, which would leave handle_chat_session_update without a validator forever. i'd rather keep the line that stops that

steel-core/src/config.rs
Purdze: it's there so nobody swaps it back to validate_relative_path

steel/src/config/tests.rs
Purdze: without it someone adds .. and the test silently stops covering new fields

CuzImClicks (dismissed):

steel-core/src/server/mod.rs
coco875: then put the comment above 703 would be better no ?

steel-core/src/config.rs
coco875: then maybe "keep it validate_clean_path for embedded server" ?

steel/src/config/tests.rs
coco875: then we can short it to say that it test all field and to extend when we add field ?

coco875 (approved):

CuzImClicks (approved):

PR #655, opened 2026-09-10
fix: keep valid chunk tickets when saved data has bad entries

### LINKED ISSUE #646: Keep valid chunk tickets when saved ticket data contains errors
a single invalid entry in `chunk_tickets.toml` currently fails the whole world load. this already happened before #577 for unknown ticket types, missing fields, and malformed TOML.

vanilla logs decoding errors and keeps partial results when available. if nothing usable loads, it creates empty ticket storage.

we should decode tickets individually, keep valid entries, and log rejected ones with enough context to find them. if the whole ticket file can't be loaded, log the failure and start with empty ticket storage.

keep this specific to chunk tickets. required fields should stay required, without guessed defaults or migrations.

test a valid -> invalid -> valid list, missing fields, unknown types, invalid levels, and an unusable file.

Mathildeuh:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Fixes #646. A single bad entry in `chunk_tickets.toml` failed the whole world load.

- `ChunkTicketStorage::from_persistent` is now infallible. An entry that cannot be restored -
  an unknown ticket type, an out-of-range level - is logged with its list index and
  skipped, keeping the rest. `add_loaded_persistent_ticket` still returns the typed error
  it always did; only `from_persistent`'s contract changed.
- `deserialize_lenient_tickets` decodes the saved list entry by entry (each as a
  `toml::Value`, then `try_into` the strict `PersistentChunkTicket`). A missing or
  wrong-typed field in one entry no longer drops the rest. Entries are not repaired -
  required fields stay required, no guessed defaults or migrations.
- `World::new_with_config_and_encoding_pool` no longer propagates a load error from
  `chunk_tickets.toml`. An unreadable document, or one that is not a ticket list at all, is
  logged and the world starts with empty ticket storage, matching vanilla.
- Kept specific to chunk tickets: the generic `SavedDataManager::load_or_default` is
  unchanged.

## How this was tested

`chunk::chunk_ticket_storage::tests`:

- `persistence_resolves_registered_type_and_skips_invalid_entries` - reworked from the old
  "rejects invalid values" test: an unknown type is dropped while a valid entry beside it
  survives; an out-of-range level is dropped.
- `an_invalid_ticket_entry_is_skipped_without_dropping_the_valid_ones` - valid ->
  missing-field -> valid.
- `a_document_that_is_not_a_ticket_list_fails_to_decode` - the caller then falls back to
  empty storage.

`cargo test`, `cargo clippy -r --all-targets --all-features`, `cargo fmt --all --check`,
`typos`, and `prek run --all-files` all pass.

## Checklist

- [ ] Self-reviewed the diff

## Additional notes

- The per-entry decode goes through `toml::Value` because chunk ticket saved data is
  TOML-only. A format-agnostic saved-data layer would need a different hook.
- `ChunkTicketStorageLoadError` keeps its `PartialEq`/`Eq` derives even though the tests no
  longer compare it directly; happy to drop them if preferred.


steel-core/src/chunk/chunk_ticket_storage.rs
```diff
@@ -577,26 +606,33 @@ mod tests {
...
-            error,
-            ChunkTicketStorageLoadError::UnknownTicketType(Identifier::new_static(
-                "test", "missing"
-            ))
+            ChunkTicketStorage::from_persistent(with_unknown_type).ticket_count(),
+            1
         );
 
+        // An out-of-range level is dropped.
         let invalid_level = ChunkTicketLevel::MAX.raw() + 1;
```
Joniii11, line 638: afaik this is existing steel behaviour but this test now calls it vanilla MAX + 1 is still valid ticket level in vanilla maybe track this separately instead of calling it invalid here

steel-core/src/chunk/chunk_ticket_storage.rs
```diff
@@ -20,10 +20,36 @@ type StoredTickets = Vec<StoredChunkTicket>;
...
+    deserializer: D,
+) -> Result<Vec<PersistentChunkTicket>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let raw_entries = Vec::<toml::Value>::deserialize(deserializer)?;
+    let mut tickets = Vec::with_capacity(raw_entries.len());
+    for (index, raw_entry) in raw_entries.into_iter().enumerate() {
+        match raw_entry.try_into::<PersistentChunkTicket>() {
+            Ok(ticket) => tickets.push(ticket),
```
Joniii11, line 44: you lose the original index here after one bad entry the later logged indexes can be wrong

Joniii11 (changes requested):
I only want to request changes for the second review this one
> you lose the original index here after one bad entry the later logged indexes can be wrong

the other thing is just a comment more or less idc

steel-core/src/chunk/chunk_ticket_storage.rs
Mathildeuh: Fixed in 34b5dcf67 - logs the ticket's (chunk_x, chunk_z) instead of an index now, so there's nothing to lose track of.

steel-core/src/chunk/chunk_ticket_storage.rs
Mathildeuh: Yeah that's pre-existing, not something this PR touches - ChunkTicketLevel::MAX here is Steel's own supported range (tied to MAX_SUPPORTED_VIEW_DISTANCE), not vanilla's actual max level. Fine with opening a separate issue for it if you want it tracked.

Joniii11: yeah sure would be nice cause im lazy

Mathildeuh: Opened #678 for it.

Mathildeuh: Closed #678 - turns out our ticket level scale isn't the same as vanilla's at all (ours is stretched to fit view distance up to 128, vanilla's is a fixed 33+radius), so comparing raw values across them doesn't really apply. Details in the issue.

CuzImClicks (changes requested):
please fix ci so this can be reviewed

PR #660, opened 2026-09-11
fix: derive global palette bit widths from the registry

### LINKED ISSUE #658: Global palette bit widths
`calculate_strategy` in `steel-core/src/chunk/paletted_container.rs` hardcodes bits per entry for a global palette, `(6, PaletteMode::Global)` for biomes and `(15, PaletteMode::Global)` for block states (deemed "approximates")

Biome: 26.2 has 66 biomes, so the width should be 7. (checked by printing biome registry length after `init_vanilla_registry()`. With <64 in registry 6 was correct, anything over went stale.)
Branch is only reached when section's 4x4x4 biome container holds >8 distinct biomes. An imported world can contain this also.

Vanilla doesn't let the declared byte size the client's storage. A global palette config carries the registry-derived width separately, the client builds its bit storage from that, then fills it with a fixed size long array with no length prefix. So the client reads 8 longs for the 64 biome cells where Steel writes 7. Those 8 extra bytes come out of the next section, and every section after it is parsed from the wrong offset. An over-read past the end of the chunk buffer throws out of `replaceWithPacketData`, uncaught. Values are wrong even where the two lengths line up, since a global palette writes raw registry ids and the client unpacks on its own stride.

Block: correct for 26.2 at 32,366 states, breaks on the 26.3 bump. 26.3-rc-1's block report has 35,723 states, so the width becomes 16, and ids >32,767 now exist, which `pack_bits` shifts into the neighbouring entry rather than just misaligning it. Branch is reached when a section holds >256 distinct block states. 26.3 also adds dappled forest, biomes to 67, still 7.

Derive both widths from the registry at startup instead of hardcoding, so the 26.3 work doesn't have to remember this one.

Test a biome container with >8 distinct biomes and a block container with >256 distinct states, against expected bit count and long array length.

Mathildeuh:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Fixes #658. `PalettedContainer::calculate_strategy` hardcoded the global-palette bit
width: `15` for block states, `6` for biomes. Both are wrong today or will be soon:

- Biomes: 26.2 has 66 registered biomes, which need 7 bits (`2^6 = 64 < 66`). The
  hardcoded `6` under-sizes the fixed-size long array Steel writes; the client still
  builds its storage from the registry-derived width, over-reads into the next
  section's bytes, and every section after that is parsed from the wrong offset.
- Block states: `15` happens to be correct for 26.2's 32,366 states, but 26.3-rc-1 has
  35,723, which needs 16 bits and would silently misalign once ids exceed 32,767.

Both widths are now derived from the registry (`REGISTRY.blocks.next_state_id`,
`REGISTRY.biomes.len()`) via `ceil_log2`, matching vanilla's `Mth.ceillog2`, and cached
in a `OnceLock` on first use instead of being recomputed or hardcoded. A future version
bump that changes either count no longer needs a matching constant edited by hand.

## How this was tested

New tests in `chunk::paletted_container::tests`:

- `ceil_log2_matches_vanilla_worked_examples` - checks the formula against the issue's
  own worked numbers (64 -> 6, 65 -> 7, 32366 -> 15, 35723 -> 16).
- `block_global_palette_uses_registry_derived_bit_width_and_long_count` - a 4096-cell
  block container with every cell distinct (well past the 256-entry hashed-palette cap)
  writes the registry-derived bit count and the exact fixed-size long-array length for
  it (no length prefix, so a wrong length misaligns everything after it).
- `biome_global_palette_uses_registry_derived_bit_width_and_long_count` - same for a
  64-cell biome container past the 8-entry cap.

Confirmed against the live 26.2 registry (32366 block states, 66 biomes) that this
produces 15 and 7 respectively - 7 for biomes is the actual fix, 15 for blocks was
already correct and now stays correct automatically on the 26.3 bump.

`cargo test`, `cargo clippy -r --all-targets --all-features`, `cargo fmt --all --check`,
`typos`, and `prek run --all-files` all pass.

## Checklist

- [x] Self-reviewed the diff

## Additional notes

- `ceil_log2` uses the new `bit_width()` integer method (clippy's `manual_bit_width`
  pointed at it) instead of the `BITS - leading_zeros()` idiom used elsewhere in the
  codebase.


steel-core/src/chunk/paletted_container.rs
```diff
@@ -425,12 +427,35 @@ impl<V: Hash + Eq + Copy + Default + Debug, const DIM: usize> PalettedContainer<
...
+
+/// Bits per entry for the biome global palette. See [`block_state_global_bits`].
+fn biome_global_bits() -> u8 {
+    static BITS: OnceLock<u8> = OnceLock::new();
+    *BITS.get_or_init(|| ceil_log2(REGISTRY.biomes.len()))
+}
+
+/// Vanilla `Mth.ceillog2`: the number of bits needed to represent every id in
+/// `0..n`.
+const fn ceil_log2(n: usize) -> u8 {
```
Joniii11: just maybe put this into a steel-utils?

Joniii11 (approved):
else lgtm

steel-core/src/chunk/paletted_container.rs
CuzImClicks: yes our Mth functions go into steel-utils as pub const fns

Mathildeuh: Moved to `steel-utils/src/mth.rs` as `pub const fn ceil_log2` (matching the `block_util.rs` pattern - one file per vanilla utility class). `paletted_container.rs` now imports it from there instead of defining it locally.

steel-core/src/chunk/paletted_container.rs
```diff
@@ -425,12 +427,29 @@ impl<V: Hash + Eq + Copy + Default + Debug, const DIM: usize> PalettedContainer<
...
+fn block_state_global_bits() -> u8 {
+    static BITS: OnceLock<u8> = OnceLock::new();
+    *BITS.get_or_init(|| ceil_log2(usize::from(REGISTRY.blocks.next_state_id)))
+}
+
+/// Bits per entry for the biome global palette. See [`block_state_global_bits`].
+fn biome_global_bits() -> u8 {
+    static BITS: OnceLock<u8> = OnceLock::new();
+    *BITS.get_or_init(|| ceil_log2(REGISTRY.biomes.len()))
+}
```
coco875, line 451: why not make them atomicU8 ?

steel-core/src/chunk/paletted_container.rs
```diff
@@ -586,4 +606,64 @@ mod tests {
...
+        let mut next_id = 1u16;
+        for slab in &mut cube {
+            for row in slab.iter_mut() {
+                for value in row.iter_mut() {
+                    *value = BlockStateId(next_id);
+                    next_id = next_id.wrapping_add(1);
+                }
+            }
+        }
+        // All 4096 cells are distinct, well past the 256-entry hashed-palette cap.
```
coco875: does the comment are necessary ?

steel-core/src/chunk/paletted_container.rs
```diff
@@ -586,4 +606,64 @@ mod tests {
 
         assert_column_matches_get(&container, x, z);
     }
+
+    /// A container with more distinct values than any linear/hash palette holds
+    /// must use the registry-derived global bit width, and the fixed-size long
+    /// array it writes (no length prefix) must match that width exactly, or every
+    /// section after it is parsed from the wrong offset.
+    #[test]
+    fn block_global_palette_uses_registry_derived_bit_width_and_long_count() {
```
coco875: aren't the name of function and the body enouhh to understand ?

steel-core/src/chunk/paletted_container.rs
```diff
@@ -586,4 +606,64 @@ mod tests {
...
+        let mut next_id = 1u16;
+        for slab in &mut cube {
+            for row in slab.iter_mut() {
+                for value in row.iter_mut() {
+                    *value = next_id;
+                    next_id += 1;
+                }
+            }
+        }
+        // All 64 cells are distinct, well past the 8-entry hashed-palette cap.
```
coco875: does the comment are necessary ?

steel-utils/src/mth.rs
```diff
@@ -0,0 +1,27 @@
...
+        assert_eq!(ceil_log2(0), 0);
+        assert_eq!(ceil_log2(1), 0);
+        // 64 possible ids still fit in 6 bits (ids 0..=63).
+        assert_eq!(ceil_log2(64), 6);
+        // A 65th id needs a 7th bit, matching the issue's 66-biome regression.
+        assert_eq!(ceil_log2(65), 7);
+        // 26.2's block-state count needs 15 bits, matching the old hardcoded value.
+        assert_eq!(ceil_log2(32_366), 15);
+        // 26.3-rc-1's block-state count needs a 16th bit.
+        assert_eq!(ceil_log2(35_723), 16);
```
coco875, line 24: I think comment can be remove

coco875 (changes requested):

steel-core/src/chunk/paletted_container.rs
coco875: hmm maybe OnceLock<NonZeroU8> would be better so like that it allow the compiler to just check if it's different from 0 to know if it's have been init or not

Mathildeuh: OnceLock already handles that - get_or_init tracks init state on its own, doesn't need atomics or NonZeroU8 to know if it's run yet. Keeping it as OnceLock<u8>.

steel-core/src/chunk/paletted_container.rs
Mathildeuh: Trimmed it down, kept just the part that isn't obvious from the name (9a3fc409f).

steel-core/src/chunk/paletted_container.rs
Mathildeuh: Removed, yeah it was redundant with the loop right above it.

steel-core/src/chunk/paletted_container.rs
Mathildeuh: Same here.

steel-utils/src/mth.rs
Mathildeuh: Removed. Left the other 3 since they tie back to real numbers (the 66-biome bug, the old hardcoded value) - lmk if you'd rather those go too.

kdcokenny (approved):

steel-utils/src/mth.rs
```diff
@@ -0,0 +1,26 @@
...
+
+    #[test]
+    fn ceil_log2_matches_vanilla_worked_examples() {
+        assert_eq!(ceil_log2(0), 0);
+        assert_eq!(ceil_log2(1), 0);
+        // 64 possible ids still fit in 6 bits (ids 0..=63).
+        assert_eq!(ceil_log2(64), 6);
+        // A 65th id needs a 7th bit, matching the issue's 66-biome regression.
+        assert_eq!(ceil_log2(65), 7);
+        // 26.2's block-state count needs 15 bits, matching the old hardcoded value.
```
coco875: I also talk about other comment x)

Mathildeuh: Removed those too (bd24d763b)

coco875 (dismissed):

steel-core/src/chunk/paletted_container.rs
```diff
@@ -425,12 +427,29 @@ impl<V: Hash + Eq + Copy + Default + Debug, const DIM: usize> PalettedContainer<
...
 
+/// Bits per entry for the block-state global palette.
+///
+/// Derived from the registered block-state count and cached on first use, so a
+/// version bump that changes the state count doesn't need a matching hardcoded
+/// constant here. Matches vanilla, which sizes the global palette from the
+/// registry rather than a fixed width.
+fn block_state_global_bits() -> u8 {
+    static BITS: OnceLock<u8> = OnceLock::new();
+    *BITS.get_or_init(|| ceil_log2(usize::from(REGISTRY.blocks.next_state_id)))
```
CuzImClicks, line 444:
```suggestion
    ceil_log2(usize::from(REGISTRY.blocks.next_state_id))
```

registry is already in a OnceLock<Registry> and imo this just looks cleaner

CuzImClicks (changes requested):

CuzImClicks (approved):

coco875 (approved):

PR #665, opened 2026-09-11
feat(command): add /ban and /pardon

Mathildeuh:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [x] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Second PR in the moderation-commands batch started by #664 (`/kick`). This one
implements `/ban` and `/pardon`, matching vanilla's `BanPlayerCommands` and
`PardonCommand`. `/ban-ip`, `/pardon-ip`, `/banlist`, and `/whitelist` are
intentionally left for follow-up PRs - see #172 for why bundling all of this
together into one PR didn't work out last time. `/shadowban` isn't vanilla
and isn't part of this batch at all; it needs its own design discussion
first.

**Storage**: a new `steel_core::ban` module (`BanEntry`, `BanListConfig`,
`BanListManager`) mirrors the existing `permission::PermissionGroupManager`
pattern - an in-memory `SyncRwLock` snapshot with persistence-first updates
behind an `AsyncMutex`, and a `BanListStore` trait so `steel-core` stays
IO-free. `steel`'s `FileBanListStore` persists it to `banned-players.toml`
next to `config.toml`/`groups.toml`/`worlds.toml` - TOML to stay consistent
with the rest of Steel's own config, not vanilla's `banned-players.json`
(discussed and decided this way, rather than repeating the custom-format
concerns raised on #172).

**Commands**: `/ban <targets> [reason]` and `/pardon <targets>` both resolve
through the existing `GameProfileArgument` (`operator.rs`'s pattern), so
banning/pardoning works for offline players too, not just online ones.
`/ban` skips a target that's already banned (matching vanilla, so re-banning
doesn't clobber the original ban's timestamp/reason) and disconnects the
player immediately if they're currently online
(`multiplayer.disconnect.banned`, matching `BanPlayerCommands`'s live-kick).

**Login enforcement**: `steel-login`'s `finish_verified_login` now checks
`server.ban_list.find(uuid)` before completing login, rejecting with vanilla's
`multiplayer.disconnect.banned.reason` (+ `...expiration` if the ban isn't
permanent), matching `PlayerList.canPlayerLogin`. Without this, `/ban` would
just be a list with no actual effect on future connections.

Also adds a small `Server::online_player(uuid)` lookup (`player_lifecycle.rs`)
since nothing previously exposed a public online-player-by-uuid accessor
outside `server/`.

## How this was tested

- `ban::tests`: entry replacement on re-ban, expiry handling, remove
  reporting whether an entry existed, and a persistence-first round trip
  through a capturing/failing `BanListStore` (mirrors
  `permission::tests::manager`'s store-testing pattern).
- `command::builtins::ban::tests` / `pardon::tests`: command graph shape
  (targets required, `/ban`'s reason optional, correct argument types).

## Checklist

- [x] Self-reviewed the diff

- New: `/ban`, `/pardon`
- New: `steel_core::ban` (`BanEntry`, `BanListConfig`, `BanListManager`,
  `BanListStore`), `steel::config::FileBanListStore`
- New: `Server::ban_list`, `Server::online_player(uuid)`
- Changed: `Server::new` / `Server::new_with_commands` now also take a
  `BanListManager` (same shape as the existing `permission_groups` parameter)
- Changed: `steel-login`'s `finish_verified_login` now rejects banned UUIDs

## Additional notes

- `chrono`'s `serde` feature is now enabled workspace-wide (needed to
  (de)serialize `BanEntry.created`/`expires` to/from `banned-players.toml`).
- This branch is stacked on #664 (`/kick`), since `/ban`'s reason argument
  reuses the `message` argument type added there. The diff currently includes
  #664's commit too - it'll shrink to just this PR's changes automatically
  once #664 merges into `master`.

### DIFF
--- a/steel-core/src/ban.rs
+++ b/steel-core/src/ban.rs
@@ -31,9 +31,10 @@ pub struct BanEntry {
     /// When the ban expires. `None` means it never expires.
     #[serde(default)]
     pub expires: Option<DateTime<Utc>>,
-    /// The ban reason, if any.
+    /// The ban reason, if any. Supports rich text (colors, hover events, ...)
+    /// via SNBT, matching a plain `TextComponent::plain` for ordinary text.
     #[serde(default)]
-    pub reason: Option<String>,
+    pub reason: Option<TextComponent>,
 }
 
 impl BanEntry {
@@ -43,14 +44,13 @@ impl BanEntry {
         self.expires.is_some_and(|expires| expires <= Utc::now())
     }
 
-    /// Returns vanilla's `BanListEntry.getReasonMessage`: the literal reason,
+    /// Returns vanilla's `BanListEntry.getReasonMessage`: the given reason,
     /// or a translated default when none was given.
     #[must_use]
     pub fn reason_message(&self) -> TextComponent {
-        self.reason.as_deref().map_or_else(
-            || TextComponent::from(&translations::MULTIPLAYER_DISCONNECT_BANNED_REASON_DEFAULT),
-            |reason| TextComponent::plain(reason.to_owned()),
-        )
+        self.reason.clone().unwrap_or_else(|| {
+            TextComponent::from(&translations::MULTIPLAYER_DISCONNECT_BANNED_REASON_DEFAULT)
+        })
     }
 
     /// Builds vanilla's `PlayerList.canPlayerLogin` rejection message for this
@@ -244,7 +244,10 @@ mod tests {
     use steel_utils::locks::SyncMutex;
     use uuid::Uuid;
 
-    use super::{BanEntry, BanListConfig, BanListManager, BanListStore, BanListStoreError, Utc};
+    use super::{
+        BanEntry, BanListConfig, BanListManager, BanListStore, BanListStoreError, TextComponent,
+        Utc,
+    };
 
     #[derive(Debug)]
     struct CapturingStore {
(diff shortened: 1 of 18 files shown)


steel-core/src/ban.rs
```diff
@@ -0,0 +1,362 @@
...
+    /// When the ban was created.
+    pub created: DateTime<Utc>,
+    /// Who (or what) created the ban, e.g. a command sender's name.
+    pub source: String,
+    /// When the ban expires. `None` means it never expires.
+    #[serde(default)]
+    pub expires: Option<DateTime<Utc>>,
+    /// The ban reason, if any.
+    #[serde(default)]
+    pub reason: Option<String>,
```
JunkyDeveloper: Please Support TextComponent like I did #172, the concept was also not perfect, and because of outdating and no time it was closed

JunkyDeveloper (changes requested):

steel-core/src/ban.rs
Mathildeuh: Done - `BanEntry.reason` is now `Option<TextComponent>`. `/ban`'s reason argument tries `TextComponent::from_snbt` first and falls back to plain text if it doesn't parse, same approach you used in #172.

PR #669, opened 2026-09-11
feat(command): add /kick, /ban, /pardon, and /whitelist

Mathildeuh:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [x] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Consolidates #664, #665, and #668 into one PR, per feedback to group
player-facing moderation commands together and IP-facing ones separately
(see the follow-up PR for the IP-side commands). Implements `/kick`,
`/ban`, `/pardon`, and the full `/whitelist` command.

**`/kick`** matches vanilla's `KickCommand`: disconnects targeted online
players with an optional reason. Adds a `message` command argument type
(greedy rest-of-line text) for the reason, reused by `/ban` and (in the
follow-up PR) `/ban-ip`.

**`/ban` and `/pardon`** match `BanPlayerCommands`/`PardonCommand`: targets
resolve through the existing `GameProfileArgument` (works offline too), bans
persist to `banned-players.toml` via a new `steel_core::ban` module
(`BanEntry`, `BanListManager`) following the same persistence-first pattern
as `permission::PermissionGroupManager`, and an already-connected banned
player is disconnected immediately. The ban reason is stored as a
`TextComponent` (parsed from SNBT when possible, falling back to plain
text), not a plain string, so operators can use formatted/colored reasons -
same approach as #172, per review feedback on the original #665.

**`/whitelist`** implements the full vanilla command: `on`/`off`/`list`/
`add`/`remove`/`reload`, in a new `steel_core::whitelist` module. Unlike the
ban list, `WhitelistStore` also supports loading (not just saving), since
`/whitelist reload` needs to pick up an operator hand-editing
`whitelist.toml` while the server is running - a real vanilla use case the
ban commands don't have an equivalent of. `/whitelist on`, `remove`, and
`reload` disconnect currently-connected players who are no longer
whitelisted, matching `kickUnlistedPlayers` - the first command here that
needs to act on every online player, via a new
`Server::online_players_snapshot()`.

**Login enforcement**: `steel-login`'s `finish_verified_login` now checks
the ban list and the whitelist, in vanilla's `canPlayerLogin` order (IP ban
check follows in the IP-side PR, after that).

`/ban-ip`, `/pardon-ip`, and `/banlist` are in a separate follow-up PR since
`/banlist` needs both ban lists to exist and IP bans are otherwise
independent of everything here.

## How this was tested

- `ban::tests`: entry replacement on re-ban, expiry handling, remove
  reporting whether an entry existed, and a persistence-first round trip
  through a capturing/failing `BanListStore`.
- `whitelist::tests`: add skips an already-whitelisted player, remove
  reports whether an entry existed, `set_enabled` persists and updates
  state, `reload` replaces in-memory state from the store, and a
  persistence-first round trip through a capturing/failing `WhitelistStore`.
- `command::builtins::{kick,ban,pardon,whitelist}::tests`: command graph
  shape for each (required/optional arguments, correct argument types).

## Checklist

- [ ] Self-reviewed the diff

- New: `/kick`, `/ban`, `/pardon`, `/whitelist` (`on`/`off`/`list`/`add`/
  `remove`/`reload`)
- New: `steel_core::ban` (`BanEntry`, `BanListConfig`, `BanListManager`,
  `BanListStore`), `steel_core::whitelist` (`WhitelistEntry`,
  `WhitelistConfig`, `WhitelistManager`, `WhitelistStore`)
- New: `steel::config::{FileBanListStore, FileWhitelistStore}`
- New: `Server::ban_list`, `Server::whitelist`, `Server::online_player(uuid)`,
  `Server::online_players_snapshot()`
- New: `SteelArgumentType::message()` / `MessageParser` / `MessageValue`
- Changed: `Server::new` / `Server::new_with_commands` now also take a
  `BanListManager` and a `WhitelistManager`
- Changed: `steel-login`'s `finish_verified_login` now rejects banned and
  (while enabled) non-whitelisted connections

## Additional notes

- `chrono`'s `serde` feature is now enabled workspace-wide (needed to
  (de)serialize `BanEntry.created`/`expires` to/from `banned-players.toml`).
- Replaces #664, #665, and #668, which are being closed in favor of this
  consolidated PR.

### DIFF
--- a/steel-core/src/command/execution/argument.rs
+++ b/steel-core/src/command/execution/argument.rs
@@ -335,6 +335,10 @@ impl SteelArgumentType {
         Self::new(ComponentParser)
     }
 
+    pub(crate) fn message() -> Self {
+        Self::new(MessageParser)
+    }
+
     pub(crate) fn nbt_path() -> Self {
         Self::new(NbtPathParser)
     }
@@ -533,6 +537,7 @@ argument_value_wrapper!(
     ComponentValue(TextComponent),
     "steel:command/value/component"
 );
+argument_value_wrapper!(MessageValue(Box<str>), "steel:command/value/message");
 argument_value_wrapper!(NbtPathValue(NbtPath), "steel:command/value/nbt_path");
 argument_value_wrapper!(
     IdentifierValue(Identifier),
@@ -1097,6 +1102,16 @@ unit_argument_parser!(
     suggest | _context,
     _builder | {},
     protocol(ProtocolArgumentType::Component, None)
+);
+unit_argument_parser!(
+    MessageParser,
+    "steel:command/parser/message",
+    MessageValue,
+    parse | reader,
+    _source | { Ok(MessageValue(reader.read_remaining().into())) },
+    suggest | _context,
+    _builder | {},
+    protocol(ProtocolArgumentType::Message, None)
 );
 unit_argument_parser!(
     NbtPathParser,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -259,6 +259,7 @@ dependencies = [
  "iana-time-zone",
  "js-sys",
  "num-traits",
+ "serde",
  "wasm-bindgen",
  "windows-link",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,7 +77,7 @@ small-map = "0.1.5"
 smallvec = "1.15.2"
 bitflags = "2.13.1"
 phf = { version = "0.14.0", features = ["macros"] }
-chrono = "0.4.45"
+chrono = { version = "0.4.45", features = ["serde"] }
 
 # UUID
 uuid = { version = "1.24.0", features = ["serde", "v4"] }
--- a/steel-core/src/ban.rs
+++ b/steel-core/src/ban.rs
@@ -1,0 +1,368 @@
+//! Persisted player ban list.
+//!
+//! Mirrors vanilla's `UserBanList`, but stored as Steel's own
+//! `banned-players.toml` for consistency with the rest of Steel's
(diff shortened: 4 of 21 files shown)


steel-core/src/command/execution/argument.rs
```diff
@@ -1098,6 +1103,16 @@ unit_argument_parser!(
...
+unit_argument_parser!(
+    MessageParser,
+    "steel:command/parser/message",
+    MessageValue,
+    parse | reader,
+    _source | { Ok(MessageValue(reader.read_remaining().into())) },
+    suggest | _context,
+    _builder | {},
+    protocol(ProtocolArgumentType::Message, None)
+);
```
Laptop59, line 1215: This does not do what Vanilla does for `MessageArgument`. It parses selectors and the result can be resolved with chat signatures.

Laptop59 (changes requested):
I actually tested the `/ban` command with a reason from an online player (not from the terminal). I get the following in the server console:

> play packet id 8 is not known

Yes, you need to handle the `SChatCommandSigned` packet (looks like the properties are already implemented) to actually get it to work properly.

steel-core/src/command/execution/argument.rs
Mathildeuh: Fixed in 941a88385 - now scans for @selectors like vanilla's Message#parseText does, resolved at use time through the same CommandTextResolver /tellraw uses. Signatures still aren't a thing since we don't have chat reporting at all.

PR #670, opened 2026-09-11
feat(command): add /ban-ip, /pardon-ip, and /banlist

Mathildeuh:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [x] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

IP-side counterpart to #669, split out per feedback to group player-facing
moderation commands separately from IP-facing ones. Implements `/ban-ip`,
`/pardon-ip`, and `/banlist`. Replaces the old #666 and #667.

**`/ban-ip` and `/pardon-ip`** match `BanIpCommands`/`PardonIpCommand`: a
new `IpBanEntry`/`IpBanListManager` (same file and persistence-first pattern
as #669's player ban list) persists to `banned-ips.toml`, enforced at login
via the client's socket address. Unlike vanilla, the target must be a
literal IP: Steel doesn't track a connected player's remote address past
login (the `NetworkConnection` trait has no such method, and ~9 call sites
including every test mock would need updating to add it), so resolving a
player name to their current IP - and kicking every other player sharing a
freshly banned IP - isn't implemented here. Reason handling matches #669's
`/ban`: a `TextComponent`, parsed from SNBT when possible.

**`/banlist`** matches `BanListCommands`: bare (players + IPs combined),
`players`, and `ips`, each listing display name, source, and reason. It's
grouped here rather than in #669 because it needs both ban lists to exist,
and `IpBanListManager` is introduced in this PR.

Also adds a generic `word` command argument type (vanilla's
`StringArgumentType.word()`) for `/ban-ip`'s and `/pardon-ip`'s target,
which isn't a `GameProfileArgument` like the player commands - it's
validated as an IP address in the command body, matching vanilla's own
structure.

Depends on #669 (`BanListManager`, the `message` argument type, and the
login-enforcement pattern this PR extends) - based on that branch, so the
diff includes its commits until it merges.

## How this was tested

- `ban::tests` (`ip_*`): entry replacement on re-ban, expiry handling, and
  remove reporting whether an entry existed - same coverage shape as #669's
  player ban list tests.
- `command::builtins::{ban_ip,pardon_ip,banlist}::tests`: command graph
  shape for each.

## Checklist

- [ ] Self-reviewed the diff

- New: `/ban-ip`, `/pardon-ip`, `/banlist` (bare, `players`, `ips`)
- New: `steel_core::ban::{IpBanEntry, IpBanListConfig, IpBanListManager,
  IpBanListStore}`, `steel::config::FileIpBanListStore`
- New: `Server::ip_ban_list`
- New: `SteelArgumentType::word()` / `WordParser` / `WordValue`
- Changed: `Server::new` / `Server::new_with_commands` now also take an
  `IpBanListManager`
- Changed: `steel-login`'s `finish_verified_login` now also rejects banned
  IPs

## Additional notes

- Replaces #666 and #667, which are being closed in favor of this
  consolidated PR.
- Tracking a connected player's remote address on `Player`/`PlayerConnection`
  (needed for vanilla-parity `/ban-ip <name>` and kicking everyone sharing a
  banned IP) would need its own PR: it touches the `NetworkConnection` trait
  used by every test mock in the codebase, not just the real connection.

### DIFF
--- a/steel-core/src/command/execution/argument.rs
+++ b/steel-core/src/command/execution/argument.rs
@@ -335,6 +335,10 @@ impl SteelArgumentType {
         Self::new(ComponentParser)
     }
 
+    pub(crate) fn message() -> Self {
+        Self::new(MessageParser)
+    }
+
     pub(crate) fn nbt_path() -> Self {
         Self::new(NbtPathParser)
     }
@@ -533,6 +537,7 @@ argument_value_wrapper!(
     ComponentValue(TextComponent),
     "steel:command/value/component"
 );
+argument_value_wrapper!(MessageValue(Box<str>), "steel:command/value/message");
 argument_value_wrapper!(NbtPathValue(NbtPath), "steel:command/value/nbt_path");
 argument_value_wrapper!(
     IdentifierValue(Identifier),
@@ -1097,6 +1102,16 @@ unit_argument_parser!(
     suggest | _context,
     _builder | {},
     protocol(ProtocolArgumentType::Component, None)
+);
+unit_argument_parser!(
+    MessageParser,
+    "steel:command/parser/message",
+    MessageValue,
+    parse | reader,
+    _source | { Ok(MessageValue(reader.read_remaining().into())) },
+    suggest | _context,
+    _builder | {},
+    protocol(ProtocolArgumentType::Message, None)
 );
 unit_argument_parser!(
     NbtPathParser,
--- a/steel-core/src/command/execution/argument.rs
+++ b/steel-core/src/command/execution/argument.rs
@@ -30,7 +30,8 @@ use crate::command::protocol::protocol_argument_type;
 use crate::entity::{ENTITIES, EntityAnchor};
 use glam::DVec3;
 use steel_protocol::packets::game::{
-    ArgumentType as ProtocolArgumentType, SuggestionType as ProtocolSuggestionType,
+    ArgumentStringTypeBehavior, ArgumentType as ProtocolArgumentType,
+    SuggestionType as ProtocolSuggestionType,
 };
 use steel_registry::damage_type::DamageTypeRef;
 use steel_registry::{
@@ -339,6 +340,10 @@ impl SteelArgumentType {
         Self::new(MessageParser)
     }
 
+    pub(crate) fn word() -> Self {
+        Self::new(WordParser)
+    }
+
     pub(crate) fn nbt_path() -> Self {
         Self::new(NbtPathParser)
     }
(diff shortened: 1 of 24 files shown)


steel-core/src/command/execution/argument.rs
```diff
@@ -1098,6 +1109,31 @@ unit_argument_parser!(
...
+    MessageParser,
+    "steel:command/parser/message",
+    MessageValue,
+    parse | reader,
+    _source | { Ok(MessageValue(reader.read_remaining().into())) },
+    suggest | _context,
+    _builder | {},
+    protocol(ProtocolArgumentType::Message, None)
+);
+unit_argument_parser!(
```
Laptop59, line 1222: We already have `SteelArgument::word()` so why do we need this?

steel-core/src/command/execution/argument.rs
```diff
@@ -1098,6 +1109,31 @@ unit_argument_parser!(
...
+    MessageParser,
+    "steel:command/parser/message",
+    MessageValue,
+    parse | reader,
+    _source | { Ok(MessageValue(reader.read_remaining().into())) },
+    suggest | _context,
+    _builder | {},
+    protocol(ProtocolArgumentType::Message, None)
+);
```
Laptop59, line 1221: This does not do what Vanilla does for `MessageArgument`. It parses selectors and the result can be resolved with chat signatures.

Laptop59 (changes requested):
- It looks like your PR implements `/ban`, `/pardon`, `/kick` and `/whitelist` which is not given in the title.
- It's better not to make the same parsers in two different PRs. 

I actually tested the `/ban-ip` command with a reason from a player (not from the terminal). I get the following in the server console:

> play packet id 8 is not known

Yes, you need to handle a new packet to actually get it to work properly.

steel-core/src/command/execution/argument.rs
Mathildeuh: Same fix as on #669 (941a88385) - that's where MessageParser lives. Will land here once this branch is rebased on it.

steel-core/src/command/execution/argument.rs
Mathildeuh: Couldn't find an existing one before adding this - word() is new here. message is greedy so it eats everything to the end, this one needs to stop at whitespace instead. Let me know if you meant something else.

Laptop59: Yeah, you're right, sorry about that.

PR #671, opened 2026-09-11
Fix/devshell rust src

ezraclintoc:
## Type of change

- [x] Chore / tooling

## Description

I kept getting a rust-analyzer sysroot error ("can't load standard library") and asked for a fix. That led to finding a few more bugs in `flake.nix` while I was in there:

1. devShell toolchain had no `rust-src`, so rust-analyzer couldn't resolve std. Actual builds were fine since rustc doesn't need it, only rust-analyzer does. Added it via a devShell-only override so the package build stays lean.
2. `nix build` failed. `steel-utils`'s build script now needs `deprecated.json` too, not just `en_us.json`. The flake never staged it, so the sandbox tried to fetch over the network and died.
3. `nix flake check --all-systems` errored on `aarch64-darwin`. `packages` wasn't scoped to Linux like `checks` already was, so it tried to evaluate a package that refuses to build there.
4. The Nix package reported the wrong version. `nix/minecraft-assets.json` was still on `0.15.2` after Cargo bumped to `0.15.3`. Regenerated it.
5. `update-minecraft-assets.sh`'s comment claimed CI checks this pin and that it can't drift. Neither is true, so fixed the comment instead of adding a CI job (already decided against that elsewhere).

## How this was tested

- Reverted each fix, reproduced the original failure, restored it, confirmed before/after for the rust-src and version-pin bugs.
- `nix build` from scratch produces a binary that actually boots.
- `nix flake check --all-systems` and `nix fmt -- --fail-on-change` both pass clean.
- Had a second pass specifically try to find more problems in `flake.nix`; it found #4.

## Checklist

- [x] Self-reviewed the diff

## Additional notes

This PR was AI-generated, but reviewed by me.


JunkyDeveloper (approved):
Was tested by another user

coco875 (approved):

Issue #571, opened 2026-08-31
DamageSource loses entity refs after removal


kdcokenny:
found this while looking through #428.

right now `DamageSource` stores weak `EntityReference`s. if a TNT or projectile gets removed and nothing else holds its `Arc`, `direct_entity()` and `causing_entity()` stop resolving. vanilla keeps the actual entity on `DamageSource`, so this doesnt match. I reproduced it with a test that removes the entity and drops the last `Arc`.

just making the refs strong isnt safe either. since `LivingEntity` stores `last_damage_source`, self damage could make this:

`entity -> damage source -> same entity`

2 entities damaging each other could do the same thing. single threaded doesnt change this since its not a race, the refs keep each other alive after the logic finishes.

my current thought is to keep normal `EntityReference`s weak, make `DamageSource` hold the exact entities, then move `last_damage_source` outside the entity into a world owned table. it would stay available through vanilla's 40 tick window and use a unique entity instance id so a reload cant resolve to a different entity.

this probably belongs in its own PR since it changes how damage refs are stored. #428 can depend on it. thoughts?

blocked by #573.

PR #675, opened 2026-09-12
fix: share game time across worlds in each domain

### LINKED ISSUE #662: share game time across dimensions in the same domain
steel currently saves and advances `game_time` separately for each world. vanilla 26.2 shares it across dimensions.

this causes problems when an entity changes worlds. `last_damage_source()` compares the current world's time against the original damage timestamp, so moving between worlds with different times can expire the record immediately or keep it around too long.

we should share game time within each domain and advance it once per simulation tick. daylight clocks can stay independent, so worlds can still have different time settings.

this needs to cover saving/loading, scheduled ticks, time packets, and other code using `game_time()`. we also need to decide how timestamps work when moving between separate domains. test normal dimension transfers, reloads, and timing with multiple worlds loaded, including fabric client gametests.

blocks #571. the damage history getter and cleanup need to agree on elapsed time before we change ownership.

kdcokenny:
## Type of change

- [x] Bug fix

## Description

fixes #662

worlds in the same domain now share one `game_time` counter. only the configured default world loads and saves it in `level.toml`. other worlds ignore their old value and remove the field when saved.

the server advances each domain once before dispatching world workers. frozen ticks don't advance it, permitted steps do. daylight clocks, rates and pauses stay separate for each world.

## How this was tested

checked against Minecraft 26.2 release sources.

ran Fabric 26.2 client GameTests against Steel before and after restart. checked dimension transfers, time synchronization, `/tick freeze`, `/tick step`, and water staying still while frozen then spreading after five steps.

## Screenshots / logs

from the client run after restart:

```text
PASS frozen iterations = 4763
PASS seven simulation steps with multiple worlds = 4770
PASS dimension transfer shares primary clock = 4770
PASS derived world steps = 4773
PASS return to primary = 4773
PASS scheduled water spreads after five permitted steps
BUILD SUCCESSFUL in 24s
```

## Checklist

- [x] Self-reviewed the diff

## Additional notes

same domain transfers keep the last damage record and its original timestamp. switching domains clears that record when the target player state is restored, not when the switch is requested. The history redesign in #571 stays out of scope.

changing the configured default to a previously derived world requires explicitly transferring the saved time. a missing primary `game_time` fails loading instead of silently resetting it.

a domain with persistent worlds now requires a default world that persists level data. startup rejects a RAM default with disk siblings before loading or rewriting any world saves. all-RAM domains remain supported.

legacy absolute timestamps, including interaction records, are not migrated when derived worlds adopt the primary clock. their elapsed ages may change when upgrading old saves with divergent clocks. this follows the project’s no-migrations policy.


steel-core/src/chunk_saver/registry.rs
```diff
@@ -143,3 +186,44 @@ fn create_ram_storage(
...
+[[domains.example.worlds]]
+name = "primary"
+default = true
+generator = "minecraft:flat"
+storage = {{ type = "steel:{primary}" }}
+[[domains.example.worlds]]
+name = "derived"
+generator = "minecraft:flat"
+storage = {{ type = "steel:{derived}" }}
+"#
```
coco875: why just define the struct directly instead of parse it

steel-core/src/level_data/game_time.rs
```diff
@@ -0,0 +1,40 @@
...
+        }
+    }
+
+    /// Returns the domain's current simulation time.
+    pub fn ticks(&self) -> i64 {
+        self.ticks.load(Ordering::Relaxed)
+    }
+
+    pub(super) fn advance(&self) {
+        // Server coordination completes all increments before dispatching workers.
```
coco875: move the comment above ?

steel-core/src/level_data/game_time_tests.rs
```diff
@@ -0,0 +1,129 @@
...
+    fs::write(
+        path.join("level.toml"),
+        toml::to_string(&data).expect("serialize table"),
+    )
+    .await
+    .expect("write fixture");
+}
+
+#[tokio::test]
+async fn game_time_primary_only_survives_both_shutdown_save_orders() {
```
coco875, line 35: maybe put some space in code

steel-core/src/player/lifecycle/spawn_sync.rs
```diff
@@ -23,7 +23,11 @@ impl Player {
     ) where
         F: FnOnce(),
     {
-        self.reset_inner_after(new_world, ResetReason::WorldChange, true, restore_state);
+        self.reset_inner_after(new_world, ResetReason::WorldChange, true, || {
+            restore_state();
+            // Damage history is transient source-domain state, not part of the restored save.
```
coco875: not sure the comment are useful

steel-core/src/server/world_tick_workers.rs
```diff
@@ -125,30 +125,70 @@ impl WorldTickWorkers {
...
+        let worlds = test_domain("delayed", &["primary", "derived"]);
+        let primary = worlds.default_world("delayed").expect("primary");
+        let derived = worlds
+            .get(&steel_utils::Identifier::new_static("delayed", "derived"))
+            .expect("derived");
+        let workers = WorldTickWorkers::spawn([primary, derived]).expect("workers");
+        for tick in 1..=2 {
+            worlds.advance_domain_game_times();
+            // Holding primary level data delays its world-local time phase, while
+            // the shared counter remains readable without this lock.
```
coco875, line 172: does the comment are useful ?

coco875 (changes requested):
I have only check comment mostly

steel-core/src/server/world_tick_workers.rs
kdcokenny: i'd keep this one, it explains why we're holding the primary's lock while waiting for the other worker. otherwise it looks like an accidental lock being held

steel-core/src/player/lifecycle/spawn_sync.rs
kdcokenny: I shortened it but i think its worth explaining why we clear it here, the damage timestamps belong to the old domain's clock

steel-core/tests/support/mod.rs
```diff
@@ -481,3 +493,66 @@ impl LevelAccessor for TestLevel {
...
+    };
+    let mut worlds = WorldMap::new(world.domain().to_owned(), &[config], &[]);
+    worlds.insert(world.key.clone(), Arc::clone(world));
+    worlds
+        .validate_game_times()
+        .expect("single-world fixture must own its clock");
+    for _ in 0..ticks {
+        worlds.advance_domain_game_times();
+    }
+}
```
CuzImClicks:
```suggestion
pub(crate) fn advance_test_game_time_to(world: &World, target: i64) {
    assert!(target >= world.game_time(), "test clock cannot move backward");
    let mut level_data = world.level_data.write();
    while world.game_time() < target {
        level_data.advance_game_time();
    }
}
```

CuzImClicks (changes requested):
one quick change then i can approve

steel-core/tests/support/mod.rs
kdcokenny: applied it :)

CuzImClicks (dismissed):

steel-core/src/entity/living_base/mod.rs
```diff
@@ -42,6 +42,8 @@ const MIN_EFFECT_AMPLIFIER: i32 = 0;
 const MAX_EFFECT_AMPLIFIER: i32 = 255;
 const SPRINT_SPEED_MODIFIER_AMOUNT: f64 = 0.3;
 const POST_IMPULSE_GRACE_TICKS: i32 = 40;
+/// Vanilla `LivingEntity.DAMAGE_SOURCE_TIMEOUT`, measured in game ticks.
```
coco875: reference vanilla are not useful because it's the same name literally x)

steel-core/src/entity/tests/damage.rs
```diff
@@ -14,6 +14,7 @@ fn generic_living_hurt_applies_health_damage() {
 #[test]
 fn generic_living_hurt_ignores_fire_damage_with_fire_resistance() {
     init_vanilla_registry();
+    init_behaviors();
```
coco875, line 17: init_behaviors are not already init in init_vanilla_registry ? and look like the test was passing before ? so why it's needed now ?

steel-core/src/server/tests/game_time.rs
```diff
@@ -0,0 +1,390 @@
...
+generator = "minecraft:flat"
+[[domains.custom.worlds]]
+name = "authority"
+generator = "minecraft:flat"
+default = true
+[domains.custom.worlds.config]
+dimension_type = "minecraft:the_end"
+"#,
+                root.display()
+            );
```
coco875: why not just define the struct directly, instead of parsing

steel-core/src/server/tests/game_time.rs
```diff
@@ -0,0 +1,390 @@
...
+generator = "minecraft:flat"
+storage = {{ type = "steel:disk" }}
+[[domains.custom.worlds]]
+name = "lobby"
+generator = "minecraft:flat"
+default = true
+storage = {{ type = "steel:ram" }}
+"#,
+                root.display()
+            );
```
coco875: why not just define the struct directly, instead of parsing

coco875 (changes requested):

steel-core/src/entity/tests/damage.rs
kdcokenny: so removing init_behaviors() makes them panic with "Mob effect behaviors not initialized." They pass standard CI/CD bc that has other tests initialize the behaviors first, which masks this issue.

steel-core/src/server/tests/game_time.rs
```diff
@@ -201,23 +203,49 @@ fn game_time_startup_and_chunk_reload_use_the_configured_primary() {
...
+                                        .into_iter()
+                                        .collect(),
+                                    )
+                                }),
+                            })
+                            .collect(),
+                    },
+                )]
+                .into(),
+            };
```
coco875: hmm can be simplified with default no ? to avoid define each field

steel-core/src/server/tests/game_time.rs
```diff
@@ -341,31 +369,51 @@ fn game_time_rejects_ephemeral_primary_before_touching_derived_save() {
...
+                                }),
+                                nether_portal_target: None,
+                                end_portal_target: None,
+                                config: None,
+                            })
+                            .collect(),
+                    },
+                )]
+                .into(),
+            };
```
coco875: hmm can be simplified with default no ? to avoid define each field

coco875 (changes requested):

coco875 (approved):

steel-core/src/server/tests/game_time.rs
kdcokenny: moved into helper

steel-core/src/server/tests/game_time.rs
kdcokenny: moved into helper

CuzImClicks (approved):

PR #680, opened 2026-09-13
feat(command): add /version

Mathildeuh:
## Type of change

- [x] Command implementation
- [x] New feature

## Description

Implements `/version`, matching vanilla's `VersionCommand`: reports the id, name, protocol number, and stability of the targeted vanilla version, op-only by default (matches vanilla's `checkPermissions ? LEVEL_GAMEMASTERS : LEVEL_ALL`).

Vanilla also reports a save-data version, a save-data version series, a build time, and resource/data pack versions - Steel doesn't track any of that anywhere yet, so those lines are left out rather than filled in with made-up numbers. Happy to add them once/if that data becomes available elsewhere in the codebase.

Also adds `CommandSource::send_system_message`, distinct from `send_success`: vanilla's `/version` sends its output via `sendSystemMessage`, which always reaches the sender regardless of the `sendCommandFeedback` game rule and never broadcasts to admins - unlike ordinary command-result feedback. This is a small, reusable addition for future commands with similar always-visible informational output.

## How this was tested

- `command::builtins::version::tests::version_is_a_bare_executable_literal`: command graph shape.
- Updated `first_builtin_slice_has_the_expected_graph_shape` to include the new root.

`cargo test`, `cargo clippy -r --all-targets --all-features`, `cargo fmt --all --check`, `typos`, and `prek run --all-files` all pass.

```
/version
```

## Checklist

- [x] Self-reviewed the diff

- New: `/version`
- New: `CommandSource::send_system_message`


steel-core/src/command/builtins/version.rs
```diff
@@ -0,0 +1,89 @@
...
+/// Reports the id, name, protocol, and stability of the targeted vanilla
+/// version. Vanilla's `/version` also reports a save-data version, a
+/// save-data version series, a build time, and resource/data pack versions -
+/// Steel doesn't track any of those yet, so those lines are left out rather
+/// than filled in with made-up numbers.
+#[expect(
+    clippy::unnecessary_wraps,
+    reason = "Command executors use a shared fallible callback signature."
+)]
+fn send_version(context: &SteelCommandContext<CommandSource>) -> Result<i32, CommandSyntaxError> {
```
SomeYellowGuy, line 29: The system messages look incomplete, though I'm not sure what the best way of storing shared constants is.

steel-core/src/command/builtins/version.rs
```diff
@@ -0,0 +1,89 @@
...
+        &translations::COMMANDS_VERSION_PROTOCOL
+            .message([
+                CURRENT_MC_PROTOCOL.to_string(),
+                format!("0x{CURRENT_MC_PROTOCOL:x}"),
+            ])
+            .component(),
+    );
+    // Steel only ever targets full releases of MC_VERSION, never a
+    // pre-release build of it (e.g. "26.3-rc-1").
+    let stable = if MC_VERSION.contains('-') {
```
SomeYellowGuy: We shouldn't assume that a dash means the version isn't stable. I think it would be a good idea to update the extractor to extract all the version info for the server.

SomeYellowGuy (changes requested):
This command is a bit more complicated to implement than it seems.

steel-core/src/command/builtins/version.rs
SomeYellowGuy: Speaking of shared constants, there is a generated `version` file in `steel_utils`, so maybe use that as the source of version info?

Mathildeuh: Fixed in 09c7f2dfd - build script now pulls version.json out of the server jar too, so all the fields are there now.

steel-core/src/command/builtins/version.rs
Mathildeuh: Yeah, fair. Replaced with the real stable flag from version.json (same commit).

steel-utils/build/build.rs
```diff
@@ -335,16 +394,59 @@ fn download_and_extract_assets(manifest_dir: &str) {
...
     println!(
         "cargo:warning=Successfully extracted datapack and translation files for Minecraft {target_ver}."
     );
 }
 
-fn build_version_constant() -> String {
+/// Generates the version constants from the `version.json` extracted from
+/// the target server jar, mirroring `DetectedVersion.tryDetectVersion`.
+fn build_version_constants(manifest_dir: &str) -> String {
     let target_version = get_target_mc_version();
```
CuzImClicks, line 410:
```suggestion
    println!("cargo:rerun-if-changed=build_assets/version.json");
    let target_version = get_target_mc_version();
```

steel-utils/build/build.rs
```diff
@@ -335,16 +394,59 @@ fn download_and_extract_assets(manifest_dir: &str) {
...
     fs::write(&version_file, &target_ver).expect("Failed to write version file");
     println!(
         "cargo:warning=Successfully extracted datapack and translation files for Minecraft {target_ver}."
     );
 }
 
-fn build_version_constant() -> String {
+/// Generates the version constants from the `version.json` extracted from
+/// the target server jar, mirroring `DetectedVersion.tryDetectVersion`.
+fn build_version_constants(manifest_dir: &str) -> String {
```
CuzImClicks, line 409: vanilla:

Image

steel:

Image

our build_time formatting looks worse

steel-core/src/command/builtins/version.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+fn command() -> CommandNodeBuilder<CommandSource, SteelCommandRuntime> {
+    literal("version").executes(send_version)
+}
+
+/// Reports the targeted vanilla version the way `VersionCommand#dumpVersion`
+/// does, using the `version.json` the build script extracts from the target
+/// server jar for everything but the protocol number (already tracked by
+/// `steel_registry::packets::CURRENT_MC_PROTOCOL`).
```
CuzImClicks, line 24: shorten comment

CuzImClicks (changes requested):

chocodev11:
suggestion: add commit id of the build

CuzImClicks:
image

PR #683, opened 2026-09-13
Add stop sound command

KalSupe:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [x] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Adds the vanilla /stopsound command for stopping sounds for players.

The command supports stopping all sounds, stopping sounds from a specific sound source, and stopping a specific sound.

Also adds the CStopSound clientbound packet required to send the stop-sound instruction to the client.

## How this was tested

Tested on Minecraft 26.2.

Examples tested:

/stopsound @s
/stopsound @s master
/stopsound @s master minecraft:music.game

Verified that sounds played with /playsound are correctly stopped by the corresponding /stopsound commands.

## Checklist

- [x] Self-reviewed the diff

/stopsound


steel-core/src/command/builtins/stopsound.rs
```diff
@@ -0,0 +1,104 @@
...
+    SoundSource::Records,
+    SoundSource::Weather,
+    SoundSource::Blocks,
+    SoundSource::Hostile,
+    SoundSource::Neutral,
+    SoundSource::Players,
+    SoundSource::Ambient,
+    SoundSource::Voice,
+    SoundSource::Ui,
+];
```
Laptop59, line 29: I also noticed that `/playsound` does the same thing, so it would be better to move this and that of `/playsound` to `SoundSource` itself, something like this, to prevent code duplication:
```rs
impl SoundSource {
    pub const VALUES: [SoundSource; 11] = [
        SoundSource::Master,
        SoundSource::Music,
        SoundSource::Records,
        SoundSource::Weather,
        SoundSource::Blocks,
        SoundSource::Hostile,
        SoundSource::Neutral,
        SoundSource::Players,
        SoundSource::Ambient,
        SoundSource::Voice,
        SoundSource::Ui,
    ];
    ...
```

steel-core/src/command/builtins/stopsound.rs
```diff
@@ -0,0 +1,104 @@
...
+    _context: &SteelCommandContext<CommandSource>,
+    source: Option<SoundSource>,
+    sound: Option<Identifier>,
+    targets: &[Arc<Player>],
+) -> Result<i32, CommandSyntaxError> {
+    let source = source.map(|source| source.as_varint());
+
+    for target in targets {
+        println!("Sending CStopSound to player");
+
```
Laptop59, line 97: I think you forgot to remove this after debugging.

steel-protocol/src/packets/game/c_stop_sound.rs
```diff
@@ -0,0 +1,35 @@
...
+            flags |= 2;
+        }
+
+        Self {
+            flags,
+            source,
+            sound,
+        }
+    }
+}
```
Laptop59: I think you will need to implement `WriteTo` for this packet manually as this packet uses the bitwise flag to check if it has to read the source and/or sound. Also, `Option`s in Minecraft are encoded with an extra byte, so it does not match Vanilla.

steel-protocol/src/packets/game/c_stop_sound.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub fn new(source: Option<i32>, sound: Option<Identifier>) -> Self {
+        let mut flags = 0;
+
+        if source.is_some() {
+            flags |= 1;
+        }
+
+        if sound.is_some() {
+            flags |= 2;
+        }
```
Laptop59, line 27: Magic constants `1` and `2`. Define them with `const` and use them.

*Vanilla actually defines the constants, but the Java compiler inlines the usages of primitives from a `static final` field to a literal number, so it looks like Vanilla directly uses a literal number, but it probably doesn't.*

steel-protocol/src/packets/game/c_stop_sound.rs
```diff
@@ -0,0 +1,35 @@
...
+use steel_registry::packets::play::C_STOP_SOUND;
+use steel_utils::Identifier;
+
+#[derive(ClientPacket, WriteTo, Clone, Debug)]
+#[packet_id(Play = C_STOP_SOUND)]
+pub struct CStopSound {
+    pub flags: u8,
+
+    #[write(as = Unprefixed(inner = VarInt))]
+    pub source: Option<i32>,
```
Laptop59, line 11: In my opinion, it would be better to store `SoundSource` in the option, implementing `WriteTo` for `SoundSource` in the file it's defined. That also means `CSound` should also be modified to take `SoundSource` instead of `i32`.

steel-core/src/command/builtins/stopsound.rs
```diff
@@ -0,0 +1,104 @@
...
+fn stop_sound(
+    context: &SteelCommandContext<CommandSource>,
+    source: SoundSource,
+) -> Result<i32, CommandSyntaxError> {
+    let targets = context.players("targets")?;
+
+    let sound = context.identifier("sound").ok().cloned();
+
+    execute(context, Some(source), sound, &targets)
+}
```
Laptop59, line 85: As `execute` doesn't need `context`, I think it should be removed as a parameter. Also, would prefer not to use a `Result` from `context.identifier()` to convert it to an `Option` using `ok()`.
If an actual error occurred with the `"sound"` parsing, then it would be interpreted as with 'no sound'.
Probably better to split this into with sound and without sound.

steel-core/src/command/builtins/stopsound.rs
```diff
@@ -0,0 +1,104 @@
...
+        argument("targets", SteelArgumentType::players()).executes(stop_all_for_targets);
+
+    for source in SOUND_SOURCES {
+        targets = targets.then(source_command(source));
+    }
+
+    literal("stopsound")
+        .executes(stop_all_for_source)
+        .then(targets)
+}
```
Laptop59, line 46: Where is the `*` literal? 26.2 also allows `*` for the sound source.

Laptop59 (changes requested):
This is pretty good, but I still have a few comments about the implementation.

CuzImClicks (changes requested):
please fix the change requested by @Laptop59

CuzImClicks (changes requested):
review by @Laptop59 and @Mathildeuh still must be addressed or resolved

KalSupe:
> review by @Laptop59 and @Mathildeuh still must be addressed or resolved
sorry I dint realise that my pr changes hadn't been added I have now applied the changes

Laptop59 (approved):

steel-core/src/command/builtins/stopsound.rs
```diff
@@ -0,0 +1,133 @@
...
+    context.source().send_success(&message, true);
+
+    target_count(&targets)
+}
+
+fn execute(
+    source: Option<SoundSource>,
+    sound: Option<&Identifier>,
+    targets: &[Arc<Player>],
+) -> Result<(), CommandSyntaxError> {
```
CuzImClicks: this function shouldnt return a result if it cannot fail/never returns the err variant

CuzImClicks (changes requested):
After that LGTM

CuzImClicks:
also please run `cargo clippy` and `cargo fmt` locally and fix problems

CuzImClicks (approved):

steel-core/src/command/builtins/stopsound.rs
```diff
@@ -0,0 +1,127 @@
...
+    target_count(&targets)
+}
+
+fn execute(source: Option<SoundSource>, sound: Option<&Identifier>, targets: &[Arc<Player>]) {
+    for target in targets {
+        target.send_packet(CStopSound::new(source, sound.cloned()));
+    }
+}
+
+fn target_count(targets: &[Arc<Player>]) -> Result<i32, CommandSyntaxError> {
```
aure31, line 123: you could create an utils to have less code duplication with the same method from the title commands

aure31 (approved):

coco875 (approved):

PR #686, opened 2026-09-15
fix max_players enforcement on join

### LINKED ISSUE #631: `max_players` isnt enforced when players join
`max_players` is sent in the status response and login packet, but it never gets checked when a player joins.

`try_reserve_player_join` only checks if the UUID already has an admission or online player. `admit_reserved_player` then inserts the player without checking the limit. different UUIDs can keep joining after the server is full, and the status response can report more online players than its max.

vanilla rejects a normal join with `multiplayer.disconnect.server_full` once the player count reaches the limit. players allowed to bypass the limit can still join.

the check should happen when the join is reserved and count online players plus `Joining` reservations under the same `player_admissions` lock. checking `player_count()` outside that lock would let multiple config tasks pass together and still go over the limit.

tests should cover:

- rejecting a normal join at the limit
- pending join reservations using a slot
- dropping a reservation making its slot available again
- the vanilla player limit bypass

kdcokenny:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

max_players was sent to clients but never enforced, so players could keep joining a full server.

this adds the early login check before duplicate session eviction and checks again after spawn preparation. the final check and insertion share the admissions lock so concurrent joins cant exceed the limit. pending joins only reserve their UUID, not a player slot.

bypass uses explicit boolean metadata, `steel:bypasses_player_limit`. being an operator alone doesnt grant it. configuration finish kicks also use the Play protocol the client expects.

closes #631

## How this was tested

checked the timing and bypass behavior against local vanilla 26.2 source.

four new tests cover admission timing, slot recovery, rejection cleanup, disconnect reasons and explicit bypass behavior, including changes during preparation.

also tested with two real Fabric GameTest clients on 26.2 against a one slot server. verified early and final admission rejection, operator status alone being rejected, explicit bypass allowing both players online, and joining after the slot is freed. both client runs passed.

## Screenshots / logs

```text
[player-limit] PASS final-admission-server-full: The server is full!
[player-limit] PASS early-login-server-full: The server is full!
[player-limit] PASS operator-alone-server-full: The server is full!
[player-limit] ALL CHECKS PASSED for LimitOccupant
[player-limit] ALL CHECKS PASSED for LimitProbe
```

## Checklist

- [x] Self-reviewed the diff

n/a


chocodev11:
lgtm and i cant open discord smh

CuzImClicks (dismissed):

steel-login/src/handlers/login.rs
```diff
@@ -50,6 +50,14 @@ impl JavaTcpClient {
         profile: GameProfile,
         reader_encryption: Option<[u8; 16]>,
     ) -> ConnectionAction {
+        // Vanilla rejects a full server before compression and duplicate-session eviction.
```
coco875: the comment are useful ?

steel-login/src/handlers/config.rs
```diff
@@ -101,14 +101,15 @@ impl JavaTcpClient {
             Ok(gameprofile) => gameprofile,
             Err(error) => return self.reject_unexpected_packet(error).await,
         };
+        // The client expects Play packets after acknowledging the end of configuration.
```
coco875: the comment are useful ?

coco875 (changes requested):

steel-login/src/handlers/login.rs
kdcokenny: I think so, it helps prevent someone moving the check during cleanup in the future which would cause problems.

steel-login/src/handlers/config.rs
kdcokenny: ya, it explains why the transition precedes a possible rejection. I can change to `// Switch before admission checks so rejection kicks use the Play protocol.` if you prefer

steel-login/src/handlers/login.rs
coco875: hmm it would be just better to move on function desc ? and remove mention of vanilla ?

steel-login/src/handlers/config.rs
coco875: the comment only say that it send a play packet, so  the line above don't really get what you mean

coco875 (changes requested):

kdcokenny:
I reworded both of them to explain why the order matters and removed the vanilla mention. but I kept them inline since they're about where those checks happen. is that ok?

steel-login/src/handlers/config.rs
```diff
@@ -101,7 +101,7 @@ impl JavaTcpClient {
             Ok(gameprofile) => gameprofile,
             Err(error) => return self.reject_unexpected_packet(error).await,
         };
-        // The client expects Play packets after acknowledging the end of configuration.
+        // Admission can reject the join; the client already expects Play disconnect packets.
```
coco875, line 104: for this one I still don't get it why put heremaybe above before hte let gameprofile are better ?

kdcokenny: it's about the protocol switch right below it. the reservation check after that can kick the player, so we need to switch to Play first. if I put it above `let gameprofile` itd look like it's explaining that check instead

coco875: just in case for me switch = match case because C x) else a bit weird if other maintainer don't see a problem then it's fine

kdcokenny: ah i meant changing the protocol to Play, not the match. that explains the confusion haha.

coco875 (approved):

CuzImClicks (approved):

PR #687, opened 2026-09-15
feat: implement new `Dispatched` strategy for non-scalar enums

Laptop59:
## Type of change

- [x] New feature

## Description

Allows a way for `WriteTo` to be derived for non-scalar enums (enums with extra data other than the variant itself), so that they can be written into packets.
- This is done by creating a new write strategy called "Dispatched", meaning that depending on the enum type, different data is written.
- Useful for implementing `CBossEvent` which use structures encoding different data depending on the type.

## How this was tested

Look at the generated code from the macro using a tool with an enum like
```rs
#[derive(WriteTo)]
#[write(as = Dispatched)]
enum A {
    A(i32, #[write(as = VarLong)] i64, bool),
    B {
        #[write(as = VarInt)]
        x: i32,
        y: i16
    },
    C,
    D
}
```
The generated implementation is
```rs
#[automatically_derived]
impl steel_utils::serial::WriteTo for A {
    fn write(&self, writer: &mut impl std::io::Write) -> std::io::Result<()> {
        match self {
            Self::A(value0, value1, value2) => {
                steel_utils::codec::VarInt(0i32).write(writer)?;
                value0.write(writer)?;
                steel_utils::codec::VarLong((*value1) as i64).write(writer)?;
                value2.write(writer)?;
            }
            Self::B { x: value0, y: value1 } => {
                steel_utils::codec::VarInt(1i32).write(writer)?;
                steel_utils::codec::VarInt((*value0) as i32).write(writer)?;
                value1.write(writer)?;
            }
            Self::C => steel_utils::codec::VarInt(2i32).write(writer)?,
            Self::D => steel_utils::codec::VarInt(3i32).write(writer)?,
        }
        Ok(())
    }
}
```

## Screenshots / logs

N/A

## Checklist

- [x] Self-reviewed the diff


steel-macros/src/write_to.rs
```diff
@@ -374,3 +388,70 @@ fn write_to_enum(name: Ident, attrs: Vec<syn::Attribute>) -> TokenStream {
...
+                if let Some(strat) = strategy {
+                    generate_write_code(&strat, quote! { (*#field_name) }, bound.as_ref())
+                } else {
+                    quote! {
+                        #field_name.write(writer)?;
+                    }
+                }
+            });
+
+            let field_names = fields.named.iter().map(|f| {
```
kdcokenny: a named field called `writer` shadows the output buffer here, so something like `Data { writer: i32 }` makes the derive fail to compile.

I'm thinking that we should prolly generate separate bindings for the fields so that their names cant collide with the output buffer.. oh & add a compile test for this case too

kdcokenny (changes requested):

steel-macros/src/write_to.rs
Laptop59: Thanks for the edge case (I did not notice that) and the new issue!

CuzImClicks (changes requested):
#689 should be fixed before this can get merged

steel-macros/src/write_to.rs
```diff
@@ -374,3 +395,72 @@ fn write_to_enum(name: Ident, attrs: Vec<syn::Attribute>) -> TokenStream {
         }
     })
 }
+
+fn dispatch_enum_variant_match_branch(
```
CuzImClicks, line 399: you can collapse this function to reuse, this spells out the same thing three times

CuzImClicks (changes requested):
After this LGTM

CuzImClicks (dismissed):

steel-macros/src/write_to.rs
```diff
@@ -292,7 +292,14 @@ fn write_to_struct(
...
 }
 
-fn write_to_enum(name: Ident, attrs: Vec<syn::Attribute>) -> TokenStream {
+fn write_to_enum(
+    s: syn::DataEnum,
+    name: Ident,
+    generics: &syn::Generics,
+    attrs: Vec<syn::Attribute>,
+) -> TokenStream {
+    let (impl_generics, ty_generics, _) = generics.split_for_impl();
```
kdcokenny, line 301: this drops the `where` clause, so a dispatched enum with `where T: WriteTo` won’t compile. could we preserve it in the generated impl?

kdcokenny (changes requested):

steel-macros/src/write_to.rs
Laptop59: Don't structs also have the same issue?

kdcokenny: you're right, i alr opened a non blocking issue for this lol i have amnesia https://github.com/Steel-Foundation/SteelMC/issues/689

kdcokenny (approved):

steel-macros/src/write_to.rs
Laptop59: > you're right, i alr opened a non blocking issue for this lol i have amnesia #689

That's referring to generics, but you were talking about the where clause.

steel-macros/src/write_to.rs
```diff
@@ -349,6 +356,20 @@ fn write_to_enum(name: Ident, attrs: Vec<syn::Attribute>) -> TokenStream {
                 }
             }
         }
+        // Write enum with dispatch based on enum variant
```
coco875, line 359: why decribe what the branch do ?

coco875: I have seen, it's the same on other branch I find it a bit dumb

steel-macros/src/write_to.rs
```diff
@@ -374,3 +395,70 @@ fn write_to_enum(name: Ident, attrs: Vec<syn::Attribute>) -> TokenStream {
...
+        steel_utils::codec::VarInt(#ordinal).write(writer)?;
+    };
+
+    match variant.fields {
+        Fields::Named(fields) => {
+            let writers = dispatch_enum_variant_field_writers(&fields.named);
+            let field_names = fields.named.iter().enumerate().map(|(i, f)| {
+                let field_name = f.ident.as_ref().expect("should have a named field");
+                let local_var_name = format_ident!("value{i}");
+                // Prevent name collisions
```
coco875, line 416: ? the comment are better above the line 416 ?

coco875 (changes requested):

steel-macros/src/write_to.rs
Laptop59: > why decribe what the branch do ?

It tells about how depending on the enum variant, different properties are written (with a match statement), hence the 'dispatch based on enum variant' (it's not really explained well admittedly). If this should be removed, then shouldn't the ones from the other branches be removed too?

steel-macros/src/write_to.rs
Laptop59: Yeah

PR #688, opened 2026-09-15
feat(registry): adding advancement parsing and registry

aure31:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

add the advancement loading from the vanilla_datapack and implement the advancement registry
also add the positioning logic of the nodes that use the same algorithm as minecraft

## How this was tested

unit test

## Checklist

- [x] Self-reviewed the diff


steel-registry/src/advancement/criterion.rs
coco875, line 1: maybe already start make the folder criterion where there is all criterion ?

coco875 (changes requested):
The code look clean, but there is a lack of example to see if it's a good api in use (add test are a way to do it or make a doc). Also do you send it to the client ? if not then I think you can add it in the same pr, so like that we can toy around, so for example try create diverse advancement tree

steel-registry/src/advancement/criterion.rs
aure31: i made the criterion here for now but the issue would be that criterion would need to interact with the player instance to validate if the player respect the conditions so i will maybe move out directly inside the steel-core directly.

aure31:
> The code look clean, but there is a lack of example to see if it's a good api in use (add test are a way to do it or make a doc). Also do you send it to the client ? if not then I think you can add it in the same pr, so like that we can toy around, so for example try create diverse advancement tree

I mean to send to the client i would need to add at least the command. Would you like that i add the command in this pull request too ?

PR #690, opened 2026-09-15
Implement secure chat verification and signed message chains

hthug06:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [X] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [X] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

### Changes for command:
- Add the greedy `message` argument parser (`minecraft:message`).
- Add `is_signed` across command parser
- Add the say command to try the message argument

### Packet handling
The packet id 8 is now handle ! No more warn message in the console lol.
More seriously, this packet handling mirror vanilla with checking for illegal characters, if the server enforce secure chat, when the sender send the last signed command... 

### Cryptography
Cryptography was the biggest part of this, but also the easiest, because the architecture already have (almost) all the struct for this pr. All the cryptographic exception are handle properly, mirroring vanilla.
For the cryptography to be used in a command, a struct `CommandSigningContext` as been added. It contains crypto info like the timestamp, the salt, ...

## How this was tested
You can try the by various way, but i'll put some screenshot for more simplicity. (all of the screenshot are taken in a SteelMC server)
First of all, test in the code.

### With enforce-secure-chat, online mode and encryption activated in the config, you can now report players with the official mojang system.

https://github.com/user-attachments/assets/51b880aa-69c7-4e78-911e-62bd2e279a46

Before, it was looking like this when reporting a player:
image

### In Console
You can now see (like in vanilla) when message a secured or not (with a cryptographic signature or not
image

### In game
You can check if the message is'nt signed in vanilla. you just have to over at the left of the message:
image
IF it say `Server message. Cannot be reported`, it doesn't have a cryptographic signature (so its an offline player, the console, a command block, ...)

## Checklist

- [X] Self-reviewed the diff

The broadcast_message() function is modified to now have an OutgoingChatMessage, that can be either a player, or a disguised (offline player, console, command block).

## Additional notes
Every command should be implementable now?


steel-core/src/player/chat/mod.rs
```diff
@@ -165,76 +400,141 @@ impl Player {
...
+            last_seen,
+        );
+
+        let link = chain.validate_and_advance(&body).map_err(|err| match err {
+            message_chain::ChainError::OutOfOrderChat => ChatValidationError::OutOfOrderChat,
+            message_chain::ChainError::ChainBroken => ChatValidationError::ChainBroken,
+            message_chain::ChainError::ExpiredProfileKey => ChatValidationError::ExpiredProfileKey,
+            message_chain::ChainError::MissingProfileKey => ChatValidationError::MissingProfileKey,
+            _ => ChatValidationError::Failed(Box::new(TextComponent::plain(
+                "Chain validation failed: {err}",
```
CuzImClicks: TextComponent::plain isnt like the format! macro so this actually just creates a textcomponent with:

`Chain validation failed: {err}`

steel-core/src/player/chat/mod.rs
```diff
@@ -493,19 +791,171 @@ impl Player {
...
+                    packet.salt,
+                    last_seen_sigs.clone(),
+                    &entry.signature,
+                ) {
+                    Ok(link) => {
+                        sender_index = link.index;
+                    }
+                    Err(err_component) => {
+                        drop(chat);
+                        self.send_message(&err_component.into_component());
```
CuzImClicks:
```java
	private void handleMessageDecodeFailure(final SignedMessageChain.DecodeException e) {
		LOGGER.warn("Failed to update secure chat state for {}: '{}'", this.player.getGameProfile().name(), e.getComponent().getString());
		this.player.sendSystemMessage(e.getComponent().copy().withStyle(ChatFormatting.RED));
	}
```

steel-protocol/src/packets/game/chat/player.rs
```diff
@@ -62,7 +71,7 @@ impl CPlayerChat {
             salt,
             previous_messages,
             unsigned_content,
-            filter_type,
+            filter_type: FilterType::PassThrough, // Change only on Reamls
```
CuzImClicks:
```suggestion
            filter_type: FilterType::PassThrough, // Change only on Realms
```

steel-core/src/command/builtins/say.rs
```diff
@@ -0,0 +1,36 @@
+//! Vanilla damage entity command.
```
CuzImClicks: fix doc comment

steel-core/src/player/chat/mod.rs
```diff
@@ -493,19 +791,171 @@ impl Player {
...
+                }
+            };
+
+            let Some(session) = chat.chat_session.clone() else {
+                drop(chat);
+                self.disconnect(CHAT_DISABLED_MISSING_PROFILE_KEY.msg());
+                return;
+            };
+
+            let argument_value = packet.command.split_once(' ').map_or("", |(_, arg)| arg);
```
CuzImClicks, line 877: `/msg Player hello` only signs `hello` on the client, but we sign `Player hello`

steel-protocol/src/packets/game/chat/player.rs
```diff
@@ -37,10 +48,9 @@ pub struct CPlayerChat {
...
 
 impl CPlayerChat {
-    #[expect(clippy::too_many_arguments)]
     #[must_use]
+    #[expect(clippy::too_many_arguments)]
     pub const fn new(
-        global_index: i32,
         sender: Uuid,
         index: i32,
         message_signature: Option<Box<[u8]>>,
```
CuzImClicks: lets convert all message signatures to use a newtype instead. since the signature always is 256 bytes long we dont need a box here

```rust
pub struct MessageSignature(pub [u8; 256]);
```

steel-core/src/player/chat/mod.rs
```diff
@@ -165,76 +400,141 @@ impl Player {
...
+        if chain.is_broken() {
+            return Err(ChatValidationError::ChainBroken);
+        }
+
+        let message_time = UNIX_EPOCH + Duration::from_millis(timestamp);
+        let now = SystemTime::now();
+        let message_age = now.duration_since(message_time).unwrap_or(Duration::ZERO);
+
+        if message_age > MESSAGE_EXPIRES_AFTER_SERVER {
+            return Err(ChatValidationError::Expired {
```
CuzImClicks:
```java
				if (unpacked.hasExpiredServer(Instant.now())) {
					SignedMessageChain.LOGGER.warn("Received expired chat: '{}'. Is the client/server system time unsynchronized?", body.content());
				}

				SignedMessageChain.this.nextLink = link.advance();
				return unpacked;
```

vanilla doesnt reject in this case

steel-core/src/player/chat/mod.rs
```diff
@@ -330,18 +621,25 @@ impl Player {
...
+        let outgoing = if self.server().enforces_secure_chat() {
+            OutgoingChatMessage::Player {
+                packet: Box::new(chat_packet),
+                signature,
+                sender_last_seen: last_seen,
             }
         } else {
-            for world in self.server().worlds.values() {
-                world.broadcast_unsigned_chat(chat_packet.clone());
+            OutgoingChatMessage::Disguised {
```
CuzImClicks:
```java
	static OutgoingChatMessage create(final PlayerChatMessage message) {
		return message.isSystem() ? new OutgoingChatMessage.Disguised(message.decoratedContent()) : new OutgoingChatMessage.Player(message);
	}
```

vanilla only puts the message in disguised if isSystem is true here

steel-core/src/command/builtins/say.rs
```diff
@@ -0,0 +1,36 @@
...
+use crate::player::chat::OutgoingChatMessage;
+use steel_registry::{RegistryEntry, vanilla_chat_types};
+use steel_utils::Identifier;
+
+pub(super) fn registration() -> CommandRegistration<CommandSource> {
+    CommandRegistration::new(Identifier::vanilla_static("say"), |_| command())
+}
+
+fn command() -> CommandNodeBuilder<CommandSource, SteelCommandRuntime> {
+    literal("say").then(argument("message", SteelArgumentType::message()).executes(
```
CuzImClicks, line 20: `/say @p` results in `@p` while in vanilla it results in `<your player name>`. this is a bug with SteelArgumentType::message

steel-core/src/command/builtins/say.rs
```diff
@@ -0,0 +1,36 @@
...
+        |ctx: &SteelCommandContext<CommandSource>| {
+            let source = ctx.source();
+
+            let message = ctx.message("message")?.to_string();
+
+            let chat_type = source.bind_chat_type(vanilla_chat_types::SAY_COMMAND.id() as i32);
+            let outgoing = OutgoingChatMessage::from_command(source, "message", message);
+
```
CuzImClicks, line 28: the result of /say is never printed to the console

steel-core/src/command/builtins/say.rs
CuzImClicks: but this is actually already fixed by @Mathildeuh in #682

CuzImClicks (changes requested):

steel-core/src/command/builtins/say.rs
hthug06: image

is this better or not what you were thinking about?

steel-core/src/command/builtins/say.rs
hthug06: So should i fix it?

PR #692, opened 2026-09-16
fix: preserve damage source entity references after removal

### LINKED ISSUE #571: DamageSource loses entity refs after removal

kdcokenny:
## Type of change

- [x] Bug fix
- [x] Breaking change

## Description

DamageSource now keeps strong references to the original entities, so damage attribution survives removal and can't resolve to a replacement entity with the same id.

callbacks that create damage sources now receive shared entity handles. recent damage history moves into server owned storage so self damage and mutual damage don't create reference cycles.

also releases equipment locks before applying enchantment effects to avoid a deadlock during thorns damage.

closes #571

## How this was tested

checked against vanilla 26.2 source. local Fabric client checks covered:

- self damage and explicit positions with `/damage @s 1 minecraft:generic by @s` and `/damage @s 1 minecraft:generic at 1.0 64.0 2.0`
- snowball and firework damage attribution
- death and respawn, dimension transfers, frozen and stepped game time, and clean shutdown

## Checklist

- [x] Self-reviewed the diff


chocodev11:
lgtm

steel-core/src/entity/damage/history.rs
```diff
@@ -0,0 +1,92 @@
...
+                    records.get(&victim).map(|record| record.source.clone()),
+                    None,
+                )
+            }
+        };
+        drop(expired);
+        source
+    }
+
+    pub(crate) fn clear_victim(&self, victim: EntityGeneration) {
```
Joniii11, line 78: So basically who clears this on normal entity remove or disconnect, cause the history is serverowned and can keep old entity & world alive if the clock is frozen, so mabye hook it into the victim lifetime idk

steel-core/src/behavior/items/firework_rocket.rs
```diff
@@ -63,12 +63,13 @@ impl ItemBehavior for FireworkRocketItem {
         );
         rocket.set_owner_uuid(Some(context.player.uuid()));
         let rocket = Self::add_rocket(context.world, rocket);
+        let owner: SharedEntity = context.player.clone();
```
Joniii11, line 65: so owner here is only used for enchant callback and rocket itself still stores only the uuid mc keeps exact player entity so after world change Steel can lose owner so maybe use the set_owner_entity

steel-core/src/behavior/items/firework_rocket.rs
```diff
@@ -98,17 +99,18 @@ impl ItemBehavior for FireworkRocketItem {
             next_entity_id(),
             Arc::downgrade(context.world),
             source_item,
-            context.player,
+            context.player.as_ref(),
```
Joniii11: attached_to_living keeps uuid & runtime id but not exact entity minewcraft stores attachedToEntity directly after entity id reuse this could rocket can attach to wrong entity

steel-core/src/behavior/items/firework_rocket.rs
```diff
@@ -98,17 +99,18 @@ impl ItemBehavior for FireworkRocketItem {
...
         );
         let rocket = Self::add_rocket(context.world, rocket);
         let has_infinite_materials = context.player.has_infinite_materials();
+        let owner: SharedEntity = context.player.clone();
```
Joniii11: so owner here is only used for enchant callback and rocket itself still stores only the uuid mc keeps exact player entity so after world change Steel can lose owner so maybe use the set_owner_entity
(same problem as in line 66)

steel-core/src/enchantment_helper/item_in_use.rs
```diff
@@ -0,0 +1,36 @@
...
+            Self::Equipped { owner, slot } => owner.with_equipment_slot(*slot, visitor),
+        }
+    }
+
+    pub(super) fn hurt_and_break(&mut self, amount: i32, infinite_materials: bool) -> bool {
+        match self {
+            Self::Stack(stack) => stack.hurt_and_break(amount, infinite_materials),
+            Self::Equipped { owner, slot } => {
+                let mut broke = false;
+                owner.with_equipment_slot_mut(*slot, &mut |stack| {
```
Joniii11, line 28: this reads the slot again after recursive effect in mc keeps exact itemstack if nested thorns changes or like breaks the item we can damage a different stack here

steel-core/src/entity/living_entity.rs
Joniii11: damagesource now keeps exact player but here we immediately reduce it back to uuid :C, mc keeps it as a cached player enttity ref, after world change this can lose kill credit

steel-core/src/world/mod.rs
```diff
@@ -204,6 +205,8 @@ pub enum ConditionalBlockSetResult {
 /// Configuration for creating a new world.
 #[derive(Clone)]
 pub struct WorldConfig {
+    /// Weak link to the server owner of exact recent damage sources.
+    pub damage_history: Weak<DamageHistory>,
```
Joniii11: worldconfig allows empty weak damage_history but a first successful living damage will panic would it be possible to make invalid world construction impossible instead of failing at runtime? 🌵

steel-core/src/server/run_loop.rs
```diff
@@ -585,6 +586,7 @@ impl Server {
     ) -> Result<(), WorldTickWorkerError> {
         if runs_normally {
             self.worlds.advance_domain_game_times();
+            self.damage_history.expire();
```
Joniii11, line 590: could we maybe benchmark it or replace it like with queued / sharded expiry? cause this scans full global damage history every tick and all worlds write through same mutex, could become a lock contention so like O(n) per tick with many records idk if that will become dramatic but could be worth looking into it yk

steel-core/src/entity/entity/mod.rs
```diff
@@ -3302,7 +3310,11 @@ pub trait Entity: EntityEventSource + ErasedType + Send + Sync + 'static {
     ///
     /// Mirrors vanilla's `Entity.move(MoverType, Vec3)`.
     /// Updates position, `on_ground`, velocity (on collision), and returns collision info.
-    fn move_entity(&self, mover_type: MoverType, delta: DVec3) -> Option<MoveResult> {
+    #[expect(
+        clippy::too_many_lines,
+        reason = "keeps vanilla movement phases in order"
+    )]
+    fn move_entity(self: Arc<Self>, mover_type: MoverType, delta: DVec3) -> Option<MoveResult> {
```
Joniii11: do we really need Arc<self> for whole movement hot path cause  normal movement now does extra atomic refcount work only because deeper path may create damagesource, mnaybe keep a shared handle closer to persistence bound or benchmark this if this is really shitty yk

steel-core/src/entity/ai/goal/selector.rs
```diff
@@ -99,7 +100,7 @@ pub trait Goal: Send {
         false
     }
 
-    fn tick(&mut self, _mob: &dyn PathfinderMob) {}
+    fn tick(&mut self, _mob: &dyn PathfinderMob, _entity: &SharedEntity) {}
```
Joniii11, line 103: Well so basically every tick a sharedentity now goes through every goal tick although most of the goals only need a borrow of it, thjis looks like a ownershipü leaking through whole ai api for only few damage / movement path

steel-core/src/behavior/item.rs
```diff
@@ -140,6 +141,7 @@ pub trait ItemBehavior: Send + Sync {
         &self,
         _world: &Arc<World>,
         user: &dyn LivingEntity,
+        _entity: &SharedEntity,
```
Joniii11: user and enttiy must be same object but api does not guarantee this, so wrong caller can run behaviour on A and use B for attribution, could this be one shared (maybe typed idk) context instead

steel-core/src/behavior/item.rs
```diff
@@ -174,6 +176,7 @@ pub trait ItemBehavior: Send + Sync {
         stack: &mut ItemStack,
         world: &Arc<World>,
         user: &dyn LivingEntity,
+        _entity: &SharedEntity,
```
Joniii11: user and enttiy must be same object but api does not guarantee this, so wrong caller can run behaviour on A and use B for attribution, could this be one shared (maybe typed idk) context instead 
(same poroblem as above lol)

steel-core/src/entity/damage/source.rs
```diff
@@ -0,0 +1,155 @@
...
+            damage_type,
+            causing_entity: None,
+            direct_entity: None,
+            source_position: None,
+        }
+    }
+
+    /// Adds the entity ultimately responsible for the damage.
+    #[must_use]
+    pub fn with_causing_entity(mut self, entity: SharedEntity) -> Self {
```
Joniii11, line 67: so basically this is fine yeah BUT many callers now set direct & causing to same entity wuith two calls would it be possible to create a helper so theres is a less chance a side is forgotten and code aesthetics lol

steel-core/src/entity/damage/source.rs
```diff
@@ -0,0 +1,155 @@
...
+        self.causing_entity = Some(entity);
+        self
+    }
+
+    /// Adds the direct entity that delivered the damage.
+    #[must_use]
+    pub fn with_direct_entity(mut self, entity: SharedEntity) -> Self {
```
Joniii11, line 74: see line 41

steel-core/src/entity/damage/tests.rs
```diff
@@ -0,0 +1,276 @@
...
+        &DamageSource::environment(&vanilla_damage_types::PLAYER_ATTACK)
+            .with_causing_entity(first.clone()),
+    );
+    let first_weak = Arc::downgrade(&first);
+    let second_weak = Arc::downgrade(&second);
+    let world_weak = Arc::downgrade(&world);
+    drop((first, second, world));
+    assert!(first_weak.upgrade().is_some());
+    assert!(second_weak.upgrade().is_some());
+    drop(history);
```
Joniii11, line 169: yes ik thius is nitpicking buttttttt this test drops global history itself, but prod keeps history alive till shutdown so this does not prove real lifecycle case

steel-core/src/enchantment_helper.rs
```diff
@@ -1420,8 +1429,8 @@ mod tests {
     fn post_attack_change_item_damage_calls_equipped_break_hook() {
```
Joniii11, line 1426: can we test real thorns recursion hjere plz

steel-core/src/entity/living_entity.rs
```diff
@@ -539,23 +552,17 @@ pub trait LivingEntity: Entity {
...
     }
 
     /// Resolves vanilla `LivingEntity.resolvePlayerResponsibleForDamage`.
-    fn resolve_player_responsible_for_damage(&self, world: &World, source: &DamageSource) {
-        let Some(entity_id) = source.causing_entity_id else {
-            return;
-        };
-        let Some(entity) = world.get_entity_by_id(entity_id) else {
+    fn resolve_player_responsible_for_damage(&self, source: &DamageSource) {
+        let Some(entity) = source.causing_entity() else {
```
Joniii11, line 563: not caused by this pr but this is what i found cuz mninecraft 26.2 also handles tamed wolf here and records its owner as lasthurtbyplayer we currently only handles player 📆

steel-core/src/entity/living_entity.rs
```diff
@@ -473,8 +473,21 @@ pub trait LivingEntity: Entity {
...
+
+    /// Records a successful hit using the victim's own level clock.
+    fn record_last_damage_source(&self, source: &DamageSource) {
+        let Some(world) = self.level() else {
+            panic!("recording living damage requires the victim's level");
+        };
+        let Some(history) = world.damage_history.upgrade() else {
+            panic!("recording living damage requires a live server damage history");
+        };
+        self.base().bind_damage_history(&history);
```
Joniii11: do we really need to bind damage_history on every successful hit cuz this adds extra lock or write in damage hot path although history normally does not change could bind once when entity enters world but probably wont be that huge of an issue ish idk

Joniii11 (changes requested):
2 or 3 comments are important the rest is just nit picking yeah ik im sorry :C

steel-core/src/entity/damage/history.rs
kdcokenny: great find, fixed!

steel-core/src/behavior/items/firework_rocket.rs
kdcokenny: fixed the owner caching here. the remaining gaps with vanilla were already there, so id handle those in a follow up PR.

steel-core/src/behavior/items/firework_rocket.rs
kdcokenny: fixed the id reuse issue and added a test. the remaining lifetime gap was already there, so id handle full vanilla parity in a follow up PR.

steel-core/src/behavior/items/firework_rocket.rs
kdcokenny: same fix here, the owner cache is now set before spawning.

steel-core/src/enchantment_helper/item_in_use.rs
kdcokenny: the stack identity gap is still open. vanilla thorns doesn't trigger another thorns pass, so id keep the ownership redesign separate unless we find a concrete case that needs fixing here.

steel-core/src/entity/living_entity.rs
kdcokenny: fixed the live player caching for kill credit and loot. the remaining lifetime and UUID lookup gaps were already there, so id handle those in a follow up PR.

steel-core/src/world/mod.rs
kdcokenny: changed `WorldConfig` to require `Arc<DamageHistory>`, so an empty weak reference can't be passed in. the caller still has to keep the history alive while the world runs.

steel-core/src/server/run_loop.rs
kdcokenny: benchmarked locally in release with 10k records. expiry took about 0.016 ms and collection took 0.36 to 0.84 ms. this wasn't a full server load test, but id keep the current mutex for now.

steel-core/src/entity/entity/mod.rs
kdcokenny: well cloning and dropping the handle measured about 12 ns locally so i don't think it's necessary. maybe revisit in a future perf pr

steel-core/src/entity/ai/goal/selector.rs
kdcokenny: this is borrowed as `&SharedEntity`, so passing it through doesn't clone it for every goal.

steel-core/src/behavior/item.rs
kdcokenny: fixed. both values now come from one `LivingEntityRef`, so callers can't supply different entities.

steel-core/src/behavior/item.rs
kdcokenny: same change here, `finish_using` takes `LivingEntityRef` now.

steel-core/src/entity/damage/source.rs
kdcokenny: Made it so that it sets both entities together now.

steel-core/src/entity/damage/source.rs
kdcokenny: same helper covers both fields now.

steel-core/src/entity/damage/tests.rs
kdcokenny: added disconnect tests with the clock frozen and the server and history still alive.

steel-core/src/enchantment_helper.rs
kdcokenny: yep

steel-core/src/entity/living_entity.rs
kdcokenny: hm ya i think it belongs in a followup

steel-core/src/entity/living_entity.rs
kdcokenny: it alr reuses the bindings but I removed the extra weak reference work on repeated hits

steel-core/src/behavior/items/firework_rocket.rs
Joniii11: sounds great

steel-core/src/enchantment_helper/item_in_use.rs
Joniii11: aight

steel-core/src/server/run_loop.rs
Joniii11: alrighty thanks for benchmarking it xD

steel-core/src/entity/entities/objects/projectiles/firework_rocket.rs
Joniii11: exact attachment is kept above but here we turn it back into runtime id again :C mc compares attachedToEntity object directly so id reuse can skip wrong nearby entity. can we keep allocation identity here?

```suggestion
            if attached
                .as_ref()
                .is_some_and(|attached| EntityArc::ptr_eq(attached, &target))
            {
```

steel-core/src/entity/entities/objects/projectiles/firework_rocket.rs
```diff
@@ -109,18 +110,19 @@ impl FireworkRocketEntity {
...
         world: Weak<World>,
         source_item: ItemStack,
-        attached_to: &dyn LivingEntity,
+        attached_to: LivingEntityRef<'_>,
     ) -> Self {
+        let attached_to = attached_to.entity();
         let rocket = Self::launched(entity_type, id, attached_to.position(), world, source_item);
-        rocket.set_owner_uuid(Some(attached_to.uuid()));
+        rocket.set_owner_entity(Some(attached_to));
+        rocket.state.lock().attached_to_entity = Some(EntityArc::downgrade(attached_to));
```
Joniii11, line 125: rebind is fixed now but lifetime still differs from vanilla here mcs attachedToEntity is direct strong livingentity ref and rocket keeps using it in tick with entityweak old attached allocation can disappear while rocket still alive should this be strong exact ref until rocket dies?

Joniii11 (changes requested):
Else lgtm

steel-core/src/entity/entities/objects/projectiles/firework_rocket.rs
kdcokenny: fixed

steel-core/src/entity/entities/objects/projectiles/firework_rocket.rs
kdcokenny: yeah you're right, the attachment is still a weak ref. that lifetime gap was already there, so id keep it in a follow up like i mentioned earlier. the collector fix handles the existing strong refs in AI goals and fishing hooks, it doesn't change the firework attachment. i think we should keep this seperate, wdyt?

Joniii11: yeah sure

Joniii11 (approved):
Looks good to me 🐱

PR #694, opened 2026-09-16
fix: fire aspect and looting on mob drops

Purdze:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

mobs killed with fire aspect dropped raw meat, and looting had no effect on mob drops.

loot entity refs never carried equipment, so the smelt condition's `direct_attacker` check always failed. looting read `ctx.tool`, which is never set when a mob dies. equipment is now included when loot rolls, and looting reads the attacker's enchantment level like vanilla. also fixes `enchanted_count_increase` applying its `limit` to the bonus instead of the stack.

## How this was tested

- new loot table tests: fire aspect → cooked porkchop, plain sword → raw, looting iii raises drop counts, looting on a helmet does nothing, stray tipped arrows capped at 1
- `cargo test --workspace`, `cargo clippy --workspace --all-targets --all-features`, `cargo fmt --all --check`, `typos .`

## Checklist

- [x] Self-reviewed the diff


steel-core/src/behavior/block/mod.rs
```diff
@@ -73,8 +75,12 @@ pub(crate) fn drop_from_block_interact_loot_table(
 
     // TODO: Add the block entity to the context when it can be done.
 
+    let interacting_equipment = interacting_entity.and_then(loot_equipment);
```
kdcokenny, line 78: this seems like it'll deadlock when carving a pumpkin. `inv.with_item` holds the inventory lock, then `loot_equipment` locks it again. maybe snapshot the shears and generate loot outside the callback? i also think that a carving test with a timeout, checking seeds, block state and durability would be helpful.

steel-registry/src/loot_table/tests.rs
```diff
@@ -294,6 +296,129 @@ fn test_pig_loot_smelt_condition_uses_entity_fire_flag() {
...
+    let player = living_entity_ref(&player_key, player_equipment);
+
+    (0..rolls)
+        .flat_map(|seed| {
+            let mut rng = rand::rngs::StdRng::seed_from_u64(seed);
+            let mut ctx = LootContext::new(&mut rng)
+                .with_killed_by_player(true)
+                .with_this_entity(victim)
+                .with_killer_entity(player)
+                .with_direct_killer_entity(player);
```
kdcokenny, line 348: could we test the real death context too? these supply equipment manually and use the same attacker and direct attacker. tests for separate attackers and conflicting tool/attacker enchantments for `RandomChanceWithEnchantedBonus` might be a goo d idea as well.

kdcokenny (changes requested):

kdcokenny (approved):

PR #695, opened 2026-09-17
feat: add the living entity set_item_slot equip hook

darkstarworks:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Adds `LivingEntity::set_item_slot` and the `on_equip_item` hook behind it.

Steel writes straight into an entity's equipment everywhere, so nothing runs when a slot changes. Vanilla routes those writes through `setItemSlot`, which plays the equip sound and emits the `equip` / `unequip` game event that sculk sensors pick up. Joniii pointed this out on #633, where a fox putting berries in its mouth skipped it.

* `set_item_slot` swaps the stack in and calls `on_equip_item` with the old and new stack
* `on_equip_item` does nothing for spectators, on the first tick, or when the item did not change. Otherwise it plays the equip sound when the item is equippable in that slot, and emits `equip` for equippable items and `unequip` for anything else
* `does_emit_equip_event` defaults to every slot, and players only emit it for armor
* A mob picking up gear now goes through `set_item_slot`
* Equipping an item on a mob or player by right-clicking uses the hook too, which replaces the two TODOs waiting on game event dispatch

## How this was tested

New tests register a game event listener next to an entity and check that equipping a helmet emits `equip`, putting berries in its hand and then emptying it emits `unequip` twice, and that re-equipping the same helmet or changing a slot on the first tick emits nothing.

`cargo fmt --all --check`, `cargo clippy -r --all-targets --all-features`, `typos` and `cargo test` all pass.

## Screenshots / logs

Not applicable, this is server-side logic.

## Checklist

- [x] Self-reviewed the diff

## Additional notes

* **Other direct writes.** Loading equipment from NBT stays a plain write, since vanilla does not run the hook there either. The fox mouth item in #513 and #633 moves to `set_item_slot` once this lands.
* **Swapping armor on from your own hand** still goes straight through the player inventory, so it plays no equip sound yet. It carries a `TODO(equip-hook)`, since fixing it means changing what `try_swap_with_equipment_slot` returns.


kdcokenny (approved):

steel-core/src/entity/living_entity.rs
```diff
@@ -1643,6 +1645,46 @@ pub trait LivingEntity: Entity {
...
+            || self.is_first_tick()
+        {
+            return;
+        }
+
+        let equippable_slot = new_stack.get_equippable().map(|equippable| equippable.slot);
+        if equippable_slot == Some(slot)
+            && let Some(sound) = self.equip_sound(slot, new_stack)
+        {
+            self.play_sound(sound, 1.0, 1.0);
```
Joniii11: so basically minecraft doesnt use the enmtity playsound here since the player playosund excludes itself so the player cant hear own equip sound, also the silent player still broadcasts here.
Mineaccraft uses playseededsound on the level

steel-core/src/entity/living_entity.rs
```diff
@@ -1643,6 +1645,46 @@ pub trait LivingEntity: Entity {
...
+            || ItemStack::is_same_item_same_components(old_stack, new_stack)
+            || self.is_first_tick()
+        {
+            return;
+        }
+
+        let equippable_slot = new_stack.get_equippable().map(|equippable| equippable.slot);
+        if equippable_slot == Some(slot)
+            && let Some(sound) = self.equip_sound(slot, new_stack)
+        {
```
Joniii11, line 1680: pig override returns none for every non saddle slot, in minecraft the pig falls back to super getequipsound so normal armor pickup on pig should still play the item equip sound

steel-core/src/player/mod.rs
```diff
@@ -2009,13 +2009,14 @@ impl LivingEntity for Player {
             self.inventory.lock().set_changed();
         }
 
-        if let Some(sound) = self.equip_sound(slot, &equipped) {
-            self.play_sound(sound, 1.0, 1.0);
-        }
-        // TODO: Emit EQUIP game event once game-event dispatch is implemented.
+        self.on_equip_item(slot, &ItemStack::empty(), &equipped);
```
Joniii11, line 2012: this only fixes equip on target only, but not normal armor inventory slot still skips the hook, in minecraft the ArmorSlot.setByPlayer calls owner.onEquiopItem
So the player behaviour stays different depending how armour was equipped

steel-core/src/entity/living_entity.rs
```diff
@@ -1643,6 +1645,46 @@ pub trait LivingEntity: Entity {
...
     }
 
+    /// Puts `stack` in `slot` and runs [`on_equip_item`](Self::on_equip_item).
+    fn set_item_slot(&self, slot: EquipmentSlot, stack: ItemStack) {
+        let new_stack = stack.clone();
+        let mut old_stack = stack;
+        self.with_equipment_slot_mut(slot, &mut |current| {
+            mem::swap(current, &mut old_stack);
+        });
+        self.on_equip_item(slot, &old_stack, &new_stack);
```
Joniii11, line 1663: there are still runtime removals bypassing this path, so equipment shearing uses raw take, but vanilla mob.shearItem does setItemSlot(slot, EMPTY) so uneqip event is missing ;C

Joniii11, line 1663: mob death equipment removal also still takes raw, minecraft calls in the fn dropCustomDeathlooth calls the setItemSlot(slot, EMPTY) after dropping items so steel misses the unequip event here again x2 heehe 2 comments go brrt

steel-core/src/player/mod.rs
```diff
@@ -2009,13 +2009,14 @@ impl LivingEntity for Player {
...
         InteractionResult::Success
     }
 
+    fn does_emit_equip_event(&self, slot: EquipmentSlot) -> bool {
```
Joniii11, line 2016: the player hand events are correctly disabled but f key hand swap still bypasses setItemSlot, minecraft does setItemInHand ->> setItemSlot for both hands so custom hand equippable can miss equip sound which is bad cause then no souznd grrr

steel-core/src/entity/tests/equipment_and_freezing.rs
```diff
@@ -347,3 +347,77 @@ fn update_fall_flying_stops_when_glider_gate_fails() {
...
+            ),
+            (
+                EquipmentSlot::MainHand,
+                ItemStack::new(&vanilla_items::SWEET_BERRIES),
+            ),
+            (EquipmentSlot::MainHand, ItemStack::empty()),
+        ],
+    );
+
+    assert_eq!(
```
Joniii11, line 397: please also add mismatched slot cases, for example head equippable put in mainhand must emit equip but not equip sound, this protects impoortant vanilla distinction here or whatever and yes this is nitpicking

steel-core/src/entity/tests/equipment_and_freezing.rs
```diff
@@ -347,3 +347,77 @@ fn update_fall_flying_stops_when_glider_gate_fails() {
...
+        ],
+    );
+    assert_eq!(same_item, vec![&vanilla_game_events::EQUIP]);
+
+    let first_tick = equip_game_events(
+        "set_item_slot_first_tick",
+        true,
+        &[(EquipmentSlot::Head, helmet)],
+    );
+    assert_eq!(first_tick, Vec::<GameEventRef>::new());
```
Joniii11, line 427: missing spectator case too

Joniii11 (changes requested):
please change that

steel-core/src/behavior/item.rs
```diff
@@ -102,6 +102,8 @@ pub trait ItemBehavior: Send + Sync {
 
         match result {
             EquipmentSwapResult::Success(overflow) => {
+                // TODO(equip-hook): run `on_equip_item` for the swapped slot, so armor
+                // put on from the hand plays its equip sound and emits the equip event.
```
Joniii11: could we not remove this todo? i mean it isnt that hard maybe something like this?
```suggestion
let (previous, result) = context.inv.with_inventory(|inventory| {
    let previous = EntityEquipment::get_ref(inventory, slot).clone();
    let result = inventory.try_swap_with_equipment_slot(
        context.hand,
        slot,
        context.player.has_infinite_materials(),
    );
    (previous, result)
});

match result {
    EquipmentSwapResult::Success(overflow) => {
        let equipped = context
            .player
            .inventory
            .lock()
            .get_ref(slot)
            .clone();

        context.player.on_equip_item(slot, &previous, &equipped);

        if !overflow.is_empty() {
            let _ = context.player.drop_item(overflow, false, false);
        }
        InteractionResult::Success
    }
    EquipmentSwapResult::Fail => InteractionResult::Fail,
}
```

Joniii11 (commented):
suggestion

darkstarworks:
> please change that

on it

steel-core/src/player/mod.rs
darkstarworks: You're right. This one needs a shape decision first: `Slot::set_by_player` takes no player and `ArmorSlot` has no owner, and `Player` is not built inside an `Arc`, so a `Weak<Player>` owner would have to be bound at every creation site. Passing `&Player` through `set_by_player` works, but it also reaches `safe_insert`, `move_item_stack_to` and `update_quick_move_source`, plus the container tests that have no player. Which would you rather have?

steel-core/src/entity/living_entity.rs
darkstarworks: Right, switched to the world sound with nobody excluded, so the wearer hears it too. The silence gate stays, `Entity::play_sound` already had that one.

steel-core/src/entity/living_entity.rs
darkstarworks: done. Added `default_equip_sound` so an override can fall back to the item's own sound, and the pig only special-cases the saddle now.

steel-core/src/entity/living_entity.rs
darkstarworks: done, `attempt_to_shear_equipment` clears the slot through `set_item_slot`.

darkstarworks: done, same for the death drop, after `spawn_at_location` like vanilla. Two for one 🙂

steel-core/src/player/mod.rs
darkstarworks: done, both hands run the hook on a swap, offhand first.

steel-core/src/entity/tests/equipment_and_freezing.rs
darkstarworks: done, a helmet in the main hand now has a case: it emits the equip event and stays silent. Not nitpicking, that distinction was untested.

steel-core/src/entity/tests/equipment_and_freezing.rs
darkstarworks: done.

steel-core/src/behavior/item.rs
darkstarworks: You're right, it wasn't hard. Took your version with one change: it reads the equipped stack back through `with_inventory` instead of taking the inventory lock a second time. Thanks for writing it out.

Purdze (approved):

steel-core/src/entity/living_entity.rs
```diff
@@ -1637,12 +1642,67 @@ pub trait LivingEntity: Entity {
...
```
CuzImClicks, line 1663: you can use `EntityEquipment::set` to make this more clean

steel-core/src/behavior/item.rs
```diff
@@ -92,16 +93,22 @@ pub trait ItemBehavior: Send + Sync {
...
-            inventory.try_swap_with_equipment_slot(
+        let (previous, result) = context.inv.with_inventory(|inventory| {
+            let previous = EntityEquipment::get_ref(inventory, slot).clone();
+            let result = inventory.try_swap_with_equipment_slot(
                 context.hand,
                 slot,
                 context.player.has_infinite_materials(),
-            )
+            );
+            (previous, result)
```
CuzImClicks, line 103: you dont have to take the lock later again if you just return this aswell here

`let equipped = EntityEquipment::get_ref(inventory, slot).clone();`

CuzImClicks (changes requested):

steel-core/src/entity/living_entity.rs
darkstarworks: done, it goes through `EntityEquipment::set` now.

steel-core/src/behavior/item.rs
darkstarworks: done, both stacks are read inside the one `with_inventory` call.

PR #611, opened 2026-09-06
Add foundations for hostile mobs

starwarslevi:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

shared foundation for hostile mobs: goals/targeting, pathfinding, attributes, combat, equipment, spawning, persistence, daylight burning, leash gating, entity conversion

left out on purpose: brain (PR #249), ranged combat (PR #423)

## How this was tested

cargo test -p steel-core (2598 passing), clippy clean, fmt + workspace green

## Checklist

- [x] Code builds w/o errors or warnings
- [x] Self-reviewed the diff
- [ ] Docs updated (if applicable)
- [x] No leftover debug code / comments

## Additional notes

ur da goat

PR #490, opened 2026-08-25
feat: implement vanilla turtle entity

darkstarworks:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [x] Entity implementation
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Adds the vanilla turtle (MC 26.2), pairing with the turtle egg block from #467.

* all nine vanilla goals at their vanilla priorities
* breeding gives the mother an egg, laid on home sand
* turtle move control, water travel, spawn rule and scute drop

Shared additions:

* `BreedGoal::with_breed` and `PanicGoal::always_seeking_water`, so the turtle reuses both goals
* `MoveToBlockGoal::without_stay_limit`
* `World::are_full_chunks_loaded_at` and `LevelReader::sea_level`
* `TurtleEggBlock::on_sand` / `is_sand` take `&dyn LevelReader`
* `Entity::default_play_swim_sound`, so the turtle can scale the volume and keep the shared pitch

## How this was tested

Unit tests for the goals, breeding, egg laying, movement, spawn rule, dimensions and save/load.

`cargo fmt --all --check`, `RUSTFLAGS="-Dwarnings" cargo clippy -r --all-targets --all-features` and `RUSTFLAGS="-Dwarnings" cargo test` pass.

## Screenshots / logs

Not applicable, server-side logic.

## Checklist

- [x] Code builds w/o errors or warnings
- [x] Self-reviewed the diff
- [ ] Docs updated (if applicable)
- [x] No leftover debug code / comments

## Classes / commands modified:

Turtle

## Additional notes

* `TODO(amphibious-navigation)`: Steel has no `AmphibiousPathNavigation`; a zero water malus stands in.
* `TODO(leash)`: waits on #696.
* `TODO(lightning)`: no lightning bolt entity yet.

PR #696, opened 2026-09-17
feat: let each mob decide whether it can be leashed

darkstarworks:
## Type of change

- [ ] Block implementation
- [ ] Item implementation
- [ ] Command implementation
- [ ] Entity implementation
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / code cleanup
- [ ] Performance improvement
- [ ] Chore / tooling

## Description

Adds a `Mob::can_be_leashed` hook that the blanket `Leashable` impl delegates to.

Every mob gets `Leashable` through `impl<T: Mob> Leashable for T`, so no single mob could override `can_be_leashed`. Vanilla has 18 classes that do (turtles, villagers, dolphins, squids and more). This keeps the blanket impl, as discussed on #490, and adds just the one delegate mobs actually need today.

* `Mob::can_be_leashed` defaults to `true`
* the blanket impl forwards `Leashable::can_be_leashed` to it
* non-mob leashables (boats) keep the `Leashable` default
* the enemy `TODO` moved onto the `Mob` default, since vanilla's rule lives on `Mob`
* the one call site that now sees both traits calls `Mob::can_be_leashed` explicitly

## How this was tested

A new mob test: a mob that returns `false` from its own `can_be_leashed` refuses a lead that it accepts otherwise. Removing the delegate makes it fail.

`cargo fmt --all --check`, `RUSTFLAGS="-Dwarnings" cargo clippy -r --all-targets --all-features` and `RUSTFLAGS="-Dwarnings" cargo test` all pass.

## Screenshots / logs

Not applicable, this is server-side logic.

## Checklist

- [x] Self-reviewed the diff

## Additional notes

* **Overlap with #611.** That PR overrides `can_be_leashed` in the blanket impl with `!self.is_enemy()`. Whichever lands second, that check moves into the `Mob` default instead, and the `TODO` goes away.
* **Delegates are added as we go.** Only `can_be_leashed` for now, per @SomeYellowGuy on #490. #490 uses it to make turtles refuse a lead.


steel-core/src/entity/entity/mod.rs
```diff
@@ -1408,7 +1408,7 @@ pub trait Entity: EntityEventSource + ErasedType + Send + Sync + 'static {
         }
 
         if player.is_secondary_use_active()
-            && mob.can_be_leashed()
+            && Mob::can_be_leashed(mob)
```
Joniii11: minecraft entity.interact calls leashable.canBeLeashed here not Mob directly since the blanket impl already delegates to Mob could we jusjt do it directly via leashable

steel-core/src/entity/mob/mod.rs
```diff
@@ -1718,6 +1724,10 @@ impl<T: Mob> Leashable for T {
     fn leash_data(&self) -> &SyncMutex<Option<LeashData>> {
         self.mob_base().leash_data()
     }
+
+    fn can_be_leashed(&self) -> bool {
```
Joniii11, line 1728: do we want the same method name on mob and leashable cuz it makes can_be_leashed ambiguous on every mob and already forces UFCS at the interaction callsite

Joniii11 (changes requested):
please do this then its a LGTM

steel-core/src/entity/entity/mod.rs
darkstarworks: done, goes through `Leashable` now. It only needed the explicit form because of the name clash below, which is gone.

steel-core/src/entity/mob/mod.rs
darkstarworks: You're right. Renamed to `mob_can_be_leashed`, which also matches `mob_ai_step` and friends, and the ambiguity is gone.

CuzImClicks (approved):

Purdze (approved):

Joniii11 (approved):
lgtm

Purdze:
pls rebase

darkstarworks:
> pls rebase

@Purdze done. My other open PRs also

